Issue #148, opened 2026-04-21
Bug Report: makeslice: cap out of range panic during WAL Segment Fencing


karnzx:
### Summary
I have been running Milvus with Woodpecker for some time. Recently, I encountered an issue where a Milvus streaming node restarted due to a panic: `makeslice: cap out of range` error. This could be related to Woodpecker running within the streaming node.

Woodpecker utilizes MinIO/S3 object storage to manage WAL segments. During operations such as a WAL channel rebalance, the streaming nodes initiate a segment fencing operation (`recoverFromStorageUnsafe`). This process relies on `StatObject` to determine the footer block size and subsequently calls `ReadObjectFull` to load the footer data into memory.

When a transient object storage error occurs (like connection reset or context cancellation), `StatObject` erroneously emits a `-1` size that propagates through to a slice allocation. This triggers an unrecoverable `panic: runtime error: makeslice: cap out of range` that crashes the entire streaming node under pressure.

### Repro
1. Deploy a Milvus cluster (v2.6.x) with WAL backed by Woodpecker and MinIO/S3.
2. Have active streaming nodes with WAL channels containing segments with footer blocks.
3. Trigger a WAL channel rebalance (e.g., node restart or scaling) while the object storage is under heavy load or experiencing intermittent transient failures.
4. The rebalance triggers `fenceAllActiveSegments()` -> `recoverFromStorageUnsafe()`. 
5. `StatObject()` fails with a transient error instead of an expected "object not found". 
6. The streaming node panics and enters a crash loop.

**Trigger Event (Streaming Node Logs):**
```json
{"level":"INFO","message":"expected state changed, do a life cycle","module":"streamingnode","channel":"search-prod-rootcoord-dml_4","expected":"(79,false)"}
{"level":"INFO","time":"2026/04/16 07:43:07.581 +00:00","caller":"walmanager/manager_impl.go:91","message":"remove wal success","module":"streamingnode","component":"wal-manager","channel":"search-prod-rootcoord-dml_4","term":79}
```

**Full Panic Stack Trace:**
```go
panic: runtime error: makeslice: cap out of range

goroutine 14773282683 gp=0xc00ab8fdc0 m=5465257 mp=0xc02a3f4008 [running]:
runtime.makeslice(0x839d788?, 0xc0285368d0?, 0x7a610e9?)
	/usr/local/go/src/runtime/slice.go:113 +0x93
[...]

PR #149, opened 2026-04-21
fix: prevent makeslice panic in recoverFromStorageUnsafe on transient StatObject errors

karnzx:
issue: #148

### DIFF
--- a/common/minio/minio_client_helper.go
+++ b/common/minio/minio_client_helper.go
@@ -250,7 +250,10 @@ func ReadObjectFull(ctx context.Context, objectReader FileReader, initReadBufSiz
 	_, sp := logger.NewIntentCtxWithParent(ctx, ObjectStorageScopeName, "ReadObjectFull")
 	defer sp.End()
 	start := time.Now()
-	// Initial buffer size - 1MB is a reasonable starting point
+	// Initial buffer size - validate to prevent panic from negative values
+	if initReadBufSize <= 0 {
+		initReadBufSize = 1024 * 1024 // fallback to 1MB default
+	}
 	buf := make([]byte, 0, initReadBufSize)
 
 	// Temporary read buffer
--- a/common/minio/minio_client_helper.go
+++ b/common/minio/minio_client_helper.go
@@ -250,10 +250,7 @@ func ReadObjectFull(ctx context.Context, objectReader FileReader, initReadBufSiz
 	_, sp := logger.NewIntentCtxWithParent(ctx, ObjectStorageScopeName, "ReadObjectFull")
 	defer sp.End()
 	start := time.Now()
-	// Initial buffer size - validate to prevent panic from negative values
-	if initReadBufSize <= 0 {
-		initReadBufSize = 1024 * 1024 // fallback to 1MB default
-	}
+	// Initial buffer size - 1MB is a reasonable starting point
 	buf := make([]byte, 0, initReadBufSize)
 
 	// Temporary read buffer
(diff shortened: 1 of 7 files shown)


common/objectstorage/minio_object_storage.go
```diff
@@ -82,7 +82,7 @@ func (minioObjectStorage *MinioObjectStorage) PutFencedObject(ctx context.Contex
 func (minioObjectStorage *MinioObjectStorage) StatObject(ctx context.Context, bucketName, objectName string, operatingNamespace string, operatingLogId string) (int64, bool, error) {
 	info, err := minioObjectStorage.minioHandler.StatObject(ctx, bucketName, objectName, minio.StatObjectOptions{}, operatingNamespace, operatingLogId)
 	if err != nil {
-		return -1, false, err
+		return 0, false, err
```
tinswzy: I don't think we should return 0, as 0 carries valid semantic meaning in this context. Explicitly returning -1 would be better; it ensures that if the downstream code uses the size illegally, it will trigger a panic. This makes bugs much easier to expose and fix—just like the issue we're currently addressing. Returning 0 could lead to unintended runtime behavior and potential data errors, which are far more difficult to debug.

common/minio/minio_client_helper.go
```diff
@@ -250,7 +250,10 @@ func ReadObjectFull(ctx context.Context, objectReader FileReader, initReadBufSiz
 	_, sp := logger.NewIntentCtxWithParent(ctx, ObjectStorageScopeName, "ReadObjectFull")
 	defer sp.End()
 	start := time.Now()
-	// Initial buffer size - 1MB is a reasonable starting point
+	// Initial buffer size - validate to prevent panic from negative values
+	if initReadBufSize <= 0 {
+		initReadBufSize = 1024 * 1024 // fallback to 1MB default
+	}
```
tinswzy: It would be better if the unit tests covered here.
codecov minio_client_helper.go detail: https://app.codecov.io/gh/zilliztech/woodpecker/pull/149?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=zilliztech

karnzx:
The the root cause is entirely in writer_impl.go. In this file, recoverFromStorageUnsafe was ignoring transient errors and passing an invalid size (-1) downstream, which caused the panic.

The changes in the other two files are just defense-in-depth additions to prevent future panics.

tinswzy:
> The the root cause is entirely in writer_impl.go. In this file, recoverFromStorageUnsafe was ignoring transient errors and passing an invalid size (-1) downstream, which caused the panic.
> 
> The changes in the other two files are just defense-in-depth additions to prevent future panics.

No defensive coding is needed here. An error is an error; this size shouldn't be used at all. When an error occurs, the size becomes meaningless. Incorrect usage should trigger a panic—think of it as a runtime 'assert' for internal method calls.

> 1. Reject "Zombie States"; Choose Decisive Death
> Defensive programming often misfires by "swallowing" errors through excessive fault tolerance (e.g., returning a default 0 or null). This creates a zombie state—the system appears operational, but its internal logic has veered off-course. This silent decay allows errors to accumulate and eventually pollute databases, turning root-cause analysis into a "needle-in-a-haystack" nightmare. In contrast, a direct Panic freezes the fault instantly, stopping the rot before it spreads.
> 
> 2. Embrace "Crash & Rebirth"; Flush Corrupt States
> Borrowing from Erlang’s "Let it Crash" philosophy: rather than writing endless if-else blocks to predict every unknowable anomaly, it is better to let the process fail. A robust reboot strategy completely flushes out corrupted in-memory states, allowing the system to instantly return to a known, clean, and correct initial environment. This is often the only effective way to resolve complex, transient "Heisenbugs."
> 
> 3. Mitigate Loss; Preserve the "Crime Scene"
> In data-sensitive operations, a Panic acts as the ultimate circuit breaker. It halts execution before a corrupted instruction can hit the disk, preventing "secondary damage" or permanent data loss. Furthermore, the resulting stack trace and core dump serve as a high-fidelity record of the crime scene, allowing developers to pinpoint the exact logical failure instead of chasing ghost-like downstream side effects.

karnzx:
I have reverted all of the "defense-in-depth" changes. The PR now only contains the core logic fix in `recoverFromStorageUnsafe`. Please review.

tinswzy (approved):
/lgtm

tinswzy:
Could you help fix CodeChecker Error? You can run `make lint-fix` locally to fix it.

PR #160, opened 2026-05-14
fix: update Prometheus image to soldevelo/prometheus:latest in docker-compose.monitor.yaml

igor-soldev:
This change replaces the bitnami/prometheus image with the soldevelo/prometheus image to ensure the test and monitoring stack remains fully functional and accessible without external dependency constraints.

The Bitnami Prometheus image is increasingly problematic in some environments due to distribution and availability limitations, which can lead to broken CI pipelines or inconsistent local setups. By switching to the SolDevelo-maintained image, we ensure that:

* the monitoring stack remains fully runnable out-of-the-box
* CI/CD pipelines do not depend on externally restricted or unstable images
* the project uses a drop-in compatible replacement with the same configuration model
* we reduce risk of future breakages caused by upstream availability or policy changes

This change aligns with previous updates in the repository where Bitnami images were replaced with SolDevelo equivalents to maintain consistency and reliability across the test environment.

### DIFF
--- a/tests/docker/monitor/docker-compose.monitor.yaml
+++ b/tests/docker/monitor/docker-compose.monitor.yaml
@@ -19,7 +19,7 @@ services:
     restart: "no"
 
   prometheus:
-    image: bitnami/prometheus:latest
+    image: soldevelo/prometheus:latest
     container_name: prometheus
     ports:
       - "9090:9090"


tinswzy:
Thanks for the suggestion! Using the official Prometheus community image `prom/prometheus:latest` might be a better choice.

igor-soldev:
@tinswzy I’m responsible for maintaining this image, so using `soldevelo/prometheus:latest` keeps the setup aligned with the rest of our maintained stack. In that case, there’s also no need to change paths or Bitnami-related dependencies/configuration, since the image is intended as a drop-in compatible replacement.

tinswzy:
> @tinswzy I’m responsible for maintaining this image, so using `soldevelo/prometheus:latest` keeps the setup aligned with the rest of our maintained stack. In that case, there’s also no need to change paths or Bitnami-related dependencies/configuration, since the image is intended as a drop-in compatible replacement.

Thanks for the clarification.

  I understand that `soldevelo/prometheus` is maintained by your team and may align better with your company’s internal stack. However, this Docker Compose file is part of the upstream open-source project and should use a neutral, broadly expected default.

  If we switch the default image to a company-maintained replacement for this case, other users may run into the same concern from the opposite direction, since they may also use their own Prometheus image or internal mirror. Those environment-specific image choices are better handled by local
  overrides or by downloading/modifying the Compose file locally, rather than changing the upstream default.

  If the Bitnami image is the problem, I think switching to the official Prometheus image, such as `prom/prometheus`, would be a better direction.

  Could you clarify whether the official Prometheus image cannot satisfy this setup for some technical reason, or whether this is mainly about keeping image management consistent with your maintained stack? That context would also be useful for other open-source users.

igor-soldev:
Thanks, I understand the concern about keeping upstream defaults neutral and broadly accessible.

soldevelo/prometheus is intended as a compatibility-maintained fork of the current Bitnami-based setup, keeping the same configuration model while addressing upstream availability and maintenance issues we've observed in CI environments.

We did consider prom/prometheus, but it would require changes in configuration assumptions used in this compose setup, while the current image remains drop-in compatible.

Additionally, the image is hardened (non-root execution, reduced package surface, and regular CVE scanning/patching), which aligns with the existing monitoring stack approach in the repository.

That said, I understand the upstream neutrality concern. If preferred, we can switch the default to prom/prometheus and document soldevelo/prometheus as an optional override for users who need a Bitnami-compatible/hardened build.

tinswzy:
Thanks, that sounds reasonable.

  If we switch the upstream default to the official Prometheus image, I think we should make the image layout explicit rather than mixing both options in a single compose file.

  My suggestion is:

  - `tests/docker/monitor/docker-compose.monitor.yaml`
    - default upstream config
    - uses `prom/prometheus`
    - removes the Bitnami-specific entrypoint
    - uses `--storage.tsdb.path=/prometheus`

  - `tests/docker/monitor/docker-compose.monitor.bitnami.yaml`
    - optional override for the Bitnami-compatible layout
    - keeps the Bitnami-specific entrypoint and data path
    - can use `bitnami/prometheus:latest` by default
    - users who need your maintained compatible image can replace it with `soldevelo/prometheus:latest`

  For example, the optional override could contain:

  ```yaml
  # Bitnami-compatible Prometheus override.
  # If bitnami/prometheus is unavailable in your environment, you may replace it
  # with a compatible image such as soldevelo/prometheus:latest.
  services:
    prometheus:
      image: bitnami/prometheus:latest
      entrypoint:
        - /opt/bitnami/prometheus/bin/prometheus
      command:
        - '--config.file=/etc/prometheus/prometheus.yml'
        - '--storage.tsdb.path=/opt/bitnami/prometheus/data'
        - '--web.enable-lifecycle'
 ```

  And the default compose file would use the official image like this:

 ```yaml
  services:
    prometheus:
      image: prom/prometheus:latest
      command:
        - '--config.file=/etc/prometheus/prometheus.yml'
        - '--storage.tsdb.path=/prometheus'
        - '--web.enable-lifecycle'
 ```
  This keeps the upstream default neutral and broadly expected, while still documenting a clear path for users who need the Bitnami-compatible or SolDevelo-maintained image.

Issue #103, opened 2026-03-10
enhance: Refactor Log Sync Mechanism to Support 100k+ Tenants via Event-Driven Scheduling


tinswzy:
**1. Background / Problem Description**
Currently, woodpecker assigns a dedicated Goroutine to each log stream on the server side. These Goroutines run in a loop to monitor and sync buffered data.
The Scalability Bottleneck:
- Resource Exhaustion: Supporting 100k+ logs on a single node causes the memory footprint of Goroutine stacks (min. 2KB-4KB each) to exceed 400MB - 800MB just for idle stacks, even before accounting for actual log buffers.
- CPU Scheduler Pressure: The Go runtime scheduler must constantly scan 100k+ Goroutines. Even if they are in a sleep state, the wake-up and context-switching overhead for such a massive number of entities severely degrades CPU efficiency.
- Inefficiency: In many multi-tenant scenarios (e.g., IoT or Microservices), only a small fraction of logs are "active" at any given millisecond. The current "One-Goroutine-Per-Log" model wastes resources on inactive tenants.

**2. Proposed Solution: Event-Driven & Lazy Activation**
We propose refactoring the sync logic from a Static Polling model to an Event-Driven Task Pool model. This decouples the "Log Entity" from the "Execution Thread".
Key Components:
- Lazy Activation: A log stream will NOT have an associated Goroutine by default. It remains a passive data structure until the first byte of data is ingested.
- Global Scheduler (Timing Wheel):
  - Instead of 100k timers, we use a Single Timing Wheel to manage all expiration events.
  - When a log becomes "Active" (first data arrives), it registers a one-time timeout task (e.g., 5s) in the timing wheel.
- Shared Worker Pool: A fixed-size pool of Worker Goroutines (e.g., $N = \text{NumCPU} \times 2$) handles the actual Sync() I/O operations.
- Task Queue: When a log is ready to sync (either via MaxDelay timeout or BufferFull event), its LogID is pushed to a central SyncQueue.
The "Silent-to-Active" Workflow:
1. Ingest: Data arrives $\rightarrow$ Update Buffer.
2. Trigger: If isActive == false: set isActive = true and register with the Global Scheduler.
3. Dispatch: Scheduler or Buffer-Threshold-Monitor pushes LogID to Worker Pool.
4. Sync & Hibernate: Worker performs I/O $\rightarrow$ If buffer is empty, set isActive = false (the log goes back to sleep).

**3. Implementation Plan**
[...]

PR #163, opened 2026-05-19
stagedstorage: add shared sync scheduler

tinswzy:
issue: https://github.com/zilliztech/woodpecker/issues/103

  Replace per-writer sync goroutines with a shared scheduler and bounded worker
  pool. Writers now schedule delayed sync checks only when they have contiguous
  syncable entries, and shared workers process each writer's flush task queue
  while preserving per-writer flush ordering.

  Also add scheduler tests, scheduled-sync regression tests for out-of-order
  writes and rescheduling, plus benchmarks for worker dispatch latency and
  delayed-check lateness

### DIFF
--- a/common/metrics/service_metrics.go
+++ b/common/metrics/service_metrics.go
@@ -223,6 +223,31 @@ var (
 		Name:      "file_stored_count",
 		Help:      "Current number of local segment files",
 	}, []string{"node_id", "namespace", "log_id"})
+
+	WpSyncSchedulerScheduled = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: woodpeckerNamespace,
+		Subsystem: serverRole,
+		Name:      "sync_scheduler_scheduled",
+		Help:      "Number of delayed staged-writer sync checks scheduled",
+	}, []string{"node_id"})
+	WpSyncSchedulerRunning = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: woodpeckerNamespace,
+		Subsystem: serverRole,
+		Name:      "sync_scheduler_running",
+		Help:      "Number of running staged-writer sync scheduler workers",
+	}, []string{"node_id"})
+	WpSyncSchedulerWaiting = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: woodpeckerNamespace,
+		Subsystem: serverRole,
+		Name:      "sync_scheduler_waiting",
+		Help:      "Number of staged-writer sync scheduler jobs waiting",
+	}, []string{"node_id"})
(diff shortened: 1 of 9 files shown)


tinswzy:
Benchmark results (-benchtime=1x)

```
  +------------------------+-------------+---------------------------------------------------------------+-------------------------------------------+-----------------------------------------------------------
  -------------------------------------------+
  | Benchmark              | Scale       | Meaning                                                       | Result                                    | Interpretation
  |
  +------------------------+-------------+---------------------------------------------------------------+-------------------------------------------+-----------------------------------------------------------
  -------------------------------------------+
  | Job dispatch           | 1k jobs     | Queueing latency from tryEnqueueJob() to worker start         | avg 20.30 us/job, max 192.5 us            | Shared worker queue overhead is low
  |
  | Job dispatch           | 10k jobs    | Same as above                                                 | avg 16.69 us/job, max 395.6 us            | Still microsecond-level under a 10k burst
  |
  | Job dispatch           | 100k jobs   | Same as above                                                 | avg 22.01 us/job, max 496.8 us            | Even under a 100k burst, the slowest dispatch is about 0.5
  ms; the worker queue is not the bottleneck |
  | Delayed check lateness | 1k checks   | Expected due time to actual delayed-check trigger, delay=10ms | avg late 1.225 ms/check, max 1.265 ms     | With 1k active writers expiring together, scheduler
  lateness is about 1 ms                           |
  | Delayed check lateness | 10k checks  | Same as above                                                 | avg late 1.916 ms/check, max 2.861 ms     | With 10k checks expiring together, lateness is still only
  a few ms                                    |
  | Delayed check lateness | 100k checks | Same as above                                                 | avg late 20.509 ms/check, max 31.083 ms   | 100k checks expiring together is an extreme burst;
  lateness grows to tens of ms                       |
  +------------------------+-------------+---------------------------------------------------------------+-------------------------------------------+-----------------------------------------------------------
  -------------------------------------------+
```

  Notes
```

  +------------------------+-------------------------------------------------------------------------------------------------------------------------+
  | Item                   | Note                                                                                                                    |
  +------------------------+-------------------------------------------------------------------------------------------------------------------------+
  | Test mode              | -benchtime=1x, submitting a one-shot burst of 1k/10k/100k tasks                                                         |
  | Delayed check scenario | All checks use delay=10ms, so this mostly simulates a batch of writers expiring at almost the same time                 |
  | Measurement scope      | This measures scheduler overhead only. It does not include real Sync() execution, file writes, or fsync cost            |
  | Latency relationship   | Scheduled sync startup latency is roughly configured delay + delayed check lateness + job dispatch latency              |
  | Conclusion             | For 100k logs with about 1k active writers, the 1k case is closer to expected pressure; 100k is the extreme burst case   |
  +------------------------+-------------------------------------------------------------------------------------------------------------------------+
```

PR #170, opened 2026-06-01
enhance: load-aware quorum node selection (#114)

### LINKED ISSUE #114: enhance: Implement Load-Aware Quorum Selection Strategy
**Description**
The current Quorum selection follows a Static Regional or Pure Random strategy. In this model, the system selects nodes either from a pre-defined Zone (e.g., us-east-1a) or picks them randomly across the cluster to satisfy the $W+R > N$ consistency requirement.

**The Problem:**
Static/Random selection is "blind" to the Runtime Heterogeneity of nodes. Even within the same designated region, certain nodes may experience:
- Resource Contention: Heavy background jobs or localized "hotkeys" causing CPU spikes.
- I/O Wait: Disk saturation on specific storage nodes.
- Network Jitter: Micro-bursts affecting specific rack-level switches.
Relying solely on regional proximity or randomness leads to Tail Latency (P99) inflation, as the Quorum is only as fast as its slowest (most loaded) member.

**Proposed Solution: Load-Aware "Smart" Selection**
We need to evolve the selection logic to a Hybrid Strategy:
1. Regional Constraint (Tier 1): Maintain data sovereignty and low-latency by filtering nodes within the preferred Zone.
2. Load Filtering (Tier 2): From the candidate pool in Tier 1, prioritize nodes based on a Dynamic Health Score.
**Proposed Algorithm Logic:**
- Weighted Probabilistic Selection: Instead of 1/N chance, use $P(node_i) = \frac{Capacity_i - CurrentLoad_i}{\sum (AvailableCapacity)}$.
- Penalty Box: Temporarily "mute" nodes whose load exceeds a critical threshold (e.g., 85% CPU) for a short window (Cool-off period).

**Technical Implementation**
- Metadata Enhancement: The Node Manager should heartbeat not just "ALIVE" status, but also a LoadFactor (0.0 - 1.0).
- Selection Interface: Refactor QuorumRouter.select() to accept a LoadProvider interface.
- Fall-through Logic: If all nodes in a high-priority Zone are heavily loaded, the strategy should decide whether to:
  - Option A: Stick to the Zone but accept higher latency.
  - Option B: "Overflow" to a lower-load node in a neighboring Zone (if latency allows).
  
 **Acceptance Criteria**
- [ ] The Quorum selector can prioritize low-load nodes within the same region.
- [ ] Introduce a max_load_threshold configuration to prevent pushing "unhealthy but alive" nodes over the edge.
- [ ] Metrics added to track "Quorum Selection Skew" (how often we deviate from random due to load).

tinswzy:
## Summary

Implements load-aware quorum node selection (closes #114). Within already-eligible candidates (Tier-1 AZ/RG filtering unchanged), selection now prefers lower-load nodes via **weighted-probabilistic** selection instead of pure random, with load propagated through the existing **gossip heartbeat metadata** (no HTTP scraping, no I/O in the selection path).

### How it works
- **Load factor (producer):** each node samples its own load in `[0,1]` — `max(cpu, ioWaitRate)` as the flow bottleneck, with memory folded in only as a high-end penalty above a soft threshold — EWMA-smoothed, and publishes it into its `NodeMeta` (`load_factor` + `load_updated_at`). memberlist gossip propagates it cluster-wide.
- **Selection (consumer):** reads each candidate's load from the in-memory discovery index (pure, no I/O):
  1. **Penalty box** — mute candidates whose fresh load ≥ `maxLoadThreshold`
  2. **Zone fall-through** — if all candidates are muted, fall through to the full set (stick to zone, accept higher latency)
  3. **Weighted pick** — `weight = 1 − load`; unknown/stale load gets a neutral weight; if no candidate has fresh load, degrade to random
- **Observability:** `woodpecker_server_quorum_selection_skew_total{mode}` counts `weighted` / `fallthrough` / `random_no_load`.
- **Config:** `woodpecker.logstore.nodeSelectionPolicy` (`loadAwareEnabled`, `maxLoadThreshold`, `loadReportInterval`, `loadTTL`, `memSoftThreshold`, `ewmaAlpha`) with defaults + validation.

### Acceptance criteria (#114)
- [x] Prioritize low-load nodes within the same region
- [x] `max_load_threshold` config to avoid pushing unhealthy-but-alive nodes over the edge
- [x] Quorum-selection-skew metric

### Design notes / deviations
- Weighted-probabilistic (not greedy argmin) — avoids herd/oscillation onto the momentarily-idle node.
- `LoadProvider` realized as the producer-side `LoadSampler` interface; the consumer reads load directly from gossiped `NodeMeta`.
- Fall-through implements Option A (stick to zone); cross-zone overflow (Option B) intentionally out of scope.
- Backward compatible: nodes that never report load (`load_updated_at = 0`) are treated as unknown and never mistaken for "idle".

## Test Plan
- [x] `go build ./...`, `go vet ./...` clean
- [x] Unit tests: `common/membership`, `common/config`, `common/metrics` pass
- [x] `go test -race ./common/membership/` — no races (covers reporter goroutine + selection)
- [x] Statistical selection tests (non-flaky over repeated runs): low-load nodes get the majority of traffic, overloaded nodes are muted, both low nodes share traffic (weighted, not greedy)
- [ ] Reviewer: validate IO-wait rate behavior under real cluster load


tinswzy:
## Load-aware quorum selection — performance benchmark

Benchmarked the load-aware node selection across cluster sizes to confirm it adds no meaningful latency. Quorum size **K = 3**; loads spread across `[0, 1)` so the weighted path (not the random fallback) is exercised.

Environment: Apple M1 Pro, `darwin/arm64`, Go `-benchmem`. Absolute values differ on x86 CI/prod, but the **O(N)** shape and orders of magnitude hold.

### 1. Core algorithm — `selectLowestLoadNodes` (weighted sampling only, candidate gathering excluded)

| Nodes | Latency | ns/op | Mem/op | Allocs/op |
|------:|--------:|------:|-------:|----------:|
| 3 | 0.14 µs | 137.8 | 48 B | 2 |
| 6 | 0.24 µs | 242.9 | 192 B | 5 |
| 20 | 0.41 µs | 412.4 | 528 B | 5 |
| 50 | 0.77 µs | 771.9 | 1.3 KB | 5 |
| 100 | 1.4 µs | 1,404 | 2.7 KB | 5 |
| 500 | 6.5 µs | 6,524 | 12 KB | 5 |
| 1,000 | 13 µs | 13,191 | 24 KB | 5 |
| 5,000 | 63 µs | 63,148 | 123 KB | 5 |
| 10,000 | 119 µs | 118,714 | 246 KB | 5 |

### 2. End-to-end — `SelectRandom` (candidate gather + tag/decommission filter + load-aware selection)

| Nodes | Latency | ns/op | Mem/op | Allocs/op |
|------:|--------:|------:|-------:|----------:|
| 3 | 0.19 µs | 192.2 | 72 B | 3 |
| 6 | 0.33 µs | 334.1 | 304 B | 7 |
| 20 | 0.65 µs | 653.5 | 1.1 KB | 9 |
| 50 | 1.3 µs | 1,296 | 2.7 KB | 10 |
| 100 | 2.3 µs | 2,314 | 5.7 KB | 11 |
| 500 | 10.2 µs | 10,172 | 25 KB | 13 |
| 1,000 | 21 µs | 20,694 | 50 KB | 14 |
| 5,000 | 133 µs | 133,203 | 319 KB | 19 |
| 10,000 | 263 µs | 262,989 | 638 KB | 21 |

### 3. Baseline — `randomSelectNodes` (no-load fallback, for comparison)

| Nodes | Latency | ns/op |
|------:|--------:|------:|
| 3 | 0.002 µs | 2.07 |
| 6 | 0.05 µs | 50.9 |
| 20 | 0.09 µs | 88.5 |
| 50 | 0.14 µs | 135.4 |
| 100 | 0.20 µs | 203.6 |
| 500 | 0.71 µs | 714.1 |
| 1,000 | 1.7 µs | 1,696 |
| 5,000 | 14 µs | 13,972 |
| 10,000 | 22 µs | 22,476 |

### Takeaways

- **Linear in node count** (K fixed): ~12 ns/node for the core algorithm, ~26 ns/node end-to-end. Cost = building the weight array `O(N)` + `K` sampling passes `O(K·N)`.
- **Negligible at realistic cluster sizes**: 3 nodes → 0.14 µs, 100 nodes → 1.4 µs — 3–4 orders of magnitude below a single `SelectNodes` gRPC round-trip (ms range).
- **Even 10k nodes < 0.3 ms** end-to-end (263 µs), so selection is never the bottleneck at any realistic (or unrealistic) scale.
- **Weighting overhead vs plain random**: ~5× at 10k (119 µs vs 22 µs), but still well under a millisecond in absolute terms.
- **Memory is `O(N)`** per call (allocs constant at 5; bytes grow with N — `weightedSampleIndices` copies the weight + index arrays). Only worth optimizing under an extreme large-cluster + high-QPS scenario; not needed today.

Benchmark added in `common/membership/load_select_benchmark_test.go` (test-only, no production code touched).

Issue #202, opened 2026-06-25
perf: client-side group commit (batched AddEntries RPC) to amortize per-entry round-trip on single-log writes


tinswzy:
## Summary

Each `AppendOp` issues one `AddEntry` RPC carrying a **single** `LogEntry`, and the per-log `SequentialExecutor` processes one op per iteration. The per-entry network round-trip therefore caps single-log append throughput. Introducing **client-side group commit** — opportunistically coalescing consecutive append ops into one batched `AddEntries` RPC — amortizes the round-trip over many entries and lifts single-log throughput toward the flush / object-storage ceiling.

## Evidence

Profiled a single-shard, `batch=1` insert workload (one woodpecker log) driven through Milvus streamingnode:

- Single-log append throughput plateaus at **~810 appends/s** regardless of client concurrency.
- **Not CPU bound** (~13–16%) and **not flush bound** (server flush ~19% busy with `maxFlushThreads=32`, ~5× headroom).
- The limiter is "one entry per round-trip through the single serial worker." With `batch=1` row inserts, rows/s == appends/s, so per-message/per-round-trip overhead dominates. The server already batches at flush time; the client→server path does not.

## Root cause (code)

- `proto/logstore.proto`: `rpc AddEntry(AddEntryRequest)` with `AddEntryRequest { ... LogEntry entry }` — one entry per RPC; no batch RPC exists.
- `woodpecker/segment/sequential_executor.go`, `worker()`: pulls and executes one op per iteration.

## Proposed change

1. **Opportunistic batch drain (no timer).** In `worker()`, after receiving one op, non-blockingly drain whatever is already queued (bounded by `maxBatchEntries` / `maxBatchBytes`) and execute them as a single batch. This is self-tuning: batch size 1 at low load (no added latency), larger batches under load when there is a backlog.

   ```go
   case op := <-se.operationQueue:
       batch := []Operation{op}
       for len(batch) < maxBatchEntries {
           select {
           case n := <-se.operationQueue:
               batch = append(batch, n)
           default:
               goto flush
           }
       }
   flush:
       se.executeBatch(batch)
   ```

2. **New batch RPC.** `rpc AddEntries(AddEntriesRequest) returns (stream AddEntriesResponse)` where `AddEntriesRequest` carries `repeated LogEntry entries`; the server streams back per-entry `Buffered` / `Synced` keyed by `entry_id`. The server handler reuses the existing sequential buffer and flush path.

[...]

PR #208, opened 2026-06-25
feat(client): opt-in client-side group commit (batched AddEntries)

tinswzy:
## What

Adds **client-side group commit** to the single-log write path, plus the batched server/client machinery around it, and turns it **on by default**. Consecutive `AppendOp`s waiting in a segment's `SequentialExecutor` are coalesced into a single `AddEntries` request per replica, amortizing the per-entry send round-trip that caps single-log append throughput.

Implements #202. Composes with #206 (parallel per-replica fan-out, merged).

> **Consolidation note.** This work was originally split as a stack #208 → #211 → #212. The follow-ups are now folded into this PR as individual commits (see the commit list), since they are one cohesive group-commit workstream and review together more easily. **#211 and #212 are closed in favor of this PR.**

## Why

Profiling a single-shard, `batch=1` write workload, single-log append throughput plateaus at ~810 appends/s while CPU (~13–16%) and server-side flush (~19% busy, 32 flush threads) have large headroom. The limiter is "one entry per round-trip through the single serial worker": `SequentialExecutor.worker()` runs one `AppendOp` per iteration and each entry travels in its own `AddEntry` stream. Coalescing entries into one request amortizes the round-trip and lifts single-log throughput toward the flush/object-storage ceiling, while high-concurrency tail latency drops correspondingly.

## Design

- **proto** — new `AddEntries(AddEntriesRequest) returns (stream AddEntriesResponse)` carrying `repeated LogEntry`; the server streams back one Buffered response per entry, then one Synced/Failed per entry, every response keyed by `entry_id`.
- **server** (`service.go`) — `AddEntries` buffers all entries first (one round-trip of Buffered acks), then streams per-entry Synced/Failed, reusing the existing per-entry buffer/flush path.
- **client** (`logstore_client_remote.go` / `_local.go`) — `LogStoreClient.AppendEntries`: the remote impl sends one stream, reads the Buffered acks (the single amortized round-trip), then asynchronously demuxes the per-entry Synced/Failed results into each entry's result sink by `entry_id`; the local (embedded) impl loops `AddEntry` since it has no round-trip to amortize.
- **segment** (`batch_append_op.go`) — `BatchAppendOp` is a **send-side grouping only**. Each entry keeps its own `AppendOp`, result channel, quorum tracking and in-order acknowledgement via `SendAppendSuccessCallbacks`.
- **executor** (`sequential_executor.go`) — opportunistic, **non-blocking** batch drain: it only coalesces ops already queued, so batch size is 1 at low load (**no added latency**) and grows under contention. Only consecutive `AppendOp`s are batched; retries and other operations run individually. `wg` accounting is preserved (one `Done` per op).
- **config** — `SegmentAppend.MaxBatchEntries` / `MaxBatchBytes`, **on by default** (see below).

## Ordered-log correctness

The log is strictly ordered with a LAC prefix watermark, so a batch's "successful part" is always a contiguous **prefix**, never an arbitrary subset. Because `BatchAppendOp` keeps a per-entry `AppendOp` and reuses `SendAppendSuccessCallbacks` (which already acks the contiguous completed prefix, in order, as soon as it is ready), partial success/failure falls out of the existing machinery: the committed prefix returns success promptly, and any un-acked suffix is retried (or fenced/rolled) by the per-entry failure path. No out-of-order acks, no holes.

## Consolidated write-path optimizations

On top of the core send-side batching, this PR folds in three refinements (previously #211 / #212 and one new), each stacked on the previous. Measured with `TestServiceWritePathBreakdown` (1KB entries, conc=8000, 10ms flush, 2MB batch, 3-replica service mode; laptop mini-cluster, directional numbers):

| Commit | Change | Throughput |
|---|---|---|
| group commit | baseline for this series | 23,165/s |
| `perf(server): batch buffer-append` | `AddEntries` handler buffers the whole batch under one `w.mu` via `AddEntryBatch` → `WriteDataBatchAsync` (was per-entry) | 26,805/s (+15.7%) |
| `perf(client): drain batch acks with one goroutine per node` | one per-node goroutine reads results in entry order (was N `receivedAckCallback` goroutines) | 28,992/s (+8.2%) |
| `perf(client): coalesce LAC sync` | one background LAC pusher per writable segment pushes only the latest LAC (was a `go syncLACToQuorumAsync` fan-out per ack round) | 32,519/s (+12.2%) |

Cumulative **+40%** over the group-commit baseline, 0 failures. LAC coalescing is advisory/best-effort (the authoritative final LAC is still recorded via `CompleteSegment`); read-only/legacy handles fall back to the per-call fan-out so their behavior is unchanged.

## On by default

Group commit is now **on by default** (`MaxBatchEntries=1000`, `MaxBatchBytes=2MB`; whichever is reached first bounds a batch). Batching is opportunistic (drains only what is already queued), so it adds no latency at low load. Set `MaxBatchEntries=1` to opt back out. `TestDisorderMultiAppendAsync_AllSuccess_InSequential`, which mocks the single-entry `AppendEntry` path, pins `MaxBatchEntries=1` so it keeps covering that path under the new default.

## Tests

New, all passing under `-race`:
- `TestRemoteClient_AppendEntries_Demux` — buffered ids + per-entry Synced routed by id (out-of-order responses).
- `TestRemoteClient_AppendEntries_StreamErrorFailsRemaining` — un-acked entries are failed on stream end (no hang).
- `TestBatchAppendOp_Execute_QuorumAck` — one `AppendEntries` per replica; every entry reaches quorum and is acked.
- `TestSequentialExecutor_CoalescesAppendOps` — queued ops coalesce into one batch (single-entry path intentionally unmocked, so a non-batched send fails loudly).

Also: `go build ./...`, `go vet`, `gofumpt`, `gci` clean; full `./woodpecker/segment/`, `./woodpecker/client/`, `./server/`, `./common/config/` suites pass, including `-race` over the rolling/close/ack paths for the LAC change. `proto/*.pb.go` and the `LogStoreClient` mock were regenerated with the repo's pinned toolchain (protoc-gen-go v1.36.2, grpc v1.5.1, vtproto v0.6.0).

Issue: #202


tinswzy:
### Benchmark: throughput vs. batch size

`BenchmarkGroupCommit` (added in this PR) isolates the **client send path** — it drives the real `SequentialExecutor` + `AppendOp` / `BatchAppendOp` code, with the `LogStoreClient` mocked to `time.Sleep(rtt)` per round-trip (`rtt = 1ms`, a representative per-entry buffered round-trip) and deliver the ack. It does **not** model the server-side durable flush or the network, so the numbers below are an upper bound on the *send-path* speedup, not an end-to-end cluster number.

Single log, single-node quorum, `rtt = 1ms`, mean of 5 runs (`-benchtime=1s -count=5`, debug logging off):

| maxBatch | throughput (appends/s) | speedup | marginal gain / doubling |
|---:|---:|---:|---:|
| 1 (off) | 796 | 1.0× | — |
| 4 | 3,140 | 3.9× | — |
| 8 | 6,239 | 7.8× | ×2.0 |
| 16 | 12,581 | 15.8× | ×2.0 |
| 32 | 24,213 | 30.4× | ×1.9 |
| 64 | 45,579 | 57.3× | ×1.9 |
| 128 | 57,418 | 72.1× | ×1.26 |
| 256 | 63,138 | 79.3× | ×1.10 |

Reproduce:
```
go test ./woodpecker/segment/ -run '^$' -bench BenchmarkGroupCommit -benchtime=1s -count=5
```

### How much it lifts the send path

Up to ~`batch 32–64` throughput grows almost linearly with batch size: this is the regime where the single ~1ms round-trip is being amortized across the batch, and it's exactly the bottleneck this PR removes. On this isolated path the worker goes from ~800 appends/s (round-trip-bound) to **~63k appends/s (~80×)**.

### Why it then plateaus — and why that's expected

Beyond ~`batch 64–128` the marginal gain collapses (×1.26 → ×1.10 per doubling) and the curve approaches a ceiling of ~65k appends/s. This is expected: **batching only amortizes the cost shared by the whole round-trip (the RTT). The per-entry work is paid for every entry regardless of batch size**, namely:

- proto marshal/unmarshal of each entry;
- server side: appending each entry to the buffer (its own entry id, buffer slot, framing);
- client side: one `AppendOp` + one result channel + one `receivedAckCallback` + ack matching per entry;
- per-entry quorum bookkeeping (`ackSet`).

If `c` is that fixed per-entry cost, the model is:

```
throughput(B) = B / (RTT + c·B)  →  1/c   as B → ∞
```

So once `c·B ≫ RTT` (here around `B ≈ 64`), the RTT is fully amortized and throughput is capped by the per-entry work the single ordered-log worker must do — not by the round-trip. The plateau is `~1/c`.

### Caveats

- **Synthetic / send-path only.** Absolute numbers depend on the machine and on `rtt`; the *shape* and the *speedup ratio* are the point. In a real cluster the realized throughput is additionally bounded by the server-side durable flush (local-disk staged writes), so the end-to-end gain is smaller than this CPU-bound upper bound. The value this demonstrates is that **the client serial-send is no longer the bottleneck once batched** — it moves to per-entry work / flush.
- The current `BatchAppendOp` spawns one goroutine per entry for the ack wait, which is part of `c`; a leaner per-batch ack loop would raise the ceiling (at very large batches — `≫256` — this goroutine fan-out also starts to *degrade* throughput).

### Takeaway

Most of the benefit is captured by a **modest batch (≈16–64)**; larger batches give diminishing returns (and cost memory + latency). This is why batching stays **off by default** (`MaxBatchEntries = 1`) and a small value is the right setting when enabling it.

server/service.go
```diff
@@ -452,6 +452,83 @@ func (s *Server) AddEntry(request *proto.AddEntryRequest, serverStream grpc.Serv
...
+	// Phase 1: buffer all entries and acknowledge each as Buffered.
+	resultChs := make([]channel.ResultChannel, n)
+	for i, reqEntry := range request.Entries {
+		entry := &proto.LogEntry{
+			SegId:   reqEntry.SegId,
+			EntryId: reqEntry.EntryId,
+			Values:  reqEntry.Values,
+		}
+		resultCh := channel.NewLocalResultChannel(fmt.Sprintf("srv-batch/%d/%d/%d", request.LogId, reqEntry.SegId, reqEntry.EntryId))
+		bufferedId, err := s.logStore.AddEntry(streamCtx, request.BucketName, request.RootPath, request.LogId, entry, resultCh)
```
tinswzy: It should return after adding batch entries to buffered at one time

common/config/configuration.go
```diff
@@ -41,6 +41,13 @@ type SegmentRollingPolicyConfig struct {
 type SegmentAppendConfig struct {
 	QueueSize  int `yaml:"queueSize"`
 	MaxRetries int `yaml:"maxRetries"`
+	// MaxBatchEntries is the maximum number of consecutive append ops coalesced
+	// into a single AddEntries request (client-side group commit). The batch is
+	// opportunistic: it only coalesces ops already waiting in the queue, so it
+	// adds no latency at low load. Values <= 1 disable batching.
+	MaxBatchEntries int `yaml:"maxBatchEntries"`
+	// MaxBatchBytes caps the total payload of a coalesced batch. 0 = no byte limit.
+	MaxBatchBytes ByteSize `yaml:"maxBatchBytes"`
```
tinswzy, line 53: The default max batch entries should be 1000 and the max batch size should be 2MB

server/service.go
```diff
@@ -452,6 +452,83 @@ func (s *Server) AddEntry(request *proto.AddEntryRequest, serverStream grpc.Serv
...
+			logger.Ctx(streamCtx).Warn("failed to send batch buffered response", zap.Error(sendErr))
+			return sendErr
+		}
+	}
+
+	// Phase 2: wait for each entry to be durably synced (entries flush in
+	// entry-id order, so reading in order adds no head-of-line latency beyond
+	// the natural durability order) and stream back its result keyed by
+	// entry_id.
+	for i, reqEntry := range request.Entries {
```
tinswzy: It should be returned all at once.

woodpecker/segment/batch_append_op.go
```diff
@@ -0,0 +1,145 @@
...
+	startRequestTime := time.Now()
+	_, err := cli.AppendEntries(ctx, first.bucketName, first.rootPath, first.logId, entries, resultChs)
+	if err != nil {
+		b.failNode(ctx, nodeIdx, serverAddr, err)
+		return
+	}
+
+	// Send succeeded: hand each op its per-replica result sink so its own
+	// callback waits for and processes the durability ack.
+	for i, op := range b.ops {
```
tinswzy: It should only require one goroutine to wait for the return of each op one by one, reducing the sudden increase in the number of coroutines

tinswzy:
## Benchmark — closed-loop throughput (consolidated stack)

Measured the full consolidated write path (group commit + server batch buffer-append + client batch-ack drain + coalesced LAC sync, group commit on by default) with a closed-loop driver (`conc` synchronous `Write`s, each blocking until durable).

**Setup:** 1 KB payload, closed-loop concurrency = 8000, 3-replica service mode, `MaxIntervalForService = 10 ms`, `MaxBatchEntries = 8192` / `MaxBatchBytes = 2 MB`, 3-node mini-cluster.

> Directional numbers, not a production benchmark: single laptop, and the 3 replicas share one physical disk, so fsync contention is pessimistic versus a real 3-disk deployment. Throughput here is bounded by `concurrency / latency` (Little's Law).

**Throughput** (4 consecutive runs):

| run | appends/s | MB/s | avg latency | fail |
|-----|-----------|------|-------------|------|
| 1 (warmup) | 30,152 | 29.4 | 260 ms | 0 |
| 2 | 31,515 | 30.8 | 250 ms | 0 |
| 3 | 31,265 | 30.5 | 252 ms | 0 |
| 4 | 31,130 | 30.4 | 252 ms | 0 |

Stable band **~31.1–31.5k appends/s, 0 failures**.

**Per-optimization contribution** (each stacked on the previous, same setup):

| stage | appends/s | delta |
|-------|-----------|-------|
| group commit (baseline) | 23,165 | — |
| + server batch buffer-append | 26,805 | +15.7% |
| + client batch-ack drain | 28,992 | +8.2% |
| + coalesced LAC sync | 32,519 | +12.2% |
| consolidated (this PR, re-measured) | ~31,300 | **~+35%** |

Cumulative **~+35–40%** over the group-commit baseline. (The gap between the 32.5k stacked figure and the ~31.3k consolidated re-run is machine-load noise — identical code.)

**Per-stage breakdown** (single run, µs/entry; server-side stages aggregate across the 3 replicas):

```
cli.ack.process        1377   client ack handling (wall-inflated by 8000-goroutine scheduling)
cli.ack.wait_durable   1093   waiting on durability
srv.wait_durable        999   server waiting on flush
srv.flush.fsync        6225   fsync (fixed cost, 832 flushes over the run)
srv.addentry_buffer      33   server buffer-append (~22 µs/entry before the batch-append change)
```

The remaining wall is the per-entry submit/ack path plus fixed fsync cost; the backend still has headroom (raw 2 MB-block disk throughput is ~323 MB/s, and the flush pipeline runs only a few percent busy at this rate). Pushing further would need structural batch submission on the client side, which is out of scope for this PR.

tinswzy:
## Benchmark — row-size × group-size sweep

Swept payload size against client group-commit size (`MaxBatchEntries`) on the full 3-replica service write path (closed-loop). Concurrency and write count are scaled down as rows grow so each cell writes a bounded amount and keeps in-flight bytes small. Added as `TestServiceRowSizeGroupSweep`.

**Setup:** 3-replica service mode, `MaxIntervalForService = 10 ms`, `MaxBatchBytes = 2 MB`, 3-node mini-cluster. Directional numbers (single laptop, 3 replicas share one disk).

| rowSize | group | QPS (app/s) | flow (MB/s) | avg latency | fail |
|---------|-------|------------:|------------:|------------:|-----:|
| 1 KB | off (1) | 4,239 | 4.1 | 1835 ms | 0 |
| **1 KB** | **on (1000)** | **31,673** | **30.9** | 250 ms | 0 |
| 10 KB | off (1) | 3,417 | 33.4 | 548 ms | 0 |
| **10 KB** | **on (1000)** | **10,653** | **104.0** | 173 ms | 0 |
| 100 KB | off (1) | 1,045 | 102.1 | 232 ms | 0 |
| 100 KB | on (1000) | 1,077 | 105.2 | 224 ms | 0 |
| 1 MB | off (1) | 143 | 139.9 | 405 ms | 0 |
| 1 MB | on (1000) | 121 | 118.5 | 479 ms | 0 |

**Reading the table:**

- **Small rows (1 KB): group commit is a ~7.5× win** (4.2k → 31.7k appends/s) — the per-entry round-trip is the dominant cost and batching amortizes it. This is the case the feature targets.
- **Medium rows (10 KB): still ~3.1×** (33 → 104 MB/s).
- **Large rows (100 KB, 1 MB): flat.** With a 2 MB batch cap a 100 KB row batches at most ~20 entries and a 1 MB row at most ~2, so group commit has little room; and a single large row already fills big flush blocks, so flow is already ~100–140 MB/s. (1 MB on-vs-off is within noise; batching 2 entries adds coordination for no gain.)

Net: group-commit value scales inversely with row size — large payloads are already block-efficient, small payloads are where batching pays off.

tinswzy:
## Batched ack responses — CPU saved & throughput

Follow-up on the batched `AddEntries` ack change (`repeated entry_id`; one frame per buffered batch and per flushed run instead of one frame per entry). Measured with `TestServiceWritePathBreakdown` (1 KB, conc=8000, 3-replica service mode; server stages aggregate across the 3 in-process replicas).

### Ack frames sent

| direction | per-entry (before) | batched (after) | reduction |
|-----------|-------------------:|----------------:|:---------:|
| Buffered | 359,505 | 19,917 | ~18× |
| Synced | 359,505 | 20,853 | ~17× |
| **total** | **719,010** | **40,770** | **~17.6×** |

### Server-side ack-send time (over the run)

| stage | before | after |
|-------|-------:|------:|
| send_buffered | 359,505 × 1.28 µs = 460 ms | 19,917 × 2.92 µs = 58 ms |
| send_synced | 359,505 × 7.29 µs = 2,621 ms | 20,853 × 9.01 µs = 188 ms |
| **total** | **~3,081 ms** | **~246 ms** |

Server ack-send work drops **~92% (~2.8 s over the run; ~8.6 µs → ~0.7 µs per entry-ack)**. Per-frame cost rises a bit (each frame now carries a run of ids), but ~18× fewer frames dominates. The client saves symmetrically — its per-node demux goroutine does ~41k `Recv()` instead of ~719k, each still a protobuf decode + channel handoff.

### Indirect throughput (interleaved A/B, same machine state, 3 rounds)

Single runs drift several % on a loaded laptop, so baseline (per-entry) and batched were run back-to-back, alternating, to cancel drift:

| round | per-entry | batched |
|-------|----------:|--------:|
| 1 | 29,773 | 31,149 |
| 2 | 29,999 | 30,456 |
| 3 | 31,529 | 31,855 |
| **avg** | **30,434** | **31,153** |

**+2.4 % appends/s, batched won all 3 rounds, 0 failures.**

### Why the throughput gain is small here (and where it isn't)

This regime is **fsync / latency-bound**: throughput = concurrency / latency (Little's Law), and the ack sends (~µs) sit off the latency critical path — they overlap with flushing (fsync ≈ 5.9 ms/flush dominates). So the ~2.8 s of freed CPU does **not** move the fsync wall; it only helps indirectly by reducing contention with the write path, which is what the +2.4 % reflects.

The saving converts to a larger throughput gain in **send-bound / CPU-tight regimes**, where the ack frames are actually the bottleneck rather than an overlapped side cost:
- very large batches, where the buffered round-trip previously did thousands of per-entry `Send`/`Recv` on the critical path;
- many concurrent streams / tiny rows, where per-frame gRPC framing is a real share of CPU.

Correctness is unchanged: acks are grouped only across entries that share one state (whole-batch Buffered, per-flush Synced), which is exactly what the ordered-log LAC-prefix invariant guarantees; failures still fail the RPC as a whole. Full segment/client/server/channel/log suites pass.

PR #211, opened 2026-07-03
perf(server): batch buffer-append in the AddEntries handler

tinswzy:
## What

The `AddEntries` (client-side group commit) handler buffered entries **one at a time**: for each entry it called `logStore.AddEntry -> processor.AddEntry -> writer.WriteDataAsync`, paying a per-entry span + metrics + logger, plus a `w.mu` (and buffer) lock acquisition. Under high concurrency that per-entry serial loop is the write-path gate.

This adds a batched path down the stack so a whole group-commit batch is buffered in a single call:

- `LogStore.AddEntryBatch` -> `SegmentProcessor.AddEntryBatch` -> (staged) `StagedFileWriter.WriteDataBatchAsync`
- one `w.mu` acquisition, one sync check, and one set of upper-layer plumbing per batch, while **reusing the proven per-entry `buffer.WriteEntryWithNotify`** so the `ExpectedNextEntryId` / buffer-roll semantics are unchanged.
- writers without a batch method (disk / object storage) fall back to a per-entry loop via a narrow `batchWriter` type assertion, so the `storage.Writer` interface is **not** widened.

## Why

A per-entry breakdown of the write path (1KB, service mode) showed the server buffer-append stage was the single biggest per-entry cost. Batching amortizes the per-entry span/metrics/logger and the lock churn across the whole batch.

## Results

3-replica service mode, 1KB payload, concurrency 8000, 10ms server flush interval, 2MB client batch cap (local mini-cluster, so the numbers are directional):

| metric | before | after |
|---|---:|---:|
| throughput | 23,165 appends/s | **26,805 appends/s (+15.7%)** |
| server buffer-append (per entry, amortized) | 22.6us | **0.72us** |

## Testing

- Full `./server/...` suite passes (incl. integration), plus `./server/storage/...` and `./server/processor/...`.
- The service write path runs with **0 failures**.

## Notes

- The staged writer has **no background flush ticker** — its interval flush is driven from the write path via `scheduleDelayedSyncCheck`. The batch method must call it too (as `WriteDataAsync` does), otherwise sub-threshold batches are never scheduled to flush and get stranded. Handled here.

## Depends on

Group commit / the `AddEntries` RPC (#208); this PR targets that branch.

issue: #202

### DIFF
--- a/mocks/mocks_server/mock_logstore.go
+++ b/mocks/mocks_server/mock_logstore.go
@@ -26,6 +26,31 @@ func (_m *LogStore) EXPECT() *LogStore_Expecter {
 }
 
 // AddEntry provides a mock function with given fields: ctx, bucketName, rootPath, logId, entry, syncedResultCh
+func (_m *LogStore) AddEntryBatch(ctx context.Context, bucketName string, rootPath string, logId int64, segmentId int64, entries []*proto.LogEntry, resultChs []channel.ResultChannel) ([]int64, error) {
+	ret := _m.Called(ctx, bucketName, rootPath, logId, segmentId, entries, resultChs)
+
+	if len(ret) == 0 {
+		panic("no return value specified for AddEntryBatch")
+	}
+
+	var r0 []int64
+	var r1 error
+	if rf, ok := ret.Get(0).(func(context.Context, string, string, int64, int64, []*proto.LogEntry, []channel.ResultChannel) ([]int64, error)); ok {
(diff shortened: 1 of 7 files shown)


tinswzy:
Note: GitHub auto-marked this **merged** because its commit `01d7b59` was pushed directly into the base branch `perf/client-group-commit` (i.e. folded into #208), **not** merged to master. The change ships as part of #208. Tracking there.

Issue #225, opened 2026-07-14
bug: robustness/correctness gaps in client-side group commit (#208) — retry channel type mismatch, unbounded LAC/batch waits, gRPC stream leak


tinswzy:
## Summary

Code review of the client-side **group commit** work (#202, landed in PR #208; builds on the per-replica fan-out #201 / PR #206) surfaced five correctness/robustness issues on the write path. Two are P0 (can cause production write failures / silent tail-reader stalls); three are P1 amplifiers that turn a transient object-storage hiccup into a false-timeout storm plus unbounded goroutine/stream growth.

Batching is **on by default** (`segmentAppend.maxBatchEntries: 1000`, `maxBatchBytes: 2MB`), so these paths are live in every default deployment under load.

This issue is the tracking umbrella. Fixes will land as **independent PRs** (P0 items separately; the two coupled P1 items together) — see *Fix plan* at the bottom.

---

### 1. Retry reuses a `LocalResultChannel` against a remote client → deterministic `ErrInternalError`

`BatchAppendOp.sendBatchToNode` stores a `NewLocalResultChannel` into `op.resultChannels[nodeIdx]` for **every** op, regardless of client type (`woodpecker/segment/batch_append_op.go:117-119`). That slot is also the retry path's channel slot. When any op in the batch later needs a retry, `HandleAppendRequestFailure` resubmits an `AppendRequestRetryOp` (`woodpecker/segment/segment_handle.go:441`), which runs the **single-entry** `sendWriteRequest`. That method only creates a fresh channel when the slot is `nil` (`woodpecker/segment/append_op.go:175`), so it **reuses the batch's `LocalResultChannel`** and passes it to the remote client's `AppendEntry`. For a `Buffered` response the remote client type-asserts `syncedResultCh.(*channel.RemoteResultChannel)` and returns `ErrInternalError` on mismatch (`woodpecker/client/logstore_client_remote.go:112-116`).

- Why it's fundamental (not just a guard): in remote mode the durable ack arrives asynchronously over the gRPC stream, and only `RemoteResultChannel` can attach/read that stream (`InitResponseStream` + `ReadResult`→`stream.Recv()`, `common/channel/remote_result_channel.go:132,187`). A `LocalResultChannel` has no stream to wire in, so it could never receive the ack anyway — the assertion just turns a silent hang into a fast error.
[...]

PR #228, opened 2026-07-14
fix(segment): bound the background LAC syncer so a hung node can't stall LAC forever

tinswzy:
## What

Fixes finding #2 of #225: a hung LAC-sync RPC permanently stalls LAC advancement for a whole segment.

`lacSyncLoop` is the single background LAC syncer for a writable segment and calls `syncLACToQuorumAsync(context.Background(), lac)` **synchronously** with a deadline-less context (`segment_handle.go:1216`). The fan-out sends `UpdateLastAddConfirmed` to each quorum node on that context (`:1307`), and the result-collection loop's only escape is `case <-ctx.Done()` — which never fires on a background context. So a node that is *connected but unresponsive* blocks the RPC forever, wedges the single `lacSyncLoop` goroutine, and newer LAC values only coalesce into `lacSyncLatest` with nobody draining them.

Effect: writes keep succeeding, but LAC stops advancing → all tail readers stall silently (for Milvus: build-index / query-sync quietly wedge) with no write-side signal, until the process restarts.

(The missing timeout predates this code, but the earlier design spawned a fresh `go syncLACToQuorumAsync` per ack round — a hang leaked one goroutine while LAC kept advancing. The single-syncer loop turned "leak one goroutine" into "permanent stall", which is what this fixes.)

## Fix

Bound the fan-out inside `syncLACToQuorumAsync` so a hung node can't wedge it:

```go
ctx, cancel := context.WithTimeout(ctx, lacSyncTimeout)
defer cancel()
```

On timeout the collection loop's existing `case <-ctx.Done(): return` unblocks (so `lacSyncLoop` proceeds to the next coalesced LAC), and `cancel()` releases any still-in-flight per-node RPCs — their late results land in the buffered `resultChan` (capacity `len(nodes)`), so no goroutine leaks. Fixing it inside the function covers every caller (the background `lacSyncLoop` and the read-only `go syncLACToQuorumAsync` path) at one point.

`lacSyncTimeout` is a package var defaulting to `30 * time.Second` — the same order as the existing `// TODO configurable` waits in the append path (`append_op.go`, `batch_append_op.go`). It's a `var` (not a literal) purely so the regression test can shrink it; a real config knob is left as the shared TODO.

## Test

`TestSyncLACToQuorumAsync_BoundedWhenNodeHangs` (regression): a mock node whose `UpdateLastAddConfirmed` blocks until its request context is cancelled, `lacSyncTimeout` shrunk to 200ms, and `syncLACToQuorumAsync` invoked with `context.Background()` (mirroring `lacSyncLoop`). Asserts it returns within 2s. Without the fix the call never returns (the background context is never done) and the test fails at its 2s guard; with the fix it returns in ~200ms.

Full `go test ./woodpecker/segment/` passes; `go build ./...` and `go vet` are clean.

## Scope

Only finding #2. Finding #1 landed in #226; findings #3/#4 (batch stream leak + shared read timeout) are the remaining follow-up PR.

Part of #225.

### DIFF
--- a/woodpecker/segment/segment_handle.go
+++ b/woodpecker/segment/segment_handle.go
@@ -1218,10 +1218,24 @@ func (s *segmentHandleImpl) lacSyncLoop() {
 	}
 }
 
+// lacSyncTimeout bounds a single LAC-sync fan-out so a hung node (connected but
+// unresponsive) can't wedge the single background LAC syncer forever. Package var
+// so tests can shrink it. TODO make configurable (like the 30s waits in the
+// append path).
+var lacSyncTimeout = 30 * time.Second
+
 // syncLACToQuorumAsync asynchronously syncs LAC to all quorum nodes
 func (s *segmentHandleImpl) syncLACToQuorumAsync(ctx context.Context, lac int64) {
 	ctx, sp := logger.NewIntentCtxWithParent(ctx, SegmentHandleScopeName, "syncLACToQuorumAsync")
 	defer sp.End()
+
+	// Bound the fan-out: the caller (lacSyncLoop) passes a deadline-less context,
+	// so without this a node that is connected but unresponsive would block the
+	// collection loop forever and stall LAC for the whole segment. On timeout the
+	// collection loop's ctx.Done() case unblocks and cancel() releases any
+	// still-in-flight per-node RPCs (their results land in the buffered channel).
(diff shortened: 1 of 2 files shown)


tinswzy:
Note: the `lacSyncTimeout` package var here is a deliberate hotfix seam. Making it (and the other scattered client→logstore timeouts) properly configurable and coherent is tracked as a follow-up audit in #229, which will subsume this var.

PR #235, opened 2026-07-15
enhance(server): disk-watermark backpressure + storage-expansion warning for WAL log store (#215)

### LINKED ISSUE #215: enhance: disk-watermark backpressure + storage-expansion warning for WAL log store
### What would you like to be added?

Add **disk-watermark-based backpressure and storage-expansion warnings** to the Woodpecker WAL log store, so that a WAL disk approaching full triggers graceful rate-limiting + an operator-facing warning **before** it hits `no space left on device` mid-append.

Concretely:

- **Soft watermark (warn):** when a data-disk / PVC crosses a configurable soft threshold (e.g. 80%), emit a rate-limited `WARN` log and a metric recommending storage expansion or retention tuning. This should be easy to alert on.
- **Hard watermark (backpressure):** when usage crosses a configurable hard threshold (e.g. 90–95%), apply backpressure to appends — reject/slow new writes with a **retriable** error (`ResourceExhausted`-style) instead of letting the write fail with `no space left on device` and cascade into a client-side `DEADLINE_EXCEEDED`.
- **Observability:** expose a per-node/per-replica disk-usage metric (used / total / free) and make the "truncated-but-not-yet-cleaned segment backlog vs. retention protection" state visible in logs/metrics, so this failure mode is distinguishable from a Milvus DML timeout without digging through server logs.
- Both thresholds should be configurable (with sane defaults) and default-on where safe.

This is the Woodpecker-side counterpart to milvus-io/milvus#51032. The Milvus maintainer plan noted there is: *"trigger backpressure-based rate limiting once disk usage reaches a configurable threshold, and emit warning messages recommending storage expansion."*

### Why is this needed?

A long TEXT-LOB DML stability workload against a Milvus cluster backed by Woodpecker (4 replicas, 30Gi PVC each, default `woodpecker.logstore.retentionPolicy.ttl: 72h`) filled the WAL disks and the client failed with:

```text
grpc StatusCode.DEADLINE_EXCEEDED
details: Stream removed (Deadline Exceeded)
```

Server-side, the real cause was WAL storage exhaustion:

```text
write /woodpecker/data/.../data.log: no space left on device
failed to open log writer
failed to fence segment ... no space left on device
append messages to wal failed ... context canceled
```

PVC usage at investigation time:

```text
woodpecker-0: 30G/30G, 0 free, 100%
woodpecker-1: 30G/30G, 0 free, 100%
woodpecker-2: 28G/30G, 2.2G free, 93%
woodpecker-3: 26G/30G, 4.0G free,  87%
```

The auditor found truncated segments but no eligible cleanup in the observed window (default 72h TTL). Because there is no disk-pressure warning or backpressure, the operational failure surfaces only as an opaque client-side upsert timeout, and there is no early signal telling the operator to expand storage or tune retention.

### Proposed scope / acceptance criteria

- [ ] Configurable soft + hard disk watermarks for the WAL log store (percentage and/or absolute-bytes headroom).
- [ ] Soft watermark → rate-limited WARN log + metric recommending expansion / retention tuning.
- [ ] Hard watermark → append backpressure returning a retriable/resource-exhausted error, so Milvus can surface it as a proper rate-limit rather than a hard write failure.
- [ ] Per-node WAL disk-usage metric exported (used/total/free/percent).
- [ ] Docs: guidance on sizing Woodpecker PVCs and tuning `retentionPolicy.ttl` for long-running DML stability.

### Anything else?

Related: earlier/eager cleanup of already-truncated WAL segments under disk pressure (when safe w.r.t. reader/checkpoint recovery guarantees) is a complementary lever — see reader/GC work in #209. This issue focuses on the **backpressure + warning** signal; retention-cleanup-under-pressure can be tracked separately if preferred.

Upstream reference: milvus-io/milvus#51032

tinswzy:
Closes #215. Woodpecker-side counterpart of milvus-io/milvus#51032.

## What

Adds **disk-watermark backpressure + storage-expansion warning** for the local WAL disk in service/local storage modes, so a node approaching full rate-limits new appends with a **retriable** error and warns the operator *before* hitting `no space left on device` mid-append.

- A `disk-watermark` `MaintenanceTask` (third sibling of the two existing tasks) samples `woodpecker.storage.rootPath` every `sampleInterval` (default 10s), computes the usage ratio as **`used/(used+free)`** (df semantics — `free` is statfs `Bavail`, so ext4 root-reserved blocks are accounted for), and stateless-evaluates **three stages**:
  - **Soft / warn** (`softThresholdRatio`, default 0.80): rate-limited WARN (1m cooldown) recommending PVC expansion or scaling out logstore nodes (bind new PVCs), plus `backpressure_state=1`.
  - **Throttle** (between soft and hard, `throttleEnabled` default true): each append is rejected with probability ramping **linearly 0%→100%** across the soft→hard band (e.g. 85% disk → 50% rejection with the default 0.80/0.90 band). Rejections are retriable, so writers slow proportionally to disk pressure instead of hitting a 0/100 cliff — this is the server-side rate-limit stage, and it also damps the unblock retry-storm sawtooth.
  - **Hard / block** (`hardThresholdRatio` default 0.90, **or** `minFreeBytes` floor default 1GiB, floor applies regardless of throttleEnabled): all new appends rejected. The task publishes a node-level rejection probability (`diskRejectBps` atomic, basis points); `AddEntry`/`AddEntryBatch` do one admission draw per call and reject with `werr.ErrLogStoreDiskPressure` (**code 2007, retryable=true**) so Milvus surfaces it as a transient rate-limit instead of a hard write failure / opaque `DEADLINE_EXCEEDED`.
- Only appends are gated. Fence/Complete/Compact/Clean/reads are never gated — they are the recovery path that frees disk.
- Recovery is automatic: the flag clears on the first sample below thresholds (no hysteresis by design; rejection is retriable, so boundary oscillation is graceful degradation).
- New metrics (all `woodpecker_server_...`): `logstore_disk_usage_ratio{node_id,path}`, `logstore_disk_free_bytes{node_id,path}`, `logstore_write_backpressure_state{node_id}` (0/1/2), `logstore_write_reject_probability{node_id}` (0..1, current throttle intensity), `logstore_write_rejected_total{node_id,reason="disk_pressure"}`.
- Config block `woodpecker.logstore.diskWatermarkPolicy` (default-on; warn-only = `throttleEnabled: false` + `hardThresholdRatio: 1.0` + `minFreeBytes: 0`; two-state warn→block = `throttleEnabled: false`; `enabled: false` = off). Registered only for service/local storage with a non-empty rootPath.
- Admin guide: `docs/admin-guides/disk-watermark.md` (PVC sizing, `retentionPolicy.ttl` tuning, alerting).
- Grafana: new collapsed row **"Disk Watermark — Backpressure"** (after System) in both server dashboards (`tests/docker/monitor/.../dashboard_server.json`, `tests/k8s/monitor/.../dashboard_server_k8s.json`): WAL disk usage ratio (with 0.80/0.90 threshold lines), free bytes, backpressure state (0/1/2, stepped), reject probability, rejected rate by reason.

## Error code note

The new error is **2007**, not the next-looking 2005: 2005/2006 were already taken (`ErrLogBeingDeleted`/`ErrMarkDeleteFailed`). Existing wire-protocol codes are untouched; 2007 is appended at the end of the logstore band.

## Behavior notes (verified during review)

- **Quorum interaction:** a single replica returning retriable `ErrLogStoreDiskPressure` is retried in place (`segment_handle.go: HandleAppendRequestFailure`) and tolerated by quorum accounting — the append completes at `Aq` acks via healthy replicas. If pressure on that replica persists past `MaxRetries`, the op enters the final-failure path and sets `rollingState=true`: a **segment-rolling side effect**, not an append failure. Expect increased segment rolling on a chronically pressured node.
- **Executor edge:** if a retry re-submission fails (executor closed/full), that op fast-fails with `ErrAppendOpRetrySubmitFailed` (non-retryable) — a shutdown/overload lifecycle edge independent of disk pressure.
- **Mixed-version safety:** old clients reconstruct errors from the wire `status.Retriable` bit (`werr.Error`), not a local code table, so unknown code 2007 is still treated as retriable. No client version floor.
- **Throttle-band retries:** a probabilistically rejected append retries and re-draws, so the throttle stage mainly adds latency/backoff pressure per replica (quorum still completes at Aq acks) — intended: it propagates a graduated slow-down signal upstream rather than precise QPS shaping; precise DML rate limiting stays in Milvus's quota system (milvus-io/milvus#51032).
- **Startup window:** on boot, appends are admissible for one scheduling hop before the first disk sample; on an already-full disk those writes fall through to the pre-existing `ErrFileWriterNoSpace` path (no regression), and the gate engages on the first tick.
- **Test suites:** mini-cluster/docker/integration suites run service mode with `TempDir` rootPath and inherit the default-on gate; it only fires ≥90% used, so CI/dev disks with headroom never trip it.

## Testing

- Unit: config defaults/validation; `GetDiskStats` (incl. non-existent path); error retryability/code/wire round-trip; watermark task level transitions, min-free floor, stat-error keeps state, warn cooldown; ramp mapping (`diskRejectBpsFor`: boundaries, throttle-off, empty band, floor precedence); admission gate (deterministic extremes + statistical n=10000 at p=0.5); gate rejection (both append paths) and registration matrix (service/local vs minio vs disabled).
- `go build ./...`, `go vet`, full `./server/`, `./common/{config,hardware,werr,metrics}` suites green.

### DIFF
--- a/common/config/configuration.go
+++ b/common/config/configuration.go
@@ -392,6 +392,7 @@ type DiskWatermarkPolicyConfig struct {
 	HardThresholdRatio float64         `yaml:"hardThresholdRatio"` // block watermark; default 0.90; >=1.0 means warn-only
 	MinFreeBytes       ByteSize        `yaml:"minFreeBytes"`       // absolute free floor that also blocks; default 1Gi
 	SampleInterval     DurationSeconds `yaml:"sampleInterval"`     // default 10s
+	ThrottleEnabled    bool            `yaml:"throttleEnabled"`    // probabilistic rejection between soft and hard; default true
 }
 
 type StorageConfig struct {
@@ -901,6 +902,7 @@ func getDefaultWoodpeckerConfig() WoodpeckerConfig {
 				HardThresholdRatio: 0.90,
 				MinFreeBytes:       ByteSize(1 * 1024 * 1024 * 1024), // 1Gi
 				SampleInterval:     DurationSeconds{Duration: Duration{duration: 10 * time.Second}},
+				ThrottleEnabled:    true,
 			},
 		},
 		Storage: StorageConfig{
--- a/common/config/configuration.go
+++ b/common/config/configuration.go
@@ -350,6 +350,7 @@ type LogstoreConfig struct {
 	ProcessorCleanupPolicy  ProcessorCleanupPolicyConfig `yaml:"processorCleanupPolicy"`
 	MaintenanceStrategy     MaintenanceStrategyConfig    `yaml:"maintenanceStrategy"`
 	NodeSelectionPolicy     NodeSelectionPolicyConfig    `yaml:"nodeSelectionPolicy"`
+	DiskWatermarkPolicy     DiskWatermarkPolicyConfig    `yaml:"diskWatermarkPolicy"`
 }
 
 // ProcessorCleanupPolicyConfig stores the background segment processor cleanup and shutdown configuration.
@@ -380,6 +381,17 @@ type NodeSelectionPolicyConfig struct {
 	LoadTTL            DurationSeconds `yaml:"loadTTL"`            // load older than this is treated as unknown; default 30s
 	MemSoftThreshold   float64         `yaml:"memSoftThreshold"`   // memory ratio above which memory escalates load; default 0.85
 	EWMAAlpha          float64         `yaml:"ewmaAlpha"`          // EWMA weight on newest sample; default 0.5
+}
+
+// DiskWatermarkPolicyConfig controls local-WAL-disk watermark warning and write
+// backpressure (issue #215). Active only for service/local storage (nodes that keep
+// WAL data on a local disk). Ratio semantics are used/(used+free), matching df.
+type DiskWatermarkPolicyConfig struct {
+	Enabled            bool            `yaml:"enabled"`            // master switch; default true
+	SoftThresholdRatio float64         `yaml:"softThresholdRatio"` // warn watermark; default 0.80
+	HardThresholdRatio float64         `yaml:"hardThresholdRatio"` // block watermark; default 0.90; >=1.0 means warn-only
+	MinFreeBytes       ByteSize        `yaml:"minFreeBytes"`       // absolute free floor that also blocks; default 1Gi
+	SampleInterval     DurationSeconds `yaml:"sampleInterval"`     // default 10s
 }
 
 type StorageConfig struct {
@@ -752,6 +764,17 @@ func (c *Configuration) validateLogstoreConfig() error {
 		}
 	}
 
+	dw := logstore.DiskWatermarkPolicy
+	if dw.Enabled {
+		if dw.SoftThresholdRatio <= 0 || dw.SoftThresholdRatio > dw.HardThresholdRatio {
+			return fmt.Errorf("diskWatermarkPolicy: require 0 < softThresholdRatio <= hardThresholdRatio, got soft=%v hard=%v",
+				dw.SoftThresholdRatio, dw.HardThresholdRatio)
+		}
+		if dw.MinFreeBytes < 0 {
+			return fmt.Errorf("diskWatermarkPolicy.minFreeBytes must be >= 0, got %d", dw.MinFreeBytes.Int64())
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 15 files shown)


xiaofan-luan:
@tinswzy 

why do we need to throttle?
I think we simply need to upgrade all data to S3 if there is no enough space, this can be more active since disk reachs 70%, and we fire alarm when disk space exceed 80% and let SRE handle the rest

tinswzy:
> @tinswzy
> 
> why do we need to throttle? I think we simply need to upgrade all data to S3 if there is no enough space, this can be more active since disk reachs 70%, and we fire alarm when disk space exceed 80% and let SRE handle the rest

The throttling was added as a temporary safeguard, mainly because auto-scaling and alerts aren't fully set up yet on cloud, and open-source deployments lack those guarantees entirely.

I totally agree this won't be needed once auto-scaling and instant disk cleanup after compact are ready. We can streamline this in future PRs—simply replacing it with a hard write-block when only a few GBs remain.

Issue #241, opened 2026-07-17
liveness: segment can stick in Completed forever when no node holds complete [0,LAC]; add reconstruct-or-refuse force-compact (follow-up to #237)


xiaofan-luan:
### Problem
Compaction validates that the compacting node's local data covers `[0, LAC]` (`server/storage/stagedstorage/writer_impl.go:2206`, `validateLACAlignment`). With `Aq < Es` (currently `es=3=wq, aq=2`), a segment can end up with **no single reachable node** holding the complete sequence — different nodes miss different entries, or the one complete node is offline. Then compaction fails on every node (`ErrInvalidLACAlignment`), the segment stays `Completed` forever, never gets an object-storage footer, never a `compacted.mark`, and `HasLocalSegmentData()` stays `true` — so the auto-decommission monitor (`server/service.go:907`) never converges. This relocates the exact hang #237 set out to remove.

Note the data is **not** lost: every committed entry is on ≥ `Aq` nodes, so the **union** covers `[0, LAC]` — it just can't be compacted from any single node's local view.

### Remediation
1. **Detect** — alert on the stuck-in-Completed signal (compaction-position stall). Tracked in #238.
2. **Force-compact admin op** — reconstruct `[0, LAC]` from the **quorum union** by reusing the existing quorum read `quorumReadBatch` (`woodpecker/segment/segment_handle.go:646`), which already fetches each entry from whichever replica has it, then compact + seal.

### Hard safety constraints on the force op
- **Never seal at a lowered LAC.** Accepting a single node's partial view and sealing short silently drops committed entries.
- **Reconstruct-or-refuse.** If the reachable union can't cover `[0, LAC]` (a node holding unique entries is down), refuse and report the exact missing entry ranges + which node holds them.
- Any truncate-to-shorter path must be a **separate, explicit, loudly-audit-logged break-glass flag**, not the default "force success."

### Deeper alternative
Make compaction itself quorum-aware — read-repair the full sequence from the quorum before merging — so it never gets stuck. The force command is the same logic on demand, and a fine first step.

Ref: PR #237 review comment https://github.com/zilliztech/woodpecker/pull/237#issuecomment-5006805687

tinswzy:
Thanks for raising this. We re-audited the completion and compaction paths against the latest `master`, and found that the liveness concern was valid, although the exact replica-divergence model in the issue needs one clarification.

### Replica data is a contiguous prefix

[...]

PR #243, opened 2026-07-24
fix(compaction): refuse to compact a replica behind the confirmed LastEntryId

### LINKED ISSUE #242: bug(compaction): a behind/empty replica can seal a segment short — validation uses the node's local footer LAC, not the coordinator-confirmed LastEntryId
## Summary

Segment compaction (`Completed` → `Sealed`) is run on **whichever quorum node answers first**
(`compactSegmentQuorum` in `woodpecker/segment/segment_handle.go`), and the client accepts that
node's result. The compact request carries **no coordinator-confirmed target**, so a node decides
whether it may compact using only its **own local footer LAC**. A replica that never received the
segment's data (or fell behind) can therefore be accepted, producing a `Sealed` segment whose
object-storage data is **shorter than the segment's confirmed `LastEntryId`** — an
already-acknowledged entry is silently dropped.

Observed in production as a **permanent reader hang**: the sealed segment has no data for the
dropped entry, direct-read returns `ErrEntryNotFound` forever, and the WAL scanner never advances.

## Impact

- **Data loss / silent skip:** an entry acknowledged to the writer (present on ≥ `Aq` nodes) is not
  in the sealed object. For a shared WAL log (e.g. a rootcoord DML channel) this is a lost DML/DDL
  record.
- **Availability:** the log stops being consumable at the dropped position — a continuous stream of
  `"direct read batch failed" error="no record extract"` and a consumer that never advances.
  Restarting the consumer does not help (it resumes from the same checkpoint and re-hangs).

## Root cause

`compactSegmentQuorum` iterates the quorum nodes in order and returns the **first** node whose
`SegmentCompact` does not error, with no check that the sealed result reaches the segment's
confirmed `LastEntryId`:

```go
// woodpecker/segment/segment_handle.go
compactSegMetaInfo, compactErr := cli.SegmentCompact(ctx, s.bucketName, s.rootPath, s.logId, s.segmentId)
if compactErr != nil { ... continue }
return compactSegMetaInfo, nil   // accepts the first node's result
```

Server-side, the node only ever validates against **its own local footer LAC**:

- `validateLACAlignment` (`server/storage/stagedstorage/writer_impl.go`) checks
  `lastEntryID >= footer.LAC`, where `footer.LAC` is **this node's** recovered footer LAC.
- #237's empty-segment guard refuses only when **this node's** `recoveredFooter.LAC >= 0`.

Neither the node nor the compact RPC is told the **coordinator-confirmed** `LastEntryId` (which the
client holds in the segment metadata). So a replica whose local footer LAC does not reflect the
confirmed LAC — because it never got the data, or was finalized from its own stale local view — is
judged "complete enough" against itself and seals short.

## Why #237 does not fully prevent this

#237 (now in master) made empty compaction upload a `footer.blk` and added a guard that refuses when
`recoveredFooter.LAC >= 0` with zero local blocks. That removes the *Sealed-without-footer* hang for
genuinely empty segments, but the guard (and `validateLACAlignment`) still key off the node's
**local** footer LAC, so two gaps remain:

1. **Empty replica whose local footer LAC is `-1`.** A node that received no data at all can be
   finalized/fenced from its local view (`lastEntryID = -1`), so `recoveredFooter.LAC < 0`. #237
   then treats it as "genuinely empty" and **uploads an empty compacted footer**. Direct-read now
   finds a compacted footer and returns **EOF → advance**, so the consumer **silently skips** the
   confirmed entry — the loud hang becomes silent data loss.
2. **Partial replica with a stale local footer LAC.** A node holding `[0, k-1]` whose local
   `footer.LAC = k-1` passes `validateLACAlignment` (`lastEntryID(k-1) >= LAC(k-1)`) and seals at
   `k-1`, dropping the confirmed entry `k`.

The only authoritative value is the coordinator's `LastEntryId` (segment metadata), which the node
never sees.

## Reproduction (production, v0.1.33, identifiers redacted)

One segment on a rootcoord DML WAL log (`logId=1`, `segId=11`, quorum size 3). node-B is a replica
that received no data for this segment (append to it timed out):

Client side:
```
"Calculated LAC from quorum fence responses"        lac=1
"Validating segment state for compaction"           LastEntryId=1
"Starting quorum compaction operation"              nodeCount=3 lastEntryId=1
"Attempting compaction on node"                     node=node-B nodeIndex=1   # first node = the empty one
"Segment compaction succeeded on node"              successfulNode=node-B     # accepted
"finish compact segment to sealed"                                            # metadata stays LastEntryId=1
```
node-B (server side):
```
"fence operation completed successfully"            lastEntryID=-1            # node-B holds nothing
"starting staged file segment compaction"
"no blocks to compact"  ->  compaction "succeeded"                            # bypassed the completeness check
```
Result: the segment is `Sealed` with metadata `LastEntryId=1`, but object storage holds no data for
it. Every subsequent read:
```
[WARN] segment/segment_handle.go "direct read batch failed" logId=1 segId=11 from=1 error="no record extract"
```
loops forever — `server/storage/objectstorage/reader_impl.go` returns `ErrEntryNotFound` ("no record
extract") and `woodpecker/log/log_reader.go` treats it as "wait and retry until EOF". (This was on
`v0.1.33`, before #237; on current master the same short-seal path is still reachable via the two
gaps above, and gap 1 converts the hang into a silent skip.)

## Proposed fix

Give the node the coordinator-confirmed target so it can judge against the authoritative value
instead of its own local footer:

- Add `expected_last_entry_id` to `CompactSegmentRequest`.
- Thread it from `segmentMeta.LastEntryId` through `compactSegmentQuorum` → `SegmentCompact` →
  `logStore.CompactSegment` → `segmentProcessor.Compact`.
- In `segmentProcessor.Compact`, **before** merging/uploading, refuse when
  `writer.GetLastEntryId() < expectedLastEntryId` with a new retryable error
  (`ErrSegmentCompactionDataBehind`). The client's existing per-node error handling then moves on to
  the next quorum node.
- `expected_last_entry_id = -1` (no expectation) preserves current behavior for manual/force
  compaction and genuinely-empty segments.

This covers both an empty node (`-1 < expected`) and a partial/behind node (`k-1 < expected`), does
not depend on the node's local footer being correct, and directly enforces #241's *"never seal at a
lowered LAC"* constraint at compaction-request time. Because `LastEntryId = N` implies a write
quorum already acknowledged entry `N`, at least one node passes on the same pass; if none is
transiently caught up, compaction fails and retries later instead of sealing short.

(Prevents new occurrences only — a segment already sealed short is not repaired by this fix; the
data still lives on the caught-up replicas' local `data.log` until GC, so recovering an
already-affected segment is a separate step.)

## Relation to #241

#241 is the **liveness** counterpart: after #237, if the per-node completeness check rejects *every*
reachable node, the segment sticks in `Completed` forever, and it proposes reconstruct-from-quorum-
union (or refuse) as the recovery. This issue is the **safety** counterpart: the per-node check
validates against the node's *local* footer LAC rather than the coordinator's `LastEntryId`, so a
behind/empty node can still be **wrongly accepted** and seal short. The `expected_last_entry_id`
fix here is the concrete enforcement of #241's *"never seal at a lowered LAC"* constraint and a
prerequisite for any reconstruct-or-refuse logic to be safe.

## Affected versions

Present on `master` (including after #237) and on released versions; observed in production on
`v0.1.33`.

tinswzy:
Fixes #242. Related to #241.

## Problem

Segment compaction (`Completed` -> `Sealed`) previously let the selected logstore
node decide the published boundary from its local state. A lagging replica could
therefore publish a compacted segment shorter than the coordinator-confirmed
`LastEntryId`, silently dropping acknowledged entries and leaving readers unable
to advance.

Compaction also needs to handle the opposite case: a replica may contain a local
tail beyond the confirmed boundary when a write reached only that replica. Those
entries must not be published merely because they exist locally.

### Compact exactly to the coordinator-confirmed boundary

- Thread the authoritative `segmentMeta.LastEntryId` through the compaction call.
- Add `CompactSegmentWithExpected` with a dedicated request carrying
  `expected_last_entry_id`.
- Keep the legacy `CompactSegment` RPC only to fail old clients explicitly with
  `unsupported version`; upgraded clients never fall back to it.
- Reject a replica whose finalized local data does not cover the expected entry.
- When local data extends beyond the expected entry, decode the local records,
  crop the unconfirmed tail, repack blocks according to the compaction block-size
  policy, and publish only through the expected entry.
- Require the compacted block indexes' maximum entry ID to equal the expected
  entry ID before publishing the footer.
- Treat an existing remote compacted footer as idempotent success only when its
  LAC exactly equals the expected entry ID.
- Allow `-1` only for a genuinely empty finalized segment; empty compaction writes
  only a compacted footer with `LAC = -1`.

### Fail closed across mixed RPC versions

- Old client -> new server: the legacy RPC returns a clear upgrade-required error
  and performs no compaction.
- New client -> old server: gRPC returns `Unimplemented`; the client reports the
  failure and does not fall back to the unsafe legacy RPC.

This intentionally requires the client/coordinator and server to be upgraded
before compaction resumes; it never guesses a missing proto3 scalar as entry `0`.

### Admin force compaction

`ForceCompact` and `wp logstore compact` require an explicit
`expected_last_entry_id`. This value is an operator assertion and is treated as
authoritative by the logstore; it is not cross-checked against coordinator
metadata. The API comments and CLI help explicitly require obtaining the exact
confirmed `LastEntryId` from coordinator metadata and warn that an incorrect value
can publish an invalid footer. This maintenance path should not be run concurrently
with normal auditor compaction.

## Tests

Coverage includes:

- lagging, caught-up, and locally-ahead staged replicas;
- record-level cropping and block repacking through the exact expected entry;
- strict remote-footer LAC matching and empty-segment footer generation;
- legacy RPC rejection without invoking compaction;
- new-client `Unimplemented` behavior without legacy fallback;
- required admin/CLI expected-entry input and payload forwarding.

Commands run locally:

```text
go test ./server ./woodpecker/client ./cmd/wpcli/cmd ./common/http/management -count=1
```


czs007:
Adversarial review surfaced six issues in the compaction LAC-guard change — one high-severity data-loss path, two medium correctness/compatibility gaps, and three low-severity operability and test-coverage items.

## High

**`server/storage/stagedstorage/writer_impl.go:1387` — Idempotent fast path skips the LAC guard, allowing healthy replicas to be authorized to delete local data**

The idempotency shortcut (checked at `:1387`, returning at `:1392`) sits *before* the new guard at `:1410`, and returns `totalSize, nil` without ever comparing `existingFooter.LAC` against `expectedLastEntryId`. Critically, `readRemoteFooter` (`:1939`) reads the *shared* `footer.blk` from object storage (`StatObject`/`GetObject` on `w.bucket`), not node-local state — so once any single node publishes a short compacted footer, every other node's `Compact` takes this branch and reports success, short-circuiting the new guard cluster-wide. Because compacted-mark distribution is gated on `State == Sealed` (`woodpecker/log/log_auditor.go:98`) and the mark authorizes reclaiming the local `data.log` (`server/logstore.go:66-68`), the consequence is real data loss. Failure scenario: node B publishes a compacted footer with `LAC=3` via `ForceCompact(-1)` (`server/writer_registry.go:77`) or a short-footer path while `metadata.LastEntryId=5` → the next compaction round on A and C hits the idempotent branch and returns success → the segment is set to `Sealed` → the mark is distributed → healthy replicas that still hold entries 4 and 5 delete their `data.log`, turning a manually recoverable state into permanent loss.

*Suggestion:* inside the idempotent branch, additionally verify `existingFooter.LAC >= expectedLastEntryId` when `expectedLastEntryId >= 0`, and reject otherwise so the segment stays in `Completed`. (raised by chyezh, bigsheeper)

## Medium

**`server/storage/stagedstorage/writer_impl.go:2365` — The guard compares `lastEntryID` but the published value is `footer.LAC`; the two can diverge and truncate data**

The new guard compares `w.lastEntryID.Load()` (`:1411`), and `validateLACAlignment` only checks `lastEntryID >= lac` (`:2365`) and `lac <= maxBlockLastEntryID` (`:2379`) — a *too-small* `footer.LAC` always passes. What actually gets written into the compacted footer is `w.recoveredFooter.LAC` (`:2331-2338` → `lac := footer.LAC` at `:2338` → received at `:1466` → `uploadCompactedFooter(ctx, newBlockIndexes, fileSizeAfterCompact, lac)` at `:1508`). Divergence is reachable in the current code: `Finalize` writes `LAC: lac` and sets `w.recoveredFooter = footer` (`:1235`/`:1248`) without truncating `lastEntryID`; re-finalizing an already-finalized writer fast-returns at `:1167-1170` and never rewrites `footer.LAC`; and `Fence` on a writer whose upload has completed returns the in-memory `w.GetLastEntryId(ctx)` (`:1337-1338`), not `footer.LAC`. Failure scenario: the first `Finalize(3)` writes `footer.LAC=3`, then `UpdateSegmentMetadata` fails (`woodpecker/segment/segment_handle.go:1638-1645` returns the error without persisting `LastEntryId`) → a second round fences that node and gets `5`, so `metadata.LastEntryId=5` → `Complete(5)` fast-returns because the writer is already finalized, leaving the footer at `3` → the guard passes (`5 >= 5`) and `validateLACAlignment` passes (`5 >= 3`) → a compacted footer with `LAC=3` is published → readers truncate at the footer LAC, and reading entry 4 returns `ErrEntryNotFound` and hangs forever — precisely the symptom this PR is meant to fix.

*Suggestion:* change (or extend) the guard to check `w.recoveredFooter.LAC >= expectedLastEntryId`. It is strictly stronger and will not falsely reject nodes that completed normally. (raised by bigsheeper)

**`proto/logstore.proto:185` — proto3 scalar has no presence: old servers silently ignore `expectedLastEntryId` and clients cannot detect it**

`int64 expected_last_entry_id = 5;` is a proto3 scalar without `optional`, so it has no field presence. A server build that predates the field drops it as an unknown field, its `CompactSegment` implementation has no such parameter, and it completes compaction and returns success as before — the guard is silently disabled and the client has no protocol-level way to tell. The client also has no fallback validation available: `segmentProcessor.Compact` populates only `State`/`SealedTime`/`Size` (`server/processor/segment_processor.go:687-691`), so `LastEntryId` is always `0`, and `woodpecker/segment/segment_handle.go:1791-1792` copies back only `SealedTime` and `Size`. Failure scenario: during a rolling upgrade of a mixed-version cluster, a request lands on an old server node; the client believes the guard is active while behavior silently reverts to the pre-fix path (short footer accepted → `Sealed` → mark distributed → data loss). Note that marking the field `optional` is not sufficient — presence only helps a receiver that recognizes the field, old servers still discard it as unknown, and the client still receives no acknowledgement.

*Suggestion:* add a capability gate (server-side capability negotiation), or echo the effective `expected_last_entry_id` back in the response so the client can validate it. (raised by chyezh, bigsheeper)

## Low

**`server/writer_registry.go:77` — `ForceCompact` bypasses the LAC guard via `sp.Compact(ctx, -1)` with no explicit unsafe confirmation or audit log**

`ForceCompact` passes `-1`, which disables authoritative LAC validation entirely (the guard only applies when `expectedLastEntryId >= 0`, `server/storage/stagedstorage/writer_impl.go:1410`), so a lagging replica can publish a short footer with no authoritative target. Combined with issue #1, the impact is amplified: that footer lives in shared object storage and causes every other node's compaction to return success immediately. The entry point is exposed via an admin HTTP endpoint (`common/http/server.go:234-238`, registered from `cmd/main.go`), and its sibling `ForceFence` takes an explicit `reason string` parameter (`server/storage/writer_registry.go`), while `ForceCompact` has neither a `reason`/`unsafe` parameter nor any `Warn`-level audit logging — only a single inline comment.

There is a genuine disagreement here worth recording: chyezh suggests requiring an authoritative LAC for force-compaction, or rejecting it when no authoritative target exists. bigsheeper disagrees with that fix and argues for keeping this at low severity: this PR turns lagging replicas into a hard rejection, but `calculateLAC` (`woodpecker/segment/segment_handle.go:1358-1384`) only guarantees that `aq` nodes reached the LAC. If those nodes are permanently lost, every remaining replica has `localLast < expected` → `compactSegmentQuorum` rejects across the board → the segment is stuck in `Completed` forever, with the auditor spinning on it each round (`woodpecker/log/log_auditor.go:58` only handles `Completed`). In that state `ForceCompact(-1)` is the only operational escape hatch that can advance the segment to `Sealed`.

*Suggestion:* keep the `-1` opt-out, but require an explicit unsafe confirmation (e.g. a `reason`/`unsafe` parameter matching `ForceFence`) and emit a `Warn`-level audit log when the guard is bypassed. (raised by chyezh, bigsheeper)

**`woodpecker/segment/segment_handle.go:1768` — No test pins the source of the authoritative expected LAC**

`s.compactSegmentQuorum(ctx, quorumInfo, currentSegmentMeta.Metadata.LastEntryId)` determines where the guard's authoritative value comes from, but no test covers this argument. The two tests that pin `expectedLAC` exactly (`TestCompactSegmentQuorum_AllNodesBehindFails` at `:5447`/`:5450` and `TestCompactSegmentQuorum_SkipsLaggingNode` at `:5521`/`:5526`) call `impl.compactSegmentQuorum` directly (`:5461`, `:5537`), bypassing `:1768`. Every test that goes through `segmentHandleImpl.Compact()` — `TestCompact_Success` (`:4155`, mock at `:4204`), `TestCompact_SealedTimeFromCompactResponse` (`:4281`), `TestCompact_CompactionFails` (`:4330`), `TestCompact_AlreadyCompacting` (`:4397`), `TestCompact_DoesNotNotifyInline` (`:4497`) — uses `mock.Anything` for the 6th matcher. Failure scenario: replacing `currentSegmentMeta.Metadata.LastEntryId` with the constant `-1` (equivalent to disabling the guard globally) or `0` leaves the entire unit suite green, while the guard is silently off in production and behavior reverts to the pre-fix path.

*Suggestion:* the fix is small — `TestCompact_Success`'s `completedMeta` already carries `LastEntryId: 10` (`:4161`), so change the 6th matcher at `:4204` from `mock.Anything` to `int64(10)`. (raised by bigsheeper)

**`woodpecker/segment/segment_handle_test.go:5540` — New `SkipsLaggingNode` assertion checks a field the server never populates, giving false confidence**

The closing assertion `assert.Equal(t, expectedLAC, result.LastEntryId)` relies on the hand-constructed `fullMeta` in the test itself (`:5524`, `&proto.SegmentMetadata{SegNo: 1, Size: 100, LastEntryId: expectedLAC}`), but the real server path never returns metadata carrying `LastEntryId`: `segmentProcessor.Compact` populates only `State`/`SealedTime`/`Size` (`server/processor/segment_processor.go:687-691`), leaving `LastEntryId` at `0`. Per its comment the assertion is meant to verify that the sealed segment really covers the confirmed LAC; in practice it only verifies that the mock's return value is passed through. Failure scenario: a server that seals a short footer and returns `State=Sealed` metadata would not trip this assertion at all — and the assertion actively misleads readers into thinking the client validates `LastEntryId` on the response, masking the missing client-side fallback described above.

*Suggestion:* drop the assertion, or replace it with an explicit TODO noting that the response does not carry `LastEntryId`, so it does not mask the missing capability gate. (surfaced during verification)

czs007:
## Summary

Adversarial review surfaced two high-severity correctness/protocol gaps in the new compaction data-behind guard, plus four medium-severity issues covering protocol encoding, an admin opt-out path, guard ordering, and a test-coverage hole.

---

### High

**`server/storage/stagedstorage/writer_impl.go:1466` — The guard validates `lastEntryID`, but the LAC actually published still comes from this node's possibly stale local footer**

The new guard only compares `w.lastEntryID` against `expectedLastEntryId` (:1410-1412). The LAC that is written into the compacted footer and made visible to every reader comes from `validateLACAlignment` → `w.recoveredFooter.LAC` (:2338, :1508), and that function only checks `lastEntryID >= lac` — never `lac >= expectedLastEntryId`. Any replica satisfying `recoveredFooter.LAC < expected <= lastEntryID` passes both checks and seals a segment whose LAC is below the coordinator-confirmed value.

Reproducible path: across multiple `doComplete` rounds in `completionManager.executeWithRetry`, round 1 lands node A's footer at `LAC=3` but fails quorum; round 2 raises `lac` to 5, yet A returns early because it is already finalized (:1168-1171), leaving its footer at 3 while `meta.LastEntryId=5`. If A is picked first during compaction, `uploadCompactedFooter(lac=3)` commits. The blocks for entries 4/5 are uploaded, but the staged reader clamps the readable range to the footer LAC (`reader_impl.go:1026-1027`, :1117-1122), so those entries become unreadable; once the local `data.log` is reclaimed, the fileless writer does `w.lastEntryID.Store(footer.LAC)` (:384) and the node permanently reports a lowered last entry id — exactly the regression issue #241 is meant to prevent.

*Suggestion:* after :1466, reject when `expectedLastEntryId >= 0 && lac < expectedLastEntryId` (or, when the data is verifiably complete, raise the published LAC to `expectedLastEntryId`; rejecting is the more conservative option). *(raised by bigsheeper)*

**`server/processor/segment_processor.go:687` — Compact response omits `LastEntryId`, leaving the client with no defense-in-depth check**

The returned `SegmentMetadata` (:687-691) sets only `State`/`SealedTime`/`Size`, so `LastEntryId` is always the proto3 zero value. On the client side, `compactSegmentQuorum` returns the response unvalidated (`segment_handle.go:1537`), and the main Compact flow copies back only `State`/`SealedTime`/`Size` (:1789-1792). During a rolling upgrade an older logstore drops the new `expected_last_entry_id` field into `unknownFields` and ignores it entirely, so the first-responder-wins ordering can select a lagging replica with nothing on the client able to detect it. This is the other end of the same hole as the issue above: one side never checks, the other checks the wrong value. Side effect: the three log statements at :1535, :1811 and :1828 print `compactSegMetaInfo.LastEntryId`, which is permanently `0` and misleading during triage. Note that `meta.LastEntryId` itself is not corrupted (:1789 clones the coordinator meta), so the impact is "undetectable" rather than "meta overwritten".

*Suggestion:* populate the LAC in the Compact response and add a client-side `>= expected` validation; worth landing together with this PR. *(raised by chyezh, bigsheeper)*

---

### Medium

**`proto/logstore_vtproto.pb.go:2892` — Non-optional proto3 field makes "absent" indistinguishable from `0`, causing empty segments to be misjudged as behind during rolling upgrade**

`-1` does encode correctly (`if m.ExpectedLastEntryId != 0` is true for `-1`, :2892-2896) and the guard only engages on `expectedLastEntryId >= 0` (`writer_impl.go:1410`), so `-1` takes the skip-check path. The real defect is that a proto3 non-optional field cannot distinguish "unset" from `0`: an old client omits field 5 entirely, the new server decodes `0`, and the guard then requires `localLast >= 0`. For a genuinely empty segment (coordinator `LastEntryId = -1`, replica-local `lastEntryID` defaulting to `-1`), every replica returns `ErrSegmentCompactionDataBehind`, so compaction fails for the whole upgrade window and the segment stays stuck in `Completed` while `log_auditor` retries fruitlessly.

*Suggestion:* declare `expected_last_entry_id` as `optional` so the server can tell "old client, unset → skip check" from "new client, explicit 0 → check". *(raised by bigsheeper)*

**`server/writer_registry.go:77` — `ForceCompact` hardcodes `-1`, leaving the admin endpoint able to seal short segments**

`ForceCompact` passes `expectedLastEntryId = -1`, fully disabling the new guard on this path — the residual entry point for this PR's primary failure scenario, exposed via the HTTP `/admin` compact handler (`common/http/server.go:234-238`). An operator manually compacting a lagging or empty replica can still seal a short segment. Passing the real coordinator-confirmed value is not directly feasible here, since `logStore` only receives `logID`/`segmentID` and has no coordinator meta channel.

*Suggestion:* one of (a) have the admin handler accept an explicit `expected_last_entry_id` query parameter and pass it through, (b) look up the segment meta on the `logStore` side and fill it in, or (c) at minimum require an explicit `force=true` before allowing `-1`, so a mistaken operation cannot silently land on the unguarded path. The existing comment documents this as a deliberate opt-out, so this is a design trade-off rather than an oversight — hence medium. *(raised by bigsheeper)*

**`server/storage/stagedstorage/writer_impl.go:1385` — Remote-footer idempotency branch returns success before the new guard runs**

The idempotency branch at :1385-1393 sits 25 lines ahead of the new guard at :1410 and returns `(totalSize, nil)` without any data-behind validation, on which the client transitions the segment to `Sealed`. Triggering it requires a short compacted footer that already existed before the upgrade (the new guard closes the path that creates new short footers), hence medium rather than high. Returning success here is semantically correct — the remote footer is the global commit point and should not be overridden by one node's local progress.

*Suggestion:* compare `existingFooter.LAC` against `expectedLastEntryId` at that point and error/alert when `LAC < expected`, instead of either succeeding silently or rejecting outright. *(raised by chyezh, bigsheeper)*

**`server/storage/stagedstorage/writer_impl_test.go:1348` — Missing case for `expected > footer LAC` with `lastEntryID >= expected`**

`TestStagedFileWriter_Compact_RealWrites_ProceedsWhenCaughtUp` establishes `footer LAC=9` via `Finalize(9)` and then calls Compact with `expected ∈ {9, 5}` — both on the `expected <= footer LAC` side. The combination that actually seals short (`expected > footer LAC` and `lastEntryID >= expected`, i.e. the first high-severity issue) is not covered at all. The other new cases (`RefusesEmptyReplicaBehindExpected`, `RefusesPartialReplicaBehindExpected`, `MeetsExpectedPassesBehindCheck`, `NoExpectationSkipsBehindCheck`) only exercise the `lastEntryID` vs `expected` relationship and never the `footer LAC` vs `expected` one.

*Suggestion:* add a case that writes and `Finalize(3)` (footer `LAC=3`), manually raises `lastEntryID` to 5, then calls `Compact(ctx, 5)` and asserts no upload occurs using a strict mock with no `PutObject` expectation. *(raised by bigsheeper)*

czs007:
## Re-review of `54f4718a9d73`

This re-review re-examined every earlier finding on this unchanged commit: all twelve are **Confirmed**, none retracted, and no new critical/high issues were found beyond them.

### Adjudication of earlier findings

- **[high] `server/storage/stagedstorage/writer_impl.go:1387` — idempotent branch does not validate the existing compacted footer's LAC** — **Confirmed**: the remote-footer fast path still returns success at `:1392` before the guard at `:1410`, and `readRemoteFooter` reads shared object storage, so a short footer published by any node is accepted cluster-wide.
- **[medium] `server/storage/stagedstorage/writer_impl.go:2365` — guard compares `lastEntryID` while `footer.LAC` is what gets published** — **Confirmed**: `validateLACAlignment` only enforces `lastEntryID >= lac` and `lac <= maxBlockLastEntryID`, so a too-small `footer.LAC` passes unconditionally.
- **[medium] `proto/logstore.proto:185` — proto3 scalar has no presence; old servers silently ignore `expectedLastEntryId`** — **Confirmed**: the field is a bare non-optional `int64`, and the response carries no `LastEntryId` for the client to cross-check.
- **[low] `server/writer_registry.go:77` — `ForceCompact` bypasses the guard via `sp.Compact(ctx, -1)` with no unsafe acknowledgement or audit log** — **Confirmed**: the callback type carries neither a reason nor an expected LAC, unlike the sibling `ForceFence`.
- **[low] `woodpecker/segment/segment_handle.go:1768` — the production source of the expected LAC is not pinned by any test** — **Confirmed**: every `Compact()` test passes `mock.Anything` for the new argument, and the exact-value tests bypass this line.
- **[low] `woodpecker/segment/segment_handle_test.go:5540` — `SkipsLaggingNode` asserts a field the server never populates** — **Confirmed**: the assertion passes only because the mock constructs `fullMeta.LastEntryId` itself.
- **[high] `server/storage/stagedstorage/writer_impl.go:1466` — the new guard checks `lastEntryID`, but the published LAC still comes from this node's possibly stale footer** — **Confirmed**: same root cause as the `:2365` item, with a reachable retry path via repeated `Finalize` fast-returns.
- **[high] `server/processor/segment_processor.go:687` — the Compact response omits `LastEntryId`, leaving the client with no defence in depth** — **Confirmed**: the response fills only state, sealed time, and size, and the client accepts the first success without validation.
- **[medium] `proto/logstore_vtproto.pb.go:2892` — absence and 0 are indistinguishable; empty segments are misjudged as lagging during rolling upgrade** — **Confirmed**: an old client omits field 5, the new server decodes 0, and the `>= 0` guard rejects every genuinely empty replica.
- **[medium] `server/writer_registry.go:77` — `ForceCompact` hard-codes `-1`, so the admin endpoint can still seal a short segment** — **Confirmed**: same bypass as the low-severity item above; the concrete short-segment scenario remains reachable.
- **[medium] `server/storage/stagedstorage/writer_impl.go:1385` — the remote-footer idempotent branch returns success ahead of the new guard** — **Confirmed**: the narrower pre-existing-footer scenario of the `:1387` item still holds.
- **[medium] `server/storage/stagedstorage/writer_impl_test.go:1348` — no test covers `expected > footer LAC` with `lastEntryID >= expected`** — **Confirmed**: both expected values in the caught-up test sit at or below the footer LAC of 9.

---

## Findings from this round

### High

**`server/storage/stagedstorage/writer_impl.go:1387` — remote-footer idempotency fast path returns success before the guard, without validating the existing footer's LAC**

The idempotent branch hits at `:1387` and returns `totalSize, nil` at `:1392`, ahead of the new guard at `:1410`, never comparing `existingFooter.LAC` against `expectedLastEntryId`. `readRemoteFooter` (`:1933`) fetches `footer.blk` from shared object storage via `StatObject`/`GetObject` (`:1939`/`:1947`) at key `rootPath/logId/segmentId/footer.blk` (`:1973`) — this is cluster-wide state, not node-private. So once any path publishes a short compacted footer (ForceCompact passing `-1`, a footer predating the guard, or the stale-footer path below), every node's `Compact` returns success unconditionally. The client then marks the segment `Sealed` and writes metadata back (`woodpecker/segment/segment_handle.go:1789-1797`), which drives compacted-mark distribution and local `data.log` reclamation, ultimately authorizing replicas that still hold the full data to delete it — permanent loss of acknowledged entries.

*Suggestion:* before returning success from the idempotent branch, compare the existing footer's LAC against the requested expected target; reject or flag it when the footer sits below the confirmed target. *(raised by yezh, yihao)*

**`server/storage/stagedstorage/writer_impl.go:1466` — the guard checks `lastEntryID` but the published value is a possibly stale `recoveredFooter.LAC`**

The guard at `:1411` compares `w.lastEntryID.Load()`, while the value actually written into the compacted footer is `w.recoveredFooter.LAC` returned by `validateLACAlignment` (`:2331-2338` → `:1466` → `:1508` `uploadCompactedFooter(..., lac)`). `validateLACAlignment` only checks `lastEntryID >= lac` (`:2365`) and `lac <= maxBlockLastEntryID` (`:2379`), never touching `expectedLastEntryId`, so a too-small `footer.LAC` always passes. With `recoveredFooter.LAC=3 < expected=5 <= lastEntryID=9`, both the guard and the alignment check succeed yet a footer with LAC=3 is published. Reachability: `fenceSegmentQuorum` derives this round's `lac` from `calculateLAC(successResults, ec)` once `ensembleCoverage` nodes answer (`segment_handle.go:1429-1443`); that value flows through `completeSegmentQuorum` → `segment_processor.go:200` `writer.Finalize(ctx, lac)` → `footer.LAC` (`:1235`). If a first round finalizes node A with `lac=3`, a later round (completion retry at `completion_manager.go:102-106`, or the recovery path at `segment_handle.go:1604/1616`) can compute `lac=5` from a changed response set and record it in metadata `LastEntryId` (`segment_handle.go:922`/`:1631`), while node A's repeated `Finalize` fast-returns at `:1167-1170` and leaves `footer.LAC` at 3. Node A's local recovery scan then yields `lastEntryID=9` from actual block records (`:2264-2274`, independent of the footer), the guard passes, and LAC=3 is published.

*Suggestion:* have compaction validate and publish the coordinator-confirmed expected LAC, and make repeated finalization detect an incompatible LAC instead of fast-returning. *(raised by yezh, yihao)*

**`server/processor/segment_processor.go:687` — the Compact response omits `LastEntryId`, so the client cannot detect an ignored target**

The response fills only `State`/`SealedTime`/`Size` (`:687-691`) and never returns the footer LAC actually published. `compactSegmentQuorum` accepts the first successful response with no integrity check (`woodpecker/segment/segment_handle.go:1537`), and the main `Compact` flow copies back only those same three fields (`:1789-1792`). When a server ignores or mishandles the expected target — an older build, the ForceCompact path, or the idempotent fast path — the client has no way to notice.

*Suggestion:* return the actually published footer LAC in the response and reject client-side any response below the expected value. *(raised by yezh, yihao)*

### Medium

**`proto/logstore.proto:185` — `expected_last_entry_id` is a presence-less proto3 scalar, so old servers silently ignore the new target**

The field is declared as a bare `int64 expected_last_entry_id = 5;` with no `optional` and no presence. During a rolling upgrade, a new client calling an old server has field 5 dropped as unknown and the server performs unguarded compaction; because the response (`server/processor/segment_processor.go:687-691`) carries no `LastEntryId`, the client cannot tell the target was ignored.

*Suggestion:* add protocol capability/version negotiation and return the compacted LAC in the response so the client can validate it. *(raised by yezh, yihao)*

**`proto/logstore_vtproto.pb.go:2892` — an absent field decodes as 0, so empty segments can never be sealed (liveness regression)**

`MarshalToSizedBufferVT` guards on `if m.ExpectedLastEntryId != 0`, so `-1` encodes fine, but an absent field decodes as 0 and is indistinguishable from an intentional 0. An old client omits field 5 → the new server forwards 0 (`server/service.go:600`) → the `expected >= 0` guard at `writer_impl.go:1410-1412` activates → every replica of an empty segment has `lastEntryID` defaulting to `-1` (`writer_impl.go:219`) → all replicas return `ErrSegmentCompactionDataBehind` and the segment can never reach `Sealed`. (The reverse direction is safe: for a new client the metadata `LastEntryId` of an empty segment comes from the fence result and is `-1`, correctly disabling the guard.)

*Suggestion:* use an `optional`/`oneof` field or explicit protocol-version handling so absence is distinguishable from a target of 0. *(raised by yezh, yihao)*

**`server/writer_registry.go:77` — `ForceCompact` hard-codes `-1`, fully disabling the guard, with no reason or audit trail**

`_, err := sp.Compact(ctx, -1)` passes `-1` unconditionally, and the guard only applies when `expectedLastEntryId >= 0` (`server/storage/stagedstorage/writer_impl.go:1410`), so this entry point turns the safety check off entirely. It is exposed through the admin handler at `common/http/server.go:234-238`, and unlike its sibling `ForceFence` (`:60`), the callback signature `ForceCompact func(logID, segmentID int64) error` (`common/http/server.go:42`) carries neither a reason nor an expected LAC. An operator can trigger compaction from a lagging node with no coordinator-confirmed target; combined with the idempotency defect above, one short footer makes every node's compact return success.

*Suggestion:* have the admin API accept and validate an expected LAC, or require an explicit unsafe acknowledgement plus an audited reason. *(raised by yezh, yihao)*

**`server/storage/stagedstorage/writer_impl_test.go:1329` — the leaking combination `lastEntryID >= expected > recoveredFooter.LAC` is untested**

The caught-up test (`:1329`) finalizes with `Finalize(ctx, 9)` at `:1348`, producing footer LAC=9, and iterates `[]int64{9, 5}` at `:1330` — both expected values sit on the `expected <= footer LAC` side. `..._RealWrites_RefusesWhenBehindExpected` (`:1295`) covers the rejection side with footer LAC=3 and `lastEntryID=3`. The other four new cases (`:715`/`:730`/`:748`/`:764`, compacting at `:723`/`:739`/`:757`/`:772`) only set `finalized.Store(true)` and drive `lastEntryID` directly, leaving `recoveredFooter` nil and never touching `footer.LAC`. The genuinely leaking combination (`lastEntryID=9`, footer LAC=3, expected=5) is uncovered.

*Suggestion:* add a real-writer case that writes through entry 9, finalizes with LAC=3, compacts with expected=5, and asserts either rejection or a published footer LAC of at least 5. *(raised by yezh, yihao)*

### Low

**`woodpecker/segment/segment_handle.go:1768` — the production source of the expected LAC is not pinned by tests**

Every test going through `segmentHandleImpl.Compact()` passes `mock.Anything` as the sixth argument (`segment_handle_test.go:4204`/`4281`/`4330`/`4397`/`4497` — `TestCompact_Success`, `TestCompact_SealedTimeFromCompactResponse`, `TestCompact_CompactionFails`, `TestCompact_AlreadyCompacting`, `TestCompact_DoesNotNotifyInline`), while the two cases that do assert an exact `expectedLAC` (`:5445`/`5447`/`5461`, `:5518`/`5521`/`5537`) call `impl.compactSegmentQuorum` directly and bypass this line.

*Suggestion:* make at least one `Compact()` test assert that the sixth RPC argument equals the metadata `LastEntryId`. *(raised by yezh, yihao)*

**`woodpecker/segment/segment_handle_test.go:5540` — the assertion targets a mock-fabricated field the real server never populates**

`assert.Equal(t, expectedLAC, result.LastEntryId)` passes only because the mock constructs `fullMeta.LastEntryId` at `:5524`; the real server response (`server/processor/segment_processor.go:687-691`) never sets `LastEntryId`, leaving it at the proto3 zero value. The assertion is disconnected from the production path and gives false confidence.

*Suggestion:* populate and validate the field in the real response, or drop the assertion until the production path supports it. *(raised by yezh, yihao)*

tinswzy:
Thanks for the detailed review. I went through the latest re-review items one by one; the earlier duplicated comments are covered by this summary.

1. Remote-footer idempotency fast path
Accepted and fixed. The remote compacted footer path now validates the footer against the requested target before returning success. We made this stricter than the suggestion: remote footer `LAC` must equal `expectedLastEntryId`, not merely be greater-or-equal.

2. Guard checks `lastEntryID` but publishes stale `recoveredFooter.LAC`
Accepted and fixed with a slightly different approach. Compaction now treats `expectedLastEntryId` as the coordinator-confirmed publish target. Before upload, it validates that local data covers the expected entry id, ignores any unconfirmed local tail beyond expected, rebuilds the compacted data only up to expected, and publishes a footer with `LAC = expectedLastEntryId`.

3. Compact response omits `LastEntryId`
Acknowledged, but intentionally not changed in this PR. After the server-side strict validation, Compact is now an atomic success/failure contract: success means the compacted data/footer already match the expected target, and failure means the segment should not be sealed. Returning the LAC can still be added later as defense-in-depth, but it is not required for this fix.

4. Proto3 scalar presence / rolling-upgrade compatibility
Acknowledged, but we are not handling rolling compatibility here. This code path has not been released yet, so we can require server and client to upgrade together. With the new behavior, callers must send the expected entry id explicitly.

5. Empty segment decoded as expected `0` during mixed-version upgrade
Same as above: this is a compatibility concern for mixed old/new deployments. For the supported path, empty segments use `expectedLastEntryId = -1`; compaction uploads only the footer, with `LAC = -1`.

6. `ForceCompact` hard-codes `-1`
Accepted and fixed. Admin force compact now requires an explicit `expected_last_entry_id` and passes it through to the writer. The admin is responsible for obtaining the authoritative value before issuing the request.

7. Missing test for `lastEntryID >= expected > recoveredFooter.LAC`
Accepted and covered by real writer tests. The important behavior is now verified through actual compact upload: local data may extend beyond expected, but compact only uploads entries up to expected, and the uploaded block index/footer must end exactly at `expectedLastEntryId`.

8. Production source of expected LAC not pinned by tests
Accepted. This is a useful low-risk test improvement on the segment-level path. It is separate from the staged writer compaction correctness fix, and can be added by asserting that `segmentHandleImpl.Compact()` passes metadata `LastEntryId` as the expected compact target.

9. `SkipsLaggingNode` asserts mock-fabricated `LastEntryId`
Accepted. Since the Compact response intentionally still does not carry `LastEntryId` in this PR, that assertion should either be removed or adjusted so it does not imply the client validates a field the server does not populate.

czs007:
## Re-review of `54f4718a9d73`

Both reviewers re-examined all twelve previously reported findings against the unchanged commit.

- **[high] `server/storage/stagedstorage/writer_impl.go:1387` — idempotent branch does not validate the existing compacted footer's LAC** — **Confirmed**: the branch still returns success before the `expectedLastEntryId` guard at `:1410`, and `readRemoteFooter` reads a shared object-storage key, so any short footer bypasses the new guard on every replica.
- **[medium] `server/storage/stagedstorage/writer_impl.go:2365` — guard compares `lastEntryID` while `footer.LAC` is what gets published** — **Confirmed**: the alignment check only rejects `lastEntryID < lac`, never `lac < expected`, so the two values can still diverge.
- **[medium] `proto/logstore.proto:185` — proto3 scalar has no presence; old servers silently ignore `expectedLastEntryId`** — **Confirmed**: the field is still a non-optional scalar, and the compact response carries nothing the client can cross-check.
- **[low] `server/writer_registry.go:77` — `ForceCompact` bypasses the LAC guard with no explicit unsafe confirmation or audit log** — **Confirmed**: `sp.Compact(ctx, -1)` still disables the guard, and the admin HTTP entry point accepts only log/segment IDs.
- **[low] `woodpecker/segment/segment_handle.go:1768` — the source of the authoritative expected LAC is not pinned by any test** — **Confirmed**: the full `Compact()` tests match the new argument with `mock.Anything` (`segment_handle_test.go:4204`, `:4281`, `:4330`), and the exact-value tests call `compactSegmentQuorum` directly, so nothing pins this line.
- **[low] `woodpecker/segment/segment_handle_test.go:5540` — `SkipsLaggingNode` asserts a field the server never populates** — **Confirmed**: the test fabricates `fullMeta.LastEntryId` and then asserts the same field, while `segment_processor.go:687-691` never sets it.
- **[high] `server/storage/stagedstorage/writer_impl.go:1466` — new guard checks `lastEntryID` but the published LAC comes from a possibly stale local footer** — **Confirmed**: `Finalize` returns early for an already-finalized writer (`:1167-1170`), so the footer set at `:1235`/`:1248` is never refreshed to a higher target.
- **[high] `server/processor/segment_processor.go:687` — Compact response does not populate `LastEntryId`, leaving the client without defense in depth** — **Confirmed**: the response still omits the field and the client accepts the first successful response without any LAC cross-check.
- **[medium] `proto/logstore_vtproto.pb.go:2892` — field absence is indistinguishable from 0, so empty segments can be misjudged during a rolling upgrade** — **Confirmed**: zero is omitted on the wire; an old client's request decodes as expected `0` and the new guard rejects a genuinely empty replica whose local last entry is `-1`.
- **[medium] `server/writer_registry.go:77` — `ForceCompact` hardcodes `-1`, so the admin endpoint can still seal a short segment** — **Confirmed**: same bypass as above; the endpoint cannot supply an authoritative target.
- **[medium] `server/storage/stagedstorage/writer_impl.go:1385` — remote-footer idempotent branch returns success ahead of the new guard** — **Confirmed**: `return totalSize, nil` at `:1392` still precedes any expectation validation.
- **[medium] `server/storage/stagedstorage/writer_impl_test.go:1348` — missing a case where expected > footer LAC and `lastEntryID >= expected`** — **Confirmed**: the caught-up cases finalize at footer LAC 9 and test only expected 9 and 5, both satisfying `expected <= footer.LAC`.

A dedicated concurrency and locking pass this round found no additional critical or high issues: the CAS plus deferred cleanup in `segment_processor.go:632-644` still serializes compaction per processor, `writer_impl.go:1378-1379` holds `w.mu` while reading `recoveredFooter`/`blockIndexes`, `Finalize` keeps a consistent `finalizeMu → drain flush → w.mu` order with no new reverse-acquisition path, and `compactSegmentQuorum` remains sequential RPC with no new goroutines or cancellation leaks.

---

## This round's findings

The new `expectedLastEntryId` guard can be bypassed or can pass while a short footer is still published, so already-acknowledged entries can be silently lost — details below.

### High

**`server/storage/stagedstorage/writer_impl.go:1387` — idempotent branch returns success without comparing the existing footer's LAC against `expectedLastEntryId`**

The idempotency check at `:1387-1393` returns `totalSize, nil` as soon as it reads a remote footer with the compacted flag, while the new authoritative guard only runs at `:1410` — the order is inverted. Because `footer.blk` uses the shared key `rootPath/logId/segId/footer.blk` (`:1973-1975`), a short footer published by any path is visible to every replica.

Failure: with quorum-confirmed `LastEntryId=5`, a compacted footer with `LAC=3` published from a lagging replica (via `ForceCompact` or the stale-LAC path at `:1508`) → the next normal compaction on *any* healthy node returns success at `:1387` without running the guard, the client seals the segment (`woodpecker/segment/segment_handle.go:1790`), compacted marks authorize `data.log` removal (`server/compacted_file_cleanup.go:398`), and the reader truncates at `footer.LAC=3` (`reader_impl.go:1653`) — entries 4 and 5 are permanently unreadable.

Suggestion: compare `existingFooter.LAC` with `expectedLastEntryId` before returning success in the idempotent branch, rejecting or quarantining a short footer — or move the guard above the idempotent branch entirely. *(raised by xaxys, chyezh)*

**`server/storage/stagedstorage/writer_impl.go:1411` — the guard validates `lastEntryID` while the published value is a potentially stale `recoveredFooter.LAC`**

The guard compares `w.lastEntryID.Load()` (`:1411-1412`), but the LAC actually written into the shared compacted footer comes from `validateLACAlignment` returning `w.recoveredFooter.LAC` (`:2331-2338` → `:1466` → `:1508` → `:1890`). These can diverge: `Finalize` writes the footer once (`:1235`, `:1248`) and returns early for an already-finalized writer (`:1167-1171`) without refreshing it to a higher LAC, whereas the quorum LAC can rise between fence rounds (`calculateLAC` at `segment_handle.go:1378-1382`).

Failure: node A has local entries through 5; round one fences `[3,3]` → `LAC=3`, so `Complete(3)` sets A's `footer.LAC=3`; round two fences `[3,5,5]` → `LAC=5`, and `Complete(5)` on A returns early at `:1170` leaving `footer.LAC=3`; compaction with `expected=5` then passes both `5 >= 5` and all three alignment checks and publishes a footer with `LAC=3` → the segment is sealed at `metadata.LastEntryId=5`, replica `data.log` files are deleted, and reads of entries 4/5 return `ErrEntryNotFound` with no recovery path.

Suggestion: require the LAC about to be published to be at least `expectedLastEntryId` before upload, or refresh the footer LAC to the expected value first. *(raised by xaxys, chyezh)*

**`server/processor/segment_processor.go:687` — Compact response omits `LastEntryId`, leaving no response-side integrity check**

`segmentProcessor.Compact` returns a `SegmentMetadata` with only `State`/`SealedTime`/`Size` (`:687-691`), which `server/service.go:604` passes through verbatim. The client accepts the first non-error response (`segment_handle.go:1512-1524`, `:1537`; `logstore_client_remote.go:338-342` only inspects `Status`) and transitions to `Sealed`. Only the request direction carries an expectation — nothing in the response can be reconciled against it. For contrast, `CompleteSegmentResponse` does populate `LastEntryId` (`server/service.go:596`).

Failure: during a rolling upgrade where the client is upgraded but a logstore node is not, the old server ignores field 5, seals a short compacted footer from its own `footer.LAC=3` while `metadata.LastEntryId=5`, and returns Success → the new client cannot detect it, seals the segment, and entries 4/5 are lost — reproducing the exact production bug this PR targets, for the whole upgrade window, with no alert or observable signal.

Suggestion: populate the actually published compacted LAC in the Compact response and have the client reject any response whose LAC is below the requested expectation. *(raised by xaxys, chyezh)*

### Medium

**`proto/logstore.proto:185` — `expected_last_entry_id` is a non-optional proto3 scalar, so absence and 0 are indistinguishable**

Field absence decodes to `0`, so the server cannot tell "old client did not send this field" from "the real expectation is 0". The guard at `writer_impl.go:1410-1419` activates for `0` and only disables at `-1`, which is precisely the sentinel this PR chose for "no expectation" — a value the absence case can never produce.

Failure: an old client sends a `CompactSegmentRequest` without field 5 against an upgraded staged server for a genuinely empty Completed segment (local `lastEntryID = -1`, `:219`) → `-1 < 0` triggers `ErrSegmentCompactionDataBehind` on every replica, `compactSegmentQuorum` fails entirely, and the segment stays `Completed` forever — the auditor retries each round and fails, `data.log` and its directory are never reclaimed, and no `footer.blk` is ever produced.

Suggestion: use `optional int64` for field presence, or an explicit protocol-version/capability negotiation, so legacy absence is distinguishable from an actual expectation of 0. *(raised by xaxys, chyezh)*

**`server/writer_registry.go:77` — `ForceCompact` hardcodes `expected=-1`, bypassing the new authoritative check**

`ForceCompact` passes `-1` to disable the guard (`:75-78`). Its only entry point is the HTTP admin route `POST /admin/logstore/compact` (`common/http/server.go:236-237` → `common/http/management/logstore_handler.go:97-101`), whose body carries only `log_id`/`segment_id`/`reason` (`:109-113`) — no authoritative expected LAC, no `unsafe`/`force` confirmation, and no indication that this bypasses a safety check.

Failure: with quorum-confirmed `LastEntryId=5`, an operator runs `curl -XPOST /admin/logstore/compact -d '{"log_id":L,"segment_id":S}'` against a staged node holding only `[0,3]` → the guard is skipped, `validateLACAlignment` is self-consistent against the local `footer.LAC=3`, and a shared compacted footer with `LAC=3` is published; the next quorum-driven compaction then accepts it through the idempotent branch, the segment is sealed, replica `data.log` files are deleted, and entries 4/5 are permanently lost from a single unconfirmed, unalerted admin command.

Suggestion: require the caller to supply an authoritative expected LAC (or have the server look it up from metadata), or make the unsafe bypass explicit and forbid it from publishing the shared commit footer. *(raised by xaxys, chyezh)*

**`server/storage/stagedstorage/writer_impl_test.go:1329` — no case covers `footer.LAC < expected <= lastEntryID`**

Neither real-writer case reaches the "guard passes but a stale footer LAC is published" path. The caught-up case (`:1329-1364`) calls `Finalize(9)`, pinning `footer.LAC=9`, then runs with `expected ∈ {9,5}` — both satisfy `expected <= footer.LAC`, so the published LAC is necessarily `>= expected`. The behind case (`:1295-1323`) is `Finalize(3)` with `expected=5`, i.e. the opposite side (`lastEntryID < expected`). The failing combination is never constructed, and the tests never assert the published `footer.LAC` at all (`:1359-1361`).

Failure: construct a staged writer with `footer.LAC=3`, local `lastEntryID=5`, and `expected=5` (reachable via the LAC rising between fence rounds with the second `Finalize` returning early at `:1167-1171`) and call `Compact` → every existing Compact test still passes, so the path that publishes a short footer and makes entries 4/5 permanently unreadable can merge without CI catching it.

Suggestion: add a real-writer case that finalizes at LAC 3, advances `lastEntryID` to 5, calls `Compact` with `expected=5`, and asserts no compacted footer is uploaded (a strict mock with no `PutObject` expectation). *(raised by xaxys, chyezh)*

tinswzy:
Thanks for the re-review. The latest review was still against `54f4718a9d73`; the fixes I described in the previous reply had not been pushed yet at that time.

I have now pushed the implementation in `8acf65a` (`fix: enforce expected LAC during staged compaction`). The remote-footer guard, stale local-footer publish path, expected-tail compaction upload, and admin `expected_last_entry_id` path should now be reflected in the PR diff. Please re-review against the new head.

czs007:
Adversarial review found two issues that should be addressed before merge: a rolling-upgrade incompatibility in the new `expected_last_entry_id` field that can still lose confirmed entries (or wedge compaction entirely), and a `-1` "no expectation" contract that the staged writer does not actually honor.

## High

**`proto/logstore.proto:185` — `expected_last_entry_id` has no field presence, breaking mixed-version compaction in both directions**

`int64 expected_last_entry_id = 5;` is a plain proto3 scalar, so absent and `0` are indistinguishable. The gRPC entry point passes the decoded value straight through (`server/service.go:600`) with no absent check, while the HTTP path already rejects a missing value via `*int64` (`common/http/management/logstore_handler.go:114-116`). The coordinator also cannot tell whether the peer enforced the new check: `CompactSegmentResponse` carries no `LastEntryId` to echo back (`server/processor/segment_processor.go:687-691`), and any successful response is unconditionally promoted to `Sealed` (`woodpecker/segment/segment_handle.go:1512-1537`, `1789-1797`).

Failure: in staged mode, a pre-PR coordinator compacts a `Completed` segment with confirmed `LastEntryId=100` against an upgraded LogStore, field 5 is absent and decodes to `expectedLastEntryId=0` → every intact replica trips `recoveredFooter.LAC(100) > expected(0)` and returns `ErrInvalidLACAlignment` (8107, non-retryable), so the quorum loop exhausts all nodes, the segment is stuck in `Completed` forever, `footer.blk` is never published and local `data.log` is never reclaimed; worse, a lagging replica whose local footer LAC happens to be `0`/`-1` with a non-empty block passes validation, truncates the merge at entry 0, and publishes a compacted footer with `LAC=0` as the global commit point, silently discarding confirmed entries 1..100. In the other direction, an old LogStore skips unknown field 5, compacts against its own local LAC and returns success, so the short-seal data loss this PR fixes still reproduces during the rollout window.

Suggestion: make the field `optional int64` and reject absent at `server/service.go:600`; add a protocol capability bit or a versioned RPC so `CompactSegmentResponse` echoes the expected LAC it actually validated, and have the coordinator skip non-matching nodes and defer compaction instead of sealing; add old-client → new-server mixed-version tests. (Note: minio mode is unaffected — `MinioFileWriter.Compact(ctx, _ int64)` ignores the parameter.) *(raised by chyezh, xaxys)*

## Medium

**`server/storage/writer.go:58` — the documented `-1 = no expectation` contract does not hold for the staged writer, disabling the ForceCompact escape hatch**

The interface comment states `expectedLastEntryId < 0 means "no expectation" and disables that check`, and the PR description says `-1` is "used by manual/force compaction (ForceCompact) and genuinely-empty segments." The staged implementation does the opposite: every non-empty segment goes through `validateLocalDataCoversExpected`, whose first branch rejects `expectedLastEntryId < 0` outright (`server/storage/stagedstorage/writer_impl.go:1524-1527`, reached from `:1405`); `-1` is only legal when the local block set is empty (`:1492`).

Failure: an operator force-compacts a non-empty segment per the documented contract with `POST /admin/logstore/compact {"log_id":42,"segment_id":7,"expected_last_entry_id":-1}` → `StagedFileWriter.Compact(-1)` returns `ErrInvalidLACAlignment` ("non-empty compaction requires non-negative expected last entry id, got -1") and HTTP 409, so compaction never runs at all. This is precisely the tool an operator would reach for when a segment is wedged in `Completed`; they must instead dig the exact confirmed `LastEntryId` out of the coordinator. The same mismatch makes `compactRemoteFooterIfPresent` fail on an already-compacted segment replayed with `-1` (remote footer LAC=100 vs expected `-1`), losing the previously idempotent success. The new processor-level `NoExpectationSkipsCheck` tests use a mock writer (`mockWriter.EXPECT().Compact(mock.Anything, mock.Anything)`) and only assert argument pass-through, so they do not cover the real staged rejection — the coverage here is illusory.

Suggestion: either make the staged implementation genuinely support `-1` by falling back to the local footer LAC, or correct the interface comment, PR description and HTTP docs, and have `ForceCompact` reject `-1` explicitly with an error message pointing at where the correct value comes from. *(surfaced during verification)*

czs007:
Adversarial review of the compaction-target changes found four issues: one operability defect that lets a single mistyped force-compact permanently wedge a segment, plus three medium issues around mixed-version compatibility, unrecoverable footer mismatches, and a broken CLI path.

## High

**`server/writer_registry.go:71` — `ForceCompact` writes the operator-supplied `expected_last_entry_id` straight into the shared footer without cross-checking segment metadata.**
The value flows unmodified from `logstore_handler.go:119` → `cmd/main.go:256` → `writer_registry.go:71` → `StagedFileWriter.Compact` → `writer_impl.go:1441 uploadCompactedFooter(...)`, becoming the LAC of `footer.blk`. Only the low side is bounded (`writer_impl.go:1532` rejects `recoveredFooter.LAC > expected`); there is no upper bound against the coordinator-confirmed LAC, and unlike `fence` (`logstore_intervene.go:41-48`) the compact path has neither `--reason` nor a `-y` confirmation.
*Failure:* segment confirmed at `LastEntryId=30` while a node has locally persisted but unconfirmed entries up to `40`; an operator posts `expected_last_entry_id: 35` → all validations pass, entries 31-35 are sealed into the quorum-shared `footer.blk` with `LAC=35`, and every subsequent coordinator-driven compaction at the real `LastEntryId=30` hits a non-retryable `ErrInvalidLACAlignment` (`remote compacted footer LAC 35 != expected 30`) on every node — the segment is stuck in `Completed` forever, local `data.log` is never reclaimed, and recovery requires hand-editing object storage.
*Suggestion:* validate the supplied target against the segment metadata's `LastEntryId` in `ForceCompact` (or require an explicit `--force` plus confirmation when it deviates), and document the hazard in the CLI help text. *(surfaced during verification)*

## Medium

**`proto/logstore.proto:184` — `expected_last_entry_id` is a non-optional proto3 scalar, so a legacy coordinator's request decodes as an explicit target of `0`.**
`server/service.go:600` passes the decoded value through with no presence or sanity check, and `writer_impl.go:1236` shows a healthy replica's footer LAC equals the coordinator-confirmed LAC, so it is always `> 0` for non-trivial segments.
*Failure:* during a rolling upgrade where logstore nodes are upgraded but the coordinator is not, any `Completed` segment with `LastEntryId >= 1` fails on every node at `writer_impl.go:1532` with the non-retryable `ErrInvalidLACAlignment` (`local footer LAC 5 > expected 0`) → the quorum compaction fails everywhere, segments never reach `Sealed`, local `data.log` grows unbounded, and `log_auditor.go:68` retries and logs errors every cycle until the coordinator is upgraded (empty segments are separately rejected at `writer_impl.go:1492` because `expected != -1`).
*Suggestion:* make the field presence-aware (`optional`) or gate it behind explicit version/capability negotiation, and reject legacy requests with a diagnosable error instead of treating a missing field as `expected = 0`. *(raised by chyezh, xiaocai2333)*

**`server/storage/stagedstorage/writer_impl.go:1464` — a mismatched existing compacted footer is a terminal error with no repair path.**
When the remote footer's LAC differs from the expected target, the code returns non-retryable `ErrInvalidLACAlignment` (`common/werr/errors.go:208`) without first checking whether this replica could safely rebuild the footer. Since `footer.blk` is shared across the quorum, every node reads the same mismatched footer and fails identically. The state is reachable: `segment_handle.go:1797-1803` only logs a warning when the post-compaction `UpdateSegmentMetadata` fails, leaving a published footer on a segment still marked `Completed`.
*Failure:* a short footer published by a legacy/mixed-version node, or a crash between footer upload and the metadata update → every `log_auditor.go:68` cycle retries compaction and every node is rejected at line 1464, so even fully caught-up replicas cannot replace the footer; the segment stays in `Completed` permanently, local `data.log` is never reclaimed, and only manual object-store repair restores it.
*Suggestion:* add a coordinator-authorized repair/overwrite path that validates local coverage against the authoritative target (`localLast >= expected` and block indexes covering `expected`) before replacing a mismatched footer, and only reject when that check fails. *(raised by chyezh)*

**`cmd/wpcli/cmd/logstore_intervene.go:70` — `wp logstore compact` was not updated for the new handler contract, so force-compact is unusable.**
The CLI payload still contains only `{log_id, segment_id}`, while `common/http/management/logstore_handler.go:115-118` now rejects a nil `expected_last_entry_id` with HTTP 400 before the compact callback is ever invoked; `logstore_intervene.go:106` then surfaces it as a network error. The breakage is masked because the mock handler at `cmd/wpcli/cmd/logstore_test.go:53` accepts any body without validation and returns success.
*Failure:* any invocation of `wp logstore compact <node> --log 42 --seg 7` → `POST /admin/logstore/compact returned 400: {"error":"expected_last_entry_id required"}`, making the forced-compaction operational path 100% unavailable while CLI tests stay green and CI catches nothing.
*Suggestion:* add a required `--expected-lac` flag and include it in the payload, and update the test mock handler to parse and assert that the posted body carries `expected_last_entry_id`. Please also fold in the safety guidance from the `ForceCompact` item above when wiring this flag. *(raised by xiaocai2333, chyezh)*

czs007:
## Summary

Adversarial review surfaced two issues worth the author's attention — one pre-existing concurrency gap that this PR narrows but does not close, and one operator-facing hazard on the manual force-compact path.

---

### Medium

**`server/storage/stagedstorage/writer_impl.go:1845` — Cross-node concurrent compaction shares deterministic block keys, allowing footer/data divergence**

Compacted block keys come from `getCompactedBlockKey(mergedBlockID)` where `mergedBlockID` is just the merge-task index `int64(i)` (`writer_impl.go:1633`), yielding `<root>/<log>/<seg>/m_<i>.blk` — independent of which node or attempt produced it. Block upload (`:1848`) and footer publication (`:1429`) are two separate object writes with no atomicity between them, and the only mutual exclusion is the `s.compacting` CAS in `segment_processor.go:633`, which is per-process and cannot stop two nodes from compacting the same segment. Merge-task grouping derives from each replica's own local `blockIndexes`/`BlockSize` (`planMergeBlockTasks`, `:1674`), so even with every replica cropping to the same `expectedLastEntryId`, block boundaries can still differ across replicas — while the reader fetches `m_<BlockNumber>.blk` using the footer's `IndexRecord{BlockNumber, BlockSize}` (`reader_impl.go:1554-1565`).

Failure: an operator runs `wpcli logstore compact node-B --log L --seg S --expected-last-entry-id N` (via `ForceCompact` → `sp.Compact`, `server/writer_registry.go:71`) while the auditor is already running `compactSegmentQuorum` against node A, and neither has published a footer yet — A writes `m_0=[0,4]`/`m_1=[5,9]` and publishes its footer, then B overwrites `m_0` with its own `[0,9]` bytes → the reader GETs `m_0` at the footer's recorded `BlockSize`, gets a truncated prefix of B's block, and fails `BlockHeaderRecord`/CRC validation, leaving that block of the sealed segment permanently unreadable (or, in the reverse interleaving, leaving orphaned `m_i` objects as a space leak).

Suggestion: this is a pre-existing infrastructure gap — the key naming and non-atomic footer publication both predate this PR and are untouched by it, and the new `compactRemoteFooterIfPresent` (`:1452`) short-circuits when a compacted footer already exists, narrowing the race window to the period before either node sees a footer. Recommend tracking it in a separate issue (attempt-unique block keys plus conditional-write / atomic footer publication) rather than blocking this PR. *(raised by chyezh, xiaocai2333)*

---

### Low

**`server/storage/stagedstorage/writer_impl.go:1524` — `--expected-last-entry-id` pointed at a lagging replica's own LAC short-seals the segment and drops acknowledged data**

The manual force-compact path relies entirely on the operator supplying an accurate quorum-confirmed LAC. `--expected-last-entry-id` is required by the CLI (`cmd/wpcli/cmd/logstore_intervene.go:77,80`) and by the HTTP handler, but neither cross-checks it against `segmentMeta.LastEntryId` — the value passes straight through to `sp.Compact`. Of the four checks in `validateLocalDataCoversExpected`, only `recoveredFooter.LAC > expected` (`:1532`) rejects a too-low value, and that check only fires on a replica that has already caught up.

Failure: the segment's quorum-confirmed `LastEntryId` is 99 and `node-3` is a lagging replica whose local footer LAC is 60; an operator reading that LAC from the node's logs (rather than from `segmentMeta.LastEntryId` in etcd) runs `wpcli logstore compact node-3 --log L --seg S --expected-last-entry-id 60` → all four validations pass (`60 > 60` is false, `localLast >= 60`, blocks cover 60), the new crop logic (`planMergeBlockTasks:1678/1709`, `extractCompactedDataRecords:1888`) actively discards local entries past 60 instead of uploading them, and the footer is published with LAC=60 — sealing the segment at 60 and authorizing replicas to reclaim their local `data.log`, so acknowledged entries 61..99 are lost permanently. That is exactly the #242 failure this PR fixes, re-entered through the admin surface.

Suggestion: this is manual-override semantics and the flag help text already says "Exact quorum-confirmed last entry ID"; the automatic auditor path uses `segmentMeta.LastEntryId` (`woodpecker/segment/segment_handle.go:1768`) and is unaffected. Consider having the server cross-check the supplied value against `segmentMeta.LastEntryId` and reject any value below it, so a mistyped or log-sourced LAC cannot silently truncate a segment. *(raised by xiaocai2333)*

czs007:

Verified:
- `woodpecker/server/segment/writer_impl.go:3003` — `validateLocalDataCoversExpected` cannot admit a mid-segment hole, because the flush path reads `ReadEntriesRange(FirstEntryId, ExpectedNextEntryId)` and the sequential buffer only advances `ExpectedNextEntryId` on a contiguous prefix, so persisted blocks are always contiguous.
- `woodpecker/server/segment/writer_impl.go:1887` — compaction cropping is fail-closed: `parseLocalBlockForCompaction` verifies whole-block CRC and per-record type before counting, and a `kept != recordsToKeep` mismatch aborts rather than uploading a short block.
- `woodpecker/server/segment/writer_impl.go:3292` — the recomputed footer `totalRecords` (`maxBlockLastEntryID - blockIndexes[0].FirstEntryID + 1`) matches the cropped block headers, since block indexes are sorted by block number and start at entry 0.
- `woodpecker/server/segment/segment_handle.go:1512` — lagging and empty replicas are rejected rather than silently compacted: per-node `ErrSegmentCompactionDataBehind` is skipped, an all-behind quorum surfaces the error, and the refusal test asserts no `PutObject` occurs.
- `woodpecker/server/service.go:2507` — cross-version behavior is fail-closed: the legacy `CompactSegment` RPC always returns `ErrUnsupportedVersionError`, and new clients never fall back to it.
- `woodpecker/common/objectstorage/writer_impl.go:2648` — ignoring `expectedLastEntryId` in object-storage mode is sound, as that mode has a single authoritative replica with no per-replica divergence.

czs007:
Following up on the High item from the adversarial review above (`expected_last_entry_id` has no field presence): a candidate patch is ready on a branch, in case it saves you the work. It is offered, not required — take it, take part of it, or ignore it.

**Branch:** [`czs007:team2-fix/T28f01905a1`](https://github.com/zilliztech/woodpecker/compare/master...czs007:team2-fix/T28f01905a1?expand=1) (8 files, +145/-24)

### What it changes

- `proto/logstore.proto` — `expected_last_entry_id` becomes `optional int64`, so absent is distinguishable from a genuine target of `0`. Comment spells out why: `0` is a legal boundary (a segment holding only entry 0), so a plain scalar lets a caller that never set the field truncate the segment to its first entry.
- `server/service.go` — `CompactSegmentWithExpected` refuses a request with the field absent, before any compaction work starts, with a message pointing the caller at the client upgrade.
- `woodpecker/client/logstore_client_remote.go` — always sends the boundary explicitly.
- `woodpecker/segment/segment_handle.go` — comment recording why a bare success response is now safe to accept: the RPC is both versioned (an older node answers `Unimplemented` rather than compacting against its own local view) and presence-tracked.
- Tests: `server/service_test.go` +62, `woodpecker/client/logstore_client_remote_test.go` +30, covering absent-field rejection and explicit-send.

It does not address the Medium item (the `-1` "no expectation" contract that the staged writer does not honor) — that one is a separate decision about which side of the contract should change.

### How far it was verified

`make build` passes on top of `20f3870`. Unit tests were run over the affected packages before and after the change and compared: identical results on both sides, no new failures.

Four packages fail identically with and without the patch, for environment reasons rather than code: `meta` and `woodpecker` want to bind `127.0.0.1:2380` for their own embedded etcd (another etcd already holds that port on the machine used), `common/objectstorage` hangs in `TestNewAzureObjectStorageClient_IAM_InvalidCredentials` waiting on the Azure IMDS address `169.254.169.254`, which only resolves inside Azure, and `server`'s `TestCompactedFileCleanup_StartupWalkSeedsStoredGauges` compares a file mtime against `task.startTime` where the clock granularity makes a just-created `data.log` look older by ~1.6ms. So those four are unverified here rather than passing — worth knowing before relying on the run.

PR #245, opened 2026-07-26
fix(log): stop auditor after writer invalidation

### LINKED ISSUE #244: Stop auditor maintenance after writer ownership invalidation
## Background

Each log writer starts background maintenance goroutines for the log auditor and compacted-segment notification distribution. Writer ownership can later be invalidated when the session lease expires, keepalive checks fail, or a newer writer fences the old owner.

The write path correctly rejects work after invalidation, but the background maintenance goroutines are currently tied only to explicit writer close. Losing writer ownership does not necessarily close the writer immediately.

## Problem

An invalidated writer can continue running auditor cycles after it no longer owns the log. If a new writer has already acquired ownership, both old and new auditors may concurrently:

- scan Completed segments;
- start segment compaction;
- update segment lifecycle metadata;
- perform truncate/cleanup work;
- distribute compacted-segment notifications.

Metadata CAS protects some final state transitions, but it does not cancel side effects that happen before the CAS, including object uploads and notifications. Writer fencing therefore needs to terminate writer-owned maintenance work as well as reject new writes.

PR #243 makes compaction validate the quorum-confirmed expected LAC, but it does not bind auditor lifetime to writer ownership.

## Expected behavior

- Give each writer an explicit maintenance context.
- Cancel it whenever writer ownership becomes invalid, including session expiration, failed lease validation, fencing/invalidation callbacks, and explicit close.
- Derive auditor-cycle contexts from the maintenance context.
- Stop a compaction pass between segments when cancellation is observed.
- Stop compacted-segment notification distribution on the same lifecycle signal.
- Make writer close wait for the auditor goroutine to exit.

## Acceptance criteria

- Tests cover invalidation callbacks and session-loss paths stopping the auditor.
- In-flight auditor work observes context cancellation and does not continue with later segments.
- Explicit writer close waits for background auditor termination.
- Normal auditor and notification behavior remains unchanged while the writer is valid.

## Out of scope

- Manual ForceCompact concurrency and operator coordination.
- Compaction LAC/data-boundary validation, which is handled by #243.

tinswzy:
## Summary

- bind the log auditor and compacted-segment notification distributor to a writer-owned maintenance context
- cancel maintenance when writer ownership is invalidated, the session expires or fails validation, or the writer closes
- propagate cancellation into auditor compaction passes and stop before processing additional segments
- wait for the auditor goroutine to exit during writer close
- cover invalidation, session expiration, canceled compaction, and close-waits-for-auditor behavior

## Motivation

Writer fencing already prevents an old owner from accepting new writes, but it did not stop the background auditor owned by that writer. After ownership moved to a new writer, the old auditor could continue compaction, cleanup, and compacted-mark notification work concurrently. This change makes writer-owned maintenance end with writer ownership.

This is separate from the compaction LAC validation merged in #243: #243 validates what may be compacted, while this PR controls which writer-owned background task may continue running.

## Testing

- go test -race -short ./woodpecker/log
- focused invalidation, session, and close race tests with count 5

Closes #244

### DIFF
--- a/woodpecker/log/internal_log_writer.go
+++ b/woodpecker/log/internal_log_writer.go
@@ -59,6 +59,9 @@ type internalLogWriterImpl struct {
 	// validation related fields
 	isWriterValid       atomic.Bool
 	onWriterInvalidated func(ctx context.Context, reason string)
+	maintenanceCtx      context.Context
+	maintenanceCancel   context.CancelFunc
+	auditorWG           sync.WaitGroup
 
 	// Mutex to ensure only one truncation cleanup task is running at a time
 	cleanupMutex      sync.Mutex
@@ -69,6 +72,7 @@ type internalLogWriterImpl struct {
 func NewInternalLogWriter(ctx context.Context, logHandle LogHandle, cfg *config.Configuration) LogWriter {
 	ctx, sp := logger.NewIntentCtxWithParent(ctx, WriterScopeName, "NewInternalLogWriter")
 	defer sp.End()
+	maintenanceCtx, maintenanceCancel := context.WithCancel(context.Background())
 	w := &internalLogWriterImpl{
 		logIdStr:           fmt.Sprintf("%d", logHandle.GetId()),
 		logHandle:          logHandle,
@@ -79,6 +83,8 @@ func NewInternalLogWriter(ctx context.Context, logHandle LogHandle, cfg *config.
 		cleanupManager:     segment.NewSegmentCleanupManager(cfg.Minio.BucketName, cfg.Minio.RootPath, logHandle.GetMetadataProvider(), logHandle.(*logHandleImpl).ClientPool),
 		notifyManager:      segment.NewSegmentCompactedNotifyManager(cfg.Minio.BucketName, cfg.Minio.RootPath, logHandle.GetMetadataProvider(), logHandle.(*logHandleImpl).ClientPool),
 		notifySegsCh:       make(chan map[int64]*meta.SegmentMeta, 1),
+		maintenanceCtx:     maintenanceCtx,
+		maintenanceCancel:  maintenanceCancel,
 	}
 	// Set sessionValid to true
 	w.isWriterValid.Store(true)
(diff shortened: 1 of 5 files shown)


czs007:
Adversarial review surfaced one non-blocking issue: the compacted-mark distributor is not fully bound to the writer-owned maintenance lifecycle.

## Medium

**`log_auditor.go:139` — Compacted-mark distributor is neither cancellable mid-pass nor joined by `Close`, so it outlives writer ownership**

This PR correctly fixes the distributor *loop* — it now stops on `maintenanceCtx.Done()` instead of only `writerClose`, so it exits at the next `select` on every ownership-loss path. The residual gap is the single in-flight pass:

1. **The pass context is not derived from `maintenanceCtx`.** `runNotifyDistributor` builds each pass context with `logger.NewIntentCtx(...)` (`log_auditor.go:139`), which is rooted at `context.Background()` (`common/logger/logger.go:197-199`). `distributeCompactedMarks` (`log_auditor.go:100-114`) has no per-segment `ctx.Err()` check — unlike `compactCompletedSegments`, which this PR did fix (`log_auditor.go:58-60`). Once a pass is entered, cancellation cannot reach it: up to `maxCompactedNotifyPerCycle` = 64 segments, each doing an etcd Get/Create plus a per-node gRPC fanout bounded only by `notifySegmentCompactedTimeout` = 10s, then a CAS write-back.
2. **The loop does not reliably exit on cancellation either.** The `select` at `log_auditor.go:136-150` races `maintenanceCtx.Done()` against `segsCh`; when a snapshot published by the last auditor cycle is still buffered in the cap-1 `notifySegsCh`, both cases are ready and Go picks uniformly at random, so a fenced writer can run one more full pass before returning.
3. **`Close` never joins it.** Both `Close` paths wait only on `auditorWG` (`log_writer.go:685-689`, `internal_log_writer.go:602-605`); nothing tracks the distributor goroutine.
4. **The auditor's own cancellation path skips the in-process mitigation.** On cancellation the auditor returns at `internal_log_writer.go:310-313` / `log_writer.go:386-389` — after `publishSegmentsSnapshot` but before `markTruncatedSegmentsReaped`. So the last snapshot handed to the distributor is published, and the reap sync that `log_auditor.go:180-189` exists to perform for exactly that snapshot never runs.

Failure: in a service-mode (staged storage) deployment where one quorum node is slow or black-holed, a distribution pass is mid-flight when the application runs the normal shutdown sequence `writer.Close(ctx)` then `client.Close(ctx)` (same shape when the writer is fenced: `onWriterInvalidated` → `stopMaintenance` mid-pass) → `Close()` returns as soon as `auditorWG` drains, and the uncancellable pass keeps issuing `NotifySegmentCompacted` RPCs and etcd reads/CAS write-backs for up to ~10 minutes afterward against the metadata provider and gRPC client pool that `woodpeckerClient.Close` (`woodpecker/woodpecker_client.go:332-339`) has just torn down — producing a stream of "notify failed; will retry next cycle" warnings for a writer that has no next cycle, and leaving any segment whose node marks were already delivered stuck in `NOTIFY_IN_PROGRESS` because its `finalizeStatus` write-back fails.

Suggestion: derive the pass context from `maintenanceCtx` and check it per segment in `distributeCompactedMarks`; check `maintenanceCtx.Err()` at the top of the `segsCh` case so a buffered snapshot cannot win the race; track the distributor in a `WaitGroup` that `Close` joins before `logHandle.Close()`. Non-blocking — a follow-up is acceptable.

Two related concerns from earlier rounds do *not* hold and need no action: a stale notify cannot resurrect a truncated segment, because `NotifySegmentCompacted` verifies the compacted footer in object storage before writing the mark (`server/logstore.go:666-685`) and the truncate reap deletes MinIO objects before local files (`server/storage/stagedstorage/segment_impl.go:109-121`), so a late notify finds no footer and is refused; and `logHandle.Close()` does not shut the `ClientPool` (`woodpecker/log/log_handle.go:1020-1039`) — that teardown lives in `woodpeckerClient.Close`. (raised by xiaocai2333, chyezh)

tinswzy:
Thanks for the detailed review. The distributor lifecycle gap is fixed in b797344.

Changes made:
- runNotifyDistributor now receives the full writer maintenance context, and each pass is derived from it with NewIntentCtxWithParent.
- cancellation is checked after receiving a buffered snapshot and before and after each per-segment notify, so ownership loss cannot start or continue another pass.
- auditor and distributor goroutines are tracked by one maintenance WaitGroup; both Close paths cancel maintenance and wait for both goroutines before closing the log handle.
- truncated segments are marked reaped before publishing the snapshot or starting compaction, so cancellation cannot skip that in-process safeguard.

Added regression coverage for the canceled-context versus buffered-snapshot race, mid-pass cancellation, truncated reap ordering, and Close waiting for the distributor. go test -race -short ./woodpecker/log passes.

PR #253, opened 2026-07-31
fix(fence): require ack-quorum coverage when resolving the completion LAC

### LINKED ISSUE #252: bug(fence): calculateLAC drops -1 votes and clamps, producing a completion target no ack quorum can cover
## Summary

`calculateLAC` can return a LAC that **no ack quorum of replicas actually covers**. The fence phase then hands that value to `completeSegmentQuorum` as the frozen completion target. Since #250 the complete phase correctly refuses to publish `Completed` unless `Aq` finalized replicas locally cover the target, so an over-estimated target makes completion **fail forever** — there is no catch-up/repair path that can raise a behind replica to an entry it never received.

The result is a permanently wedged log: `fenceAllActiveSegments` treats the failure as fatal, WAL open fails, and every write to that pchannel is blocked indefinitely.

## Impact (observed in production)

A chaos test killed woodpecker service pods while writes were in flight. Afterwards `rootcoord-dml_0` (logId=2) never recovered — segment 22 retried fence→complete every few seconds for 3+ hours and never completed. Milvus inserts on that pchannel were blocked the entire time.

```
Successfully fenced segment on node   node-1  lastEntryId=0
Successfully fenced segment on node   node-3  lastEntryId=-1
Successfully fenced segment on node   node-4  lastEntryId=-1
Calculated LAC from quorum fence responses  lac=0  successCount=3  ensembleCoverage=2  allResults="[0,-1,-1]"
Successfully completed segment on node  node-1  lastEntryId=0   targetLAC=0
Replica finalized but does not cover target LAC  node-3  localLastEntryId=-1  targetLAC=0
Replica finalized but does not cover target LAC  node-4  localLastEntryId=-1  targetLAC=0
Insufficient qualified responses for quorum complete  finalizedCount=3 qualifiedCount=1 requiredAckQuorum=2
```

Entry 0 had only ever reached node-1 (the other two were down at that moment, so the append was never acked to the caller). The correct LAC is `-1`; fence produced `0`.

Upstream effect:

```
OpenInternalLogWriter -> fenceAllActiveSegments activeSegmentIds="[22]"
  -> "failed to fence log: ... insufficient qualified complete responses: qualified=1 finalized=3 required=2"
  -> walmanager "remove wal success because that previous open operation is failure"
  -> balancer "channel is not assigned to any server" [term=2482]
  -> producer backoff loop; inserts blocked
```

## Root cause

`woodpecker/segment/segment_handle.go`, `calculateLAC`:

```go
// Filter out negative values (e.g. -1 returned by nodes with no data).
for _, r := range results { if r >= 0 { validResults = append(validResults, r) } }
if len(validResults) == 0 { return -1 }
sort.Slice(validResults, ...)          // ascending
lacIndex := len(validResults) - ackQuorum
if lacIndex < 0 { lacIndex = 0 }       // <-- fabricates an answer
return validResults[lacIndex]
```

The index formula itself is right: on an ascending sort, `len - ackQuorum` is the **ackQuorum-th largest**, i.e. the largest value covered by at least `ackQuorum` responders. Two things around it are wrong:

1. **Filtering `-1` breaks the order statistic.** Failed responses never reach `successResults` (they take the `result.err != nil` branch), so every element is a successful fence response and a `-1` is an honest "I hold nothing" vote — a vote *against* every entry id >= 0. Dropping it shrinks `len` while `ackQuorum` stays fixed, so the index goes negative.
2. **The clamp turns "cannot prove" into "assume".** `lacIndex < 0` means there are fewer surviving values than the quorum requires, i.e. no value can be shown to have `Aq` coverage. Clamping to 0 returns the minimum survivor instead of failing.

With `[0,-1,-1]`, Aq=2:

| | result |
|---|---|
| keep all | sorted `[-1,-1,0]`, index `3-2=1` -> `-1` (correct) |
| filtered | `[0]`, index `1-2=-1` -> clamp 0 -> `0` (claims entry 0 has >=2 copies; it has 1) |

A third, latent issue: the call site passes `ensembleCoverage` (`Es-Aq+1`) into a parameter named `ackQuorum`. For the only reachable ensemble sizes (`GetEnsembleSize()` returns 3 or 5) `ec == aq`, so this is currently invisible, but the two quantities mean different things and should not be conflated.

## Invariant that should hold

```
lac == -1  ||  count(r >= lac for r in results) >= ackQuorum
```

A property check over the existing test table plus ~4000 random inputs: the current implementation violates this in 774/4015 cases; keeping all responses and failing when `len(results) < ackQuorum` gives 0 violations and always returns the maximal legal value.

## Why the tests did not catch it

This is not a coverage gap — `TestCalculateLAC` covers the mixed case and **asserts the wrong answer**:

```go
{ name: "one node empty returns -1",                     results: []int64{0,-1}, ackQuorum: 2, expected: 0 },  // name contradicts assertion
{ name: "bug scenario: fence with [0, -1] ackQuorum=2",  results: []int64{0,-1}, ackQuorum: 2, expected: 0 },
{ name: "ackQuorum exceeds valid count uses index 0",    results: []int64{3},    ackQuorum: 2, expected: 3 },  // clamp written in as spec
```

All three violate the invariant above. The fully-empty case `[-1,-1] -> -1` passes only because of the `len(validResults) == 0` early return, which never reaches the order statistic.

The filter was added in d54940e (#92) to fix the opposite direction (an empty/restarted node's `-1` dragging LAC down and causing readers to EOF). That change traded a truncation bug for an over-estimation bug. Before #250 the over-estimation was silent, because `completeSegmentQuorum` counted any finalized replica regardless of its tail — which is exactly the data-integrity hole #250 closed.

## Proposed fix

Fold the response-count gate and the LAC computation into one function so the precondition that makes the index safe lives next to the index:

```go
func (s *segmentHandleImpl) resolveFenceLAC(results []int64, ackQuorum int) (int64, error) {
	if len(results) < ackQuorum {
		return -1, werr.ErrAppendOpQuorumFailed.WithCauseErrMsg(...)
	}
	sorted := make([]int64, len(results))
	copy(sorted, results)
	sort.Slice(sorted, func(i, j int) bool { return sorted[i] < sorted[j] })
	return sorted[len(sorted)-ackQuorum], nil
}
```

and pass `quorumInfo.Aq` rather than `ensembleCoverage`. The gate value is numerically unchanged for the reachable ensemble sizes (3 and 5, where `ec == aq`), so there is no behaviour change there; what changes is that the threshold is now the one the computation actually requires. Because `aq >= ec` always holds, gating on `aq` still implies the ensemble-coverage property.

Tests: add the invariant as a property test, and correct the three expectations above.

## Note on rollout order

An already-finalized replica has the over-estimated value durably recorded — `Finalize` writes `footer.LAC = lac` (the requested value), not its local tail, so the two behind replicas above hold `footer.LAC=0` while returning `-1`.

- Against a pre-#250 server, a finalized writer returns `(localLastEntryId, nil)` for any subsequent `Finalize`, so once fence computes `-1` these segments complete and self-heal.
- Against a #250 server, `recoveredFooter.LAC != lac` returns `ErrInvalidLACAlignment`, so a segment already poisoned with `footer.LAC=0` cannot be repaired by lowering the target.

Lowering a completion target is always safe (it retracts an assertion that should never have been published) while raising it is not; #250's server-side check currently refuses both. Worth a follow-up.

tinswzy:
Fixes #252

## Problem

`calculateLAC` filtered out `-1` responses before applying its order statistic, then clamped the resulting negative index to `0`:

```go
for _, r := range results { if r >= 0 { validResults = append(validResults, r) } }
lacIndex := len(validResults) - ackQuorum
if lacIndex < 0 { lacIndex = 0 }
```

Failed nodes take the `result.err != nil` branch and never reach that slice, so every element is a successful fence response and a `-1` is an honest "I hold nothing" vote — a vote against every entry ID `>= 0`. Discarding it shrinks the sample while `ackQuorum` stays fixed, so the index goes negative and the clamp returns the minimum survivor instead of failing.

With `[0,-1,-1]` and `aq=2` the filter left `[0]`, the index clamped to `0`, and fence returned `0` — a target only one of three replicas could ever cover.

Since #250 the complete phase correctly refuses to publish `Completed` unless `Aq` finalized replicas locally cover the target, and no replica can catch up to an entry it never received. So an over-estimated target wedges the log permanently: `fenceAllActiveSegments` keeps failing, WAL open fails, and every write to that pchannel is blocked. Issue #252 has the production trace — a chaos-killed node left `rootcoord-dml_0` retrying for 3+ hours with `qualifiedCount=1 requiredAckQuorum=2`.

## Change

`calculateLAC` becomes `resolveFenceLAC`, which folds the response-count gate into the computation so the precondition that makes the index safe lives next to the index:

```go
func (s *segmentHandleImpl) resolveFenceLAC(results []int64, ackQuorum int) (int64, error) {
	if len(results) < ackQuorum {
		return -1, werr.ErrAppendOpQuorumFailed.WithCauseErrMsg(...)
	}
	sorted := make([]int64, len(results))
	copy(sorted, results)
	sort.Slice(sorted, func(i, j int) bool { return sorted[i] < sorted[j] })
	return sorted[len(sorted)-ackQuorum], nil
}
```

- All responses are counted. On an ascending sort, `len - ackQuorum` is the `ackQuorum`-th largest, i.e. the largest value covered by at least `ackQuorum` responders — the index formula was always right, only the input to it was wrong.
- Fewer than `ackQuorum` responses is a failure to prove coverage, not a reason to clamp.
- The input slice is copied, so the caller's slice is still untouched now that there is no filter pass.
- The call site passes `quorumInfo.Aq` instead of `ensembleCoverage`. The parameter was named `ackQuorum` but received `Es-Aq+1`; the two mean different things. `aq >= ec` always holds, so requiring `aq` responses still guarantees the fence intersects every possible ack quorum, and for the reachable ensemble sizes (`GetEnsembleSize()` returns 3 or 5) `ec == aq`, so the threshold value is unchanged.

## Tests

The existing table asserted the buggy answers, so this is not a coverage gap being filled but wrong expectations being corrected:

| results | aq | before | after |
|---|---|---|---|
| `[0,-1]` | 2 | `0` | `-1` |
| `[2,-1]` | 2 | `2` | `-1` |
| `[3]` | 2 | `3` | error |
| `[]` | 2 | `-1` | error |

All other rows are unchanged. Added:

- a regression row for the production shape `[0,-1,-1] aq=2 -> -1`;
- `TestResolveFenceLACCoverageInvariant`, an exhaustive check over all response vectors of length 1..4 drawn from `{-1,0,1,3}` and every valid `ackQuorum`, asserting the target is covered by at least `ackQuorum` responders **and** is the largest value with that coverage — so the fix cannot drift into truncating acked data either.

`go test ./woodpecker/segment/ ./woodpecker/log/ ./woodpecker/client/` passes.

## Relationship to #92

d54940e (#92) added the filter to fix the opposite direction: an empty/restarted node's `-1` dragging LAC down and causing readers to EOF. That traded a truncation bug for an over-estimation bug, which stayed silent until #250 started enforcing qualified coverage.

Removing the filter is correct provided a `-1` from a **successful** fence is truthful. `segmentProcessor.Fence` opens the writer in recover mode (`getOrCreateSegmentWriter(ctx, true)`) and returns an error rather than a value when recovery fails, so a `-1` that reaches this function means the replica genuinely holds nothing. The new property test pins the maximality half of the invariant, which is exactly the direction #92 cared about.

## Rollout note

This does not by itself repair segments already poisoned by the old behaviour. `Finalize` records `footer.LAC = lac` (the requested value, not the local tail), so the behind replicas in #252 hold `footer.LAC=0` while returning `-1`.

- Against a pre-#250 server, a finalized writer returns `(localLastEntryId, nil)` for any later `Finalize`, so once fence computes `-1` those segments complete and self-heal.
- Against a #250 server, `recoveredFooter.LAC != lac` returns `ErrInvalidLACAlignment`, so lowering the target is refused and the segment stays stuck.

Lowering a completion target is always safe — it retracts an assertion that should never have been published — while raising it is not; the server-side check currently refuses both. Tracked separately, along with a `wpcli` metadata-correction tool for existing stuck segments.

### DIFF
--- a/woodpecker/segment/segment_handle.go
+++ b/woodpecker/segment/segment_handle.go
@@ -1425,42 +1425,44 @@ func (s *segmentHandleImpl) completeSegmentOnNode(ctx context.Context, node stri
 	}
 }
 
-// calculateLAC calculates the Log All Committed (LAC) based on the quorum responses.
-// LAC is the highest entry ID that is guaranteed to be committed on a majority of nodes.
-// For example: if nodes return [4, 6, 7] and ackQuorum is 2, then LAC is 6
-// because entries 1-6 are committed on at least 2 nodes (majority).
-func (s *segmentHandleImpl) calculateLAC(results []int64, ackQuorum int) int64 {
-	// Filter out negative values (e.g. -1 returned by nodes with no data).
-	// Only non-negative entry IDs represent real committed data.
-	validResults := make([]int64, 0, len(results))
-	for _, r := range results {
-		if r >= 0 {
-			validResults = append(validResults, r)
-		}
-	}
-	if len(validResults) == 0 {
-		return -1
+// resolveFenceLAC picks the completion target from the fence responses: the highest
+// entry ID that at least ackQuorum responding replicas hold. Sorted ascending, that is
+// the ackQuorum-th largest element, so exactly ackQuorum replicas cover it and no
+// larger value does.
+//
+// Every element is a successful fence response -- failed nodes take the result.err
+// branch and never reach here -- so a -1 is an honest "I hold nothing" vote against
+// every entry ID >= 0 and must be counted. Dropping it would shrink the sample while
+// ackQuorum stays fixed, and the resulting target could exceed what any ack quorum
+// covers; completion would then retry forever, because a behind replica has no way to
+// catch up to an entry it never received.
+//
+// Fewer than ackQuorum responses means no value can be shown to have quorum coverage,
+// which is a failure to prove rather than a reason to pick a smaller element.
+func (s *segmentHandleImpl) resolveFenceLAC(results []int64, ackQuorum int) (int64, error) {
+	if len(results) < ackQuorum {
+		return -1, werr.ErrAppendOpQuorumFailed.WithCauseErrMsg(
+			fmt.Sprintf("insufficient successful fence responses: got %d, need %d", len(results), ackQuorum))
 	}
 
-	// Sort in ascending order
-	sort.Slice(validResults, func(i, j int) bool {
-		return validResults[i] < validResults[j]
+	// Sort a copy in ascending order, leaving the caller's slice untouched.
+	sorted := make([]int64, len(results))
+	copy(sorted, results)
+	sort.Slice(sorted, func(i, j int) bool {
+		return sorted[i] < sorted[j]
 	})
 
-	// LAC is the (len - ackQuorum)th element, ensuring at least ackQuorum nodes
-	// have committed up to this entry.
-	lacIndex := len(validResults) - ackQuorum
-	if lacIndex < 0 {
-		lacIndex = 0
-	}
-	return validResults[lacIndex]
+	return sorted[len(sorted)-ackQuorum], nil
 }
 
 // fenceSegmentQuorum sends FenceSegment requests to all quorum nodes
 // and calculates the Log All Committed (LAC) based on majority consensus
 func (s *segmentHandleImpl) fenceSegmentQuorum(ctx context.Context, quorumInfo *proto.QuorumInfo) (int64, error) {
-	nodeCount := len(quorumInfo.Nodes)                         // always es=len(nodes)
-	ensembleCoverage := int(quorumInfo.Es - quorumInfo.Aq + 1) // currently always es=3=wq,aq=2, so ec=es-aq+1=2=aq
+	nodeCount := len(quorumInfo.Nodes) // always es=len(nodes)
+	// Requiring aq responses subsumes the ensemble-coverage bound ec=es-aq+1: aq >= ec
+	// always holds, so the fence still intersects every possible ack quorum. For the
+	// reachable ensemble sizes (3 and 5) the two are equal, so the threshold is unchanged.
+	ackQuorum := int(quorumInfo.Aq)
 
 	// Channel to collect results from all nodes
 	resultChan := make(chan nodeFenceResult, nodeCount)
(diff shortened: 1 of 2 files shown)


czs007:

Verified:
- The high-severity finding raised earlier was re-examined by an independent reviewer and confirmed as agreed, with no further issues surfacing in that area.
- A second independent reviewer conducted a full re-review pass and concurred with the existing assessment, surfacing no additional issues.

PR #264, opened 2026-08-04
fix(server): refuse segment lifecycle operations on a decommissioned node

### LINKED ISSUE #257: [bug] Fence/complete path re-creates an empty staged file on a decommissioned node, flipping safe_to_terminate back to false forever
## Summary

A node that has already latched the terminal `decommissioned` state can be handed a **brand-new, empty staged `data.log`** by the fence/complete path. `HasLocalSegmentData()` then reports `true` again, so `has_local_data` flips back to `true` and `safe_to_terminate` back to `false` — permanently, because the decommission monitor has already exited and never re-checks.

For an operator draining a node via `wp node drain-status`, this means: the node reports `decommissioned` / `safe_to_terminate: true`, and then flips back to `safe_to_terminate: false` and stays there forever.

## Root cause

`Server.Decommission()` (`server/service.go:837`) calls `s.logStore.RejectNewWrites()`. But in `server/logstore.go`, `rejectWrites` is only consulted by two entry points:

| Line | Method | Guards `rejectWrites`? |
|------|--------|------------------------|
| 248 | `AddEntry` | yes |
| 284 | `AddEntryBatch` | yes |
| 482 | `CompleteSegment` | **no** |
| 508 | `FenceSegment` | **no** |

The client calls `FenceSegment` / `CompleteSegment` against **all** quorum nodes when rolling a segment. On a node that holds no data for that segment, this creates a segment processor and a staged file writer in recovery mode, which creates the `data.log` on disk. `HasLocalSegmentData()` scans for exactly such files.

## Evidence

Observed in nightly run [30765806633](https://github.com/zilliztech/woodpecker/actions/runs/30765806633) (2026-08-02), failing `TestServiceUpgrade_DecommissionCompleteAfterDataCleanup` at `tests/integration/e2e_service_upgrade_test.go:459`.

The decommissioned node in that run was **node3** (`e2e_service_upgrade_test.go:389: Node 3 has local segment data, decommissioning it`). Progress readings from the drain loop:

```
check: state=decommissioning, hasLocalData=true,  truncateAt=seg26, remainingProcessors=6, safeToTerminate=false
check: state=decommissioning, hasLocalData=false, truncateAt=seg28, remainingProcessors=6, safeToTerminate=true
check: state=decommissioned,  hasLocalData=true,  truncateAt=seg30, remainingProcessors=7, safeToTerminate=false
```

Between the drain at seg28 and the check at seg30, node3 acquired a file for the **new** segment 29 — it had no data for it and never accepted an append:

```
20:44:28.434 [DEBUG] [stagedstorage/reader_impl.go:121] local staged file absent and no compacted ... /node3/.../51/29/data.log
20:44:28.434 [DEBUG] [stagedstorage/writer_impl.go:189] segment directory and file path prepared   ... /node3/.../51/29/data.log
20:44:28.436 [INFO]  [stagedstorage/writer_impl.go:319] staged file writer created successfully
                     [logId=51] [segmentId=29] [recoveryMode=true] [startEntryId=0] [recovered=false]
20:44:28.466 [DEBUG] [stagedstorage/writer_impl.go:1264] finalized staged file [lastEntryId=-1] [writtenBytes=78]
```

`lastEntryId=-1` / `writtenBytes=78` — a header-and-footer-only file with zero entries. `rejectWrites` did its job on the append path; the fence path simply doesn't consult it.

## Relationship to #203

This is the same test as #203 (closed). #203 covered the *never drains* failure mode and its fix added the rolling-truncation loop. This is a different, residual window: the node **does** drain, latches `decommissioned`, and is then re-populated by a fence on a segment it holds no data for. The #203 fix does not cover it.

Frequency: 1 in 15 nightly runs — it needs the fence to land on the decommissioned node in the same iteration that the state flips.

## Open question — how to fix

Three candidate directions, listed for discussion rather than as a decision:

1. **Guard the fence/complete path** — have `FenceSegment` / `CompleteSegment` avoid materialising a staged file on a node in `decommissioning`/`decommissioned`. Most direct, but needs care that fencing correctness (quorum LAC resolution) is not weakened — cf. #252/#253.
2. **Make `HasLocalSegmentData()` ignore empty segment files** — narrower, but "empty" needs a precise definition and it papers over the node still being handed work.
3. **Stop latching** — have the monitor keep re-checking after marking decommissioned. Larger semantic change to the lifecycle state machine; probably the wrong lever.

The e2e assertion at `e2e_service_upgrade_test.go:457-460` should not simply be relaxed to an `Eventually` — that would hide the operator-visible behaviour above.

tinswzy:
Fixes #257.

## The defect

A node that had already reached the terminal `decommissioned` state could still be handed a **brand-new** segment and would serve its fence/complete, creating a staged `data.log` for a segment it holds no data for. `HasLocalSegmentData` counted that file, so `has_local_data` flipped back to `true` and `safe_to_terminate` to `false` — **permanently**, because the decommission monitor exits on transition and never re-checks.

Operationally: `wp node drain-status` reports `decommissioned` / `safe_to_terminate: true`, then flips back to `false` and stays stuck forever.

## What actually happened

From nightly run [30765806633](https://github.com/zilliztech/woodpecker/actions/runs/30765806633):

```
20:44:21.399  start decommission, rejectWrites on
20:44:21.579  gossip broadcast status=decommissioning
20:44:26.580  hasLocalSegmentData=false -> marked decommissioned, monitor returns
20:44:28.422  client asks seed 127.0.0.1:36637 for a 3-node quorum
              -> endpoints=[127.0.0.1:36637, 127.0.0.1:37033, 127.0.0.1:38583]   (:38583 = node3)
20:44:28.425  segment 29 created with node3 in its ensemble
20:44:28.434  node3 creates data.log for seg29   (fence -> getOrCreateSegmentWriter(recovery))
20:44:28.466  complete -> Finalize writes header+footer = 78 bytes
```

**node3 never took a single entry for segment 29.** `rejectWrites` did its job: the fence answered `lastEntryId=-1`, `allResults="[-1,0,0]"`, and the completion quorum excluded it —

```
"Replica finalized but does not cover target LAC; excluding it from completion quorum"
  node=127.0.0.1:38583  localLastEntryId=-1
  finalizedCount=3  qualifiedCount=2  ackQuorum=2
```

It contributed nothing to the quorum and still left the file behind. The file was pure side effect.

### Why it was selected at all

Gossip is eventually consistent. There are exactly five `Server updated: node3 (Version: 2)` events in that run: **four at 20:44:21, and a fifth at 20:44:28** — one peer lagged seven seconds, and the client happened to ask that straggler for a quorum. `excludeDecommissioning` is correct and is applied on every selection path including the gRPC seed handler; it simply cannot see what has not arrived yet.

Exclusion shortens that window but can never close it, so the authoritative enforcement has to live on the node itself.

## The fix

Split enforcement by state, matching what each state means:

| | `decommissioning` | `decommissioned` |
|---|---|---|
| AddEntry / AddEntryBatch | refused | refused |
| Fence / Complete / Compact / UpdateLAC | **served** | **refused** |
| Clean / Evict / compacted-mark | served | served |
| reads | served | served |

Keeping fence/complete/compact open while draining is deliberate and load-bearing: that is exactly how the node's own segments get closed, uploaded and cleaned so it can reach the terminal state at all. Once retired there is nothing left to close, and continuing to serve them is what causes the bug.

Refusals return the new **non-retryable** `ErrLogStoreRetired`, so a caller drops the node from the quorum and re-discovers rather than backing off and retrying against a node that will never accept it again.

Cleanup and reads stay open on purpose: they remove data or create nothing, and blocking them would strand whatever the node still holds locally (it can retire with compacted-but-not-yet-GC'd files, which `HasLocalSegmentData` deliberately skips).

Quorum impact is nil for the observed case — fence `successCount` 3→2 with `ackQuorum=2`, and `qualifiedCount` was already 2 because node3 had been excluded anyway.

### Second, independent hole: the restart path

`Start()` keyed off `IsDecommissioning()`, which matches **only** the draining state. A node persisted as `decommissioned` therefore restarted with no write rejection, no retirement, and no status tag broadcast at all — back in the cluster indistinguishable from an active node, accepting appends. That is now driven off the actual persisted state.

The three duplicated gossip-tag blocks are folded into one `broadcastNodeStatus` helper, and the restart branch is extracted to `restoreLifecycleEnforcement()` so it is directly testable.

## Verification

`go test ./server/... ./common/werr/...` passes. `golangci-lint run --config .golangci.yml` reports **0 issues** (run with v2.11.4, the same version CI installs). The `LogStore` mock was regenerated with mockery, adding only the new method (+32 lines, nothing else touched).

New tests, each pinned to a specific claim above:

- `TestLogStore_Retired_RefusesSegmentLifecycleOps` — fence / complete / compact / updateLAC all return `ErrLogStoreRetired`
- `TestLogStore_Decommissioning_StillServesSegmentLifecycleOps` — the draining node still serves fence and complete (against a mock processor) while appends are already refused; this is the property that must not regress
- `TestLogStore_Retired_AllowsCleanupAndReads` — clean and reads reach the processor
- `TestLogStore_Retired_AlsoRejectsWrites` — retiring implies write rejection
- `TestLogStore_Retired_NotClearedByAllowNewWrites` — one-way latch
- `TestErrLogStoreRetired_NotRetryable`
- `TestServer_RestoreLifecycleEnforcement` — table over active / decommissioning / decommissioned, asserting write rejection, retirement and whether the monitor resumes
- `TestServer_DecommissionMonitor_RetiresLogStoreOnTransition` — the monitor retires the store before exiting, which is the only chance it gets
- `TestNodeLifecycleManager_IsDecommissionedOnlyInTerminalState` and an extension of the existing persistence test asserting `IsDecommissioning()` and `IsDecommissioned()` are not conflated

## Not done, deliberately

`HasLocalSegmentData` still counts a zero-entry finalized file as local data. That was considered as a belt-and-braces second layer and rejected: the `decommissioning` window is self-healing (the monitor keeps polling and truncation drives `CleanSegment`, which removes the file), so only the terminal state — where the monitor has already exited — needs a gate.

An earlier idea of not writing the file at all was also rejected: an empty footer is not equivalent to no footer. The footer is how a replica self-describes "this segment is closed here and holds zero entries", and it is read back for idempotent re-finalize and compact safety checks.

### DIFF
--- a/common/werr/errors.go
+++ b/common/werr/errors.go
@@ -98,6 +98,10 @@ var (
 	ErrMarkDeleteFailed        = newWoodpeckerError("failed to durably mark log/instance deleted", 2006, true)
 	ErrLogStoreDiskPressure    = newWoodpeckerError("logstore local disk above hard watermark, write backpressure active", 2007, true)
 	ErrLogStoreInvalidRootPath = newWoodpeckerError("invalid rootPath", 2008, false)
+	// ErrLogStoreRetired is returned by a decommissioned node for segment-lifecycle operations.
+	// Non-retryable on purpose: the node is terminally retired, so the caller should drop it from
+	// the quorum and re-discover rather than back off and retry against it.
+	ErrLogStoreRetired = newWoodpeckerError("logstore is retired (node decommissioned)", 2009, false)
 
 	// segment_processor errors (2100-2199)
 	ErrSegmentProcessorNotFound          = newWoodpeckerError("segment processor not found", 2100, false)
--- a/mocks/mocks_server/mock_logstore.go
+++ b/mocks/mocks_server/mock_logstore.go
@@ -934,6 +934,38 @@ func (_c *LogStore_HasLocalSegmentData_Call) RunAndReturn(run func() bool) *LogS
 	return _c
 }
 
+// MarkRetired provides a mock function with no fields
+func (_m *LogStore) MarkRetired() {
+	_m.Called()
+}
+
+// LogStore_MarkRetired_Call is a *mock.Call that shadows Run/Return methods with type explicit version for method 'MarkRetired'
+type LogStore_MarkRetired_Call struct {
+	*mock.Call
+}
+
+// MarkRetired is a helper method to define mock.On call
+func (_e *LogStore_Expecter) MarkRetired() *LogStore_MarkRetired_Call {
+	return &LogStore_MarkRetired_Call{Call: _e.mock.On("MarkRetired")}
+}
+
+func (_c *LogStore_MarkRetired_Call) Run(run func()) *LogStore_MarkRetired_Call {
+	_c.Call.Run(func(args mock.Arguments) {
+		run()
+	})
+	return _c
+}
+
+func (_c *LogStore_MarkRetired_Call) Return() *LogStore_MarkRetired_Call {
+	_c.Call.Return()
+	return _c
+}
+
+func (_c *LogStore_MarkRetired_Call) RunAndReturn(run func()) *LogStore_MarkRetired_Call {
+	_c.Run(run)
+	return _c
+}
+
 // NotifySegmentCompacted provides a mock function with given fields: ctx, bucketName, rootPath, logId, segmentId
 func (_m *LogStore) NotifySegmentCompacted(ctx context.Context, bucketName string, rootPath string, logId int64, segmentId int64) error {
 	ret := _m.Called(ctx, bucketName, rootPath, logId, segmentId)
--- a/server/lifecycle.go
+++ b/server/lifecycle.go
@@ -101,6 +101,15 @@ func (m *NodeLifecycleManager) IsDecommissioning() bool {
 	return m.state == NodeStateDecommissioning
 }
 
+// IsDecommissioned returns true if the node is in the terminal decommissioned state.
+// Distinct from IsDecommissioning: the two states enforce different things (writes are
+// refused in both, segment-lifecycle operations only once decommissioned).
+func (m *NodeLifecycleManager) IsDecommissioned() bool {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+	return m.state == NodeStateDecommissioned
+}
+
 // StartDecommission transitions the node from active to decommissioning.
 // Idempotent if already decommissioning. Returns error if already decommissioned.
 func (m *NodeLifecycleManager) StartDecommission() error {
--- a/server/lifecycle_test.go
+++ b/server/lifecycle_test.go
@@ -165,6 +165,23 @@ func TestNodeLifecycleManager_Persistence_DecommissionedSurvivesRestart(t *testi
 	m2, err := NewNodeLifecycleManagerWithPersistence(dir)
 	require.NoError(t, err)
 	assert.Equal(t, NodeStateDecommissioned, m2.GetState())
+	// The restart path keys off these two predicates to decide what to re-enforce, and they
+	// must not be conflated: only the terminal state also refuses segment lifecycle ops.
+	assert.False(t, m2.IsDecommissioning(), "decommissioned is not decommissioning")
+	assert.True(t, m2.IsDecommissioned())
+}
+
+func TestNodeLifecycleManager_IsDecommissionedOnlyInTerminalState(t *testing.T) {
+	m := NewNodeLifecycleManager()
+	assert.False(t, m.IsDecommissioned(), "active")
+
+	m.StartDecommission()
+	assert.True(t, m.IsDecommissioning())
+	assert.False(t, m.IsDecommissioned(), "draining is not yet retired")
+
+	require.NoError(t, m.MarkDecommissioned())
+	assert.False(t, m.IsDecommissioning())
+	assert.True(t, m.IsDecommissioned())
 }
 
 func TestNodeLifecycleManager_Persistence_ClearState(t *testing.T) {
--- a/server/logstore.go
+++ b/server/logstore.go
@@ -73,6 +73,7 @@ type LogStore interface {
 	GetActiveProcessorCount() int
 	RejectNewWrites()
 	AllowNewWrites()
+	MarkRetired()
 	HasLocalSegmentData() bool
 	EvictLog(ctx context.Context, bucketName string, rootPath string, logId int64) error
 	EvictInstance(ctx context.Context, bucketName string, rootPath string) error
@@ -110,8 +111,11 @@ type logStore struct {
 	maintenance      *NodeMaintenanceManager
 	compactedCleanup *compactedFileCleanupTask // event queue fed by NotifySegmentCompacted; drop path for compacted data.log
 	stopped          atomic.Bool
-	rejectWrites     atomic.Bool  // separate from stopped: only blocks new writes during decommission, not reads
-	diskRejectBps    atomic.Int32 // rejection probability in basis points (0..10000), set by diskWatermarkTask; gates new appends only
+	rejectWrites     atomic.Bool // separate from stopped: only blocks new writes during decommission, not reads
+	retired          atomic.Bool // set on reaching the terminal decommissioned state; additionally refuses
+	// segment-lifecycle participation (fence/complete/compact/LAC). One-way latch:
+	// decommissioned is terminal, CancelDecommission refuses it.
+	diskRejectBps atomic.Int32 // rejection probability in basis points (0..10000), set by diskWatermarkTask; gates new appends only
 }
 
 // admitAppend returns false when the disk-watermark policy rejects this append.
@@ -483,6 +487,9 @@ func (l *logStore) CompleteSegment(ctx context.Context, bucketName string, rootP
 	if l.stopped.Load() {
 		return -1, werr.ErrLogStoreShutdown
 	}
+	if l.retired.Load() {
+		return -1, werr.ErrLogStoreRetired
+	}
 	ctx, sp := logger.NewIntentCtxWithParent(ctx, LogStoreScopeName, "CompleteSegment")
 	defer sp.End()
 	logIdStr := strconv.FormatInt(logId, 10)
@@ -508,6 +515,9 @@ func (l *logStore) CompleteSegment(ctx context.Context, bucketName string, rootP
 func (l *logStore) FenceSegment(ctx context.Context, bucketName string, rootPath string, logId int64, segmentId int64) (int64, error) {
 	if l.stopped.Load() {
 		return -1, werr.ErrLogStoreShutdown
+	}
+	if l.retired.Load() {
+		return -1, werr.ErrLogStoreRetired
 	}
 	ctx, sp := logger.NewIntentCtxWithParent(ctx, LogStoreScopeName, "FenceSegment")
 	defer sp.End()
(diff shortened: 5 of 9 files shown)


czs007:
> Also reported: 1 pre-existing hazard (real, but not introduced here — follow-up only, does **not** block this PR).

Adversarial review of this PR found two issues attributable to the change itself, both low severity, plus four real but pre-existing hazards listed separately for follow-up.

## Low

**`server/logstore.go:805` — `AllowNewWrites` does not consult `retired`, so the "one-way latch" is guarded by the caller rather than by the latch**
`MarkRetired` establishes the invariant `retired ⇒ rejectWrites` (`logstore.go:831-837`), but `AllowNewWrites` only performs `l.rejectWrites.Swap(false)` and never reads `retired`, so it can unilaterally break that invariant: after `AllowNewWrites()` a retired store admits `AddEntry` again and can materialise a fresh `data.log` (fence/complete stay refused). This is not reachable in production today — the only production caller, `Server.CancelDecommission` (`service.go:811-816`), is guarded by `lifecycle.CancelDecommission()`, which rejects the terminal state (`lifecycle.go:184-186`), and `ClearState()` has no production caller. The new test at `logstore_retired_test.go:134-146` is titled "Retirement is a one-way latch" but only re-checks `FenceSegment`, so it does not cover the half that is actually breakable. Suggestion: add `if l.retired.Load() { return }` at the top of `AllowNewWrites`, and add an `AddEntry` assertion to that test so the code matches the invariant the comment and test title claim. *(raised by czs007)*

**`server/service.go:819` — one of the three gossip-tag blocks the PR description claims to fold was not migrated**
The PR description states "The three duplicated gossip-tag blocks are folded into one `broadcastNodeStatus` helper", but only `start()` and `Decommission()` (`service.go:850`) were migrated; `CancelDecommission` (`service.go:819-835`) still carries its own inline copy of the same five steps (`GetMeta` / copy tags / set `status` / `UpdateMeta` / log), differing only in map pre-allocation and in having no return value. `gh pr diff 264` contains no hunk touching `CancelDecommission`. This is more than duplication: the inline copy silently skips the broadcast when `node == nil` and gives the caller no signal, so nothing can decide to retry. Suggestion: either migrate it to `broadcastNodeStatus` so the boolean return is available, or amend the PR description to say two. *(surfaced during verification)*

- **`server/logstore.go:251` — the decommission gate in `AddEntry`/`AddEntryBatch` is checked once at entry, so an already-admitted append can make `data.log` non-empty after retirement.** Both methods do a single `if l.stopped.Load() || l.rejectWrites.Load()` (`:251-254`, `:287-290`); an append admitted microseconds before `rejectWrites` is set synchronously creates the `StagedFileWriter` — `os.OpenFile(..., O_CREATE|O_TRUNC)` at `writer_impl.go:280` materialises a 0-byte `data.log`, which `HasLocalSegmentData` skips because it only counts `info.Size() > 0` (`logstore.go:862`) — while the entry sits in the buffer awaiting async flush → the monitor's next tick sees `hasData=false`, runs `MarkDecommissioned` + `MarkRetired` and returns at `service.go:983`, then the async flush writes into that `data.log`, flipping `has_local_data` back to true and `safe_to_terminate` to false with no monitor left to re-check, i.e. the node is permanently un-drainable. Unrelated to this PR: the PR touches `logstore.go` only to add `MarkRetired` (`:76`, `:831-837`), the `retired` field (`:114-118`) and the four lifecycle-op gates (`:490/519/622/759`) — the `AddEntry`/`AddEntryBatch` entry checks and `rejectWrites` semantics are unchanged, and the race holds identically before the PR. *(raised by chyezh)*
- **`server/service.go:954` — the same entry-only gating leaves a millisecond-wide window on the fence/complete path.** Fence goes through `getOrCreateSegmentWriter(recovery=true)`, whose constructor opens `data.log` with `O_CREATE|O_APPEND` (`writer_impl.go:276`) at 0 bytes — invisible to the `Size() > 0` scan — and content only appears at `Complete → writer.Finalize` (`segment_processor.go:190-200`). A straggler peer's `FenceSegment` passes the entry gate and creates the empty file → the monitor's `HasLocalSegmentData()` at `service.go:954` returns false → `MarkDecommissioned` persists → the matching `CompleteSegment` (~30ms later, per the PR's own timeline) still reads `retired == false` and `Finalize` writes header+footer → `MarkRetired` runs afterwards and the monitor returns at `:983`, leaving `has_local_data` permanently true and `wp node drain-status` stuck at not-safe-to-terminate. Unrelated to this PR: the failure mode *is* #257 — the `Size() > 0` criterion and the unconditional monitor `return` with no re-check are untouched existing code, and the new gate strictly narrows the window rather than opening a path. A cheap, self-healing follow-up: re-run `HasLocalSegmentData()` once after `MarkRetired()`, and if it comes back true, log/emit a metric and keep the monitor loop alive instead of returning — that also covers the append race above and any future operation that materialises local state. *(raised by czs007)*
- **`server/service.go:271` — the restart broadcast publishes a state snapshot captured once, and the monitor's broadcast result is discarded.** `restoreLifecycleEnforcement` captures `state := s.lifecycle.GetState()` and the retry goroutine republishes that captured value forever (`:284`), while `broadcastNodeStatus` returns false whenever `s.serverNode` is still nil (`:864-870`) — and `waitAndStartCurrentNode` can take up to 60s to set it (`:1033-1104`). Tags are never periodically re-emitted (`server_node.go:176-179`), so a wrong value stays wrong until process restart. A node restarts with persisted `decommissioning` before gossip is up; at t=3s the operator runs `wp node cancel-decommission`, so lifecycle flips to active and writes are re-enabled, but `CancelDecommission`'s own broadcast is skipped because `node == nil` (`:819-822`); at t≈10s gossip comes up and the restart goroutine publishes the captured `"decommissioning"` → an active node that is accepting appends advertises `decommissioning`, and `excludeDecommissioning` (`service_discovery.go:1138`) permanently drops it from quorum selection, silently costing the cluster one node of capacity. Unrelated to this PR: the damaging half has the same shape pre-PR — the old code started the same goroutine under `IsDecommissioning()` with a hard-coded `"decommissioning"` tag and the same `node != nil` skip in `CancelDecommission`; this PR only folds it into `broadcastNodeStatus`, which is an equivalent rewrite for that state. A follow-up fix would publish `string(s.lifecycle.GetState())` in the loop instead of the captured value, return without publishing once the state is active, and route `CancelDecommission` through the same retrying helper. *(raised by czs007)*
- **`server/service.go:253` — on restart the retired gate is applied after `logStore.Start()`, with a full directory walk in between.** The order is `s.logStore.Start()` (`:253`) → `s.updateDecommissionMetrics()` (`:256`, which runs `HasLocalSegmentData`'s `filepath.WalkDir` over the whole root path) → `restoreLifecycleEnforcement()` (`:259`). gRPC is already listening and the gossip node was created by `Prepare()`'s async goroutine carrying only the static config tags, with no `status` tag yet. On a node persisted as `decommissioned` with a large data directory, a peer sends `FenceSegment`/`CompleteSegment` during that walk → both are served before `MarkRetired` takes effect, creating and finalising a non-empty `data.log` for a segment this node holds no data for, and since the terminal state no longer starts the monitor, `has_local_data` is permanently true — the same symptom as #257. Unrelated to this PR: before it, a restarted `decommissioned` node had no gate at all (that is the second hole this PR closes), so the window was unbounded; this is late gate placement, not a regression. `MarkRetired` only sets two atomics and has no dependency on `logStore` being started, so it can be hoisted ahead of `Start()`. *(surfaced during verification)*

PR #266, opened 2026-08-11
fix(meta): keep reader temp info alive with a session lease

### LINKED ISSUE #265: [bug] Reader temp info silently expires 60s after open: the lease is never renewed and update can never recreate it
## Summary

Every reader's temp info silently disappears from etcd **60 seconds after the reader opens**, and can never come back for the lifetime of that reader.

`CreateReaderTempInfo` attaches the key to a 60s lease, but nothing ever renews that lease: an etcd `Put` with `WithLease` re-binds the key to the lease — it does **not** refresh the lease TTL (only `LeaseKeepAlive` does). Once the lease expires, `UpdateReaderTempInfo` (a strict get-then-put) fails with `reader temp info not found` on every call, forever, because the only creation path is `OpenLogReader`.

Reader temp info is the only input that protects truncated segments still needed by lagging readers from physical cleanup. That protection silently switches off after 60s — for exactly the readers (slow/stalled consumers) it exists to protect.

## Root cause

Timeline for any reader, with the current code:

```
t=0     OpenLogReader → CreateReaderTempInfo: Put with fresh 60s lease   (meta/metadata_etcd.go)
t=30s   UpdateReaderTempInfo: Get → Put WithLease(existing)  ← does NOT renew the lease TTL
t=60s   lease expires, etcd deletes the key
t>60s   every UpdateReaderTempInfo: Get returns 0 kvs → "reader temp info not found" warn, every ~30s, forever
```

Three code facts combine into the defect:

1. `CreateReaderTempInfo` grants a fixed 60s lease (`meta/metadata_etcd.go`). The only `KeepAliveOnce` in the codebase belongs to the **writer lock** session (`CheckSessionLockAlive`); the reader lease has no keepalive of any kind.
2. `UpdateReaderTempInfo` re-puts with the existing lease ID. In etcd, `Put ... WithLease(id)` attaches the key to the lease without resetting the lease's TTL countdown — verified against etcd 3.6: three puts at t=0/2/4s on a 5s lease do not extend it one second; the lease expires exactly at t=5s and takes the key with it.
3. `UpdateReaderTempInfo` returns an error when the key is gone (`len(resp.Kvs) == 0`) instead of recreating it, so the state is unrecoverable without reopening the reader.

The update cadence in `woodpecker/log/log_reader.go` (`UpdateReaderInfoIntervalMs = 30000`, i.e. 2× inside the TTL window) shows the design intent was "periodic update acts as the heartbeat" — but the update never actually renews anything.

## Impact

- The truncated-segment cleanup task (`woodpecker/log/log_writer.go`, via `GetAllReaderTempInfoForLog`) computes the min segment still in use from readers' `RecentReadSegmentId`. After 60s it no longer sees any long-lived reader, so truncated segments still needed by a lagging reader can be physically deleted.
- Every reader older than 60s logs `reader temp info not found` / `update reader info failed` every ~30s — chronic log noise that masquerades as an incident signal during triage.
- Observed in the field as part of milvus-io/milvus#52341: after a `3.0.0 → newer 3.0 → 3.0.0` rollback under continuous DML, StreamingNode scanners stall permanently on channels whose segments can no longer be read, with these warnings flooding the logs. The dead reader-protection path is a contributing factor to that scenario (the stuck-retry behavior on `no record extract` is a separate defect and is not addressed here).

## Proposed fix

Separate **liveness** from **progress**, mirroring the existing writer-lock pattern:

1. `CreateReaderTempInfo` opens a `concurrency.Session` (TTL 60s, automatic keepalive at ~TTL/3) and attaches the key to the session lease. Liveness no longer depends on read progress, so idle/slow readers keep their metadata; a crashed process still auto-expires within the TTL.
2. `UpdateReaderTempInfo` becomes an unconditional upsert built from the reader's cached open position, gated on the in-process session entry: a reader this process owns self-heals its key after lease loss (e.g. an etcd partition outliving the TTL), but a closed reader can never resurrect its key — a resurrected key with a live keepalive would pin the cleanup low-watermark forever.
3. `DeleteReaderTempInfo` / provider `Close()` close the session (revoking the lease removes the key immediately), keeping the explicit etcd delete as a fallback.

No key layout or value format changes; mixed-version deployments keep working (keys written by old clients keep the old 60s-expiry behavior).

tinswzy:
## Summary

Reader temp info was created with a fixed 60s lease that nothing ever renewed — an etcd `Put` with `WithLease` re-binds the key but does **not** refresh the lease TTL — so every reader older than 60s silently lost its metadata, and every subsequent `UpdateReaderTempInfo` (a strict get-then-put) failed with `reader temp info not found` forever. Since this metadata is what protects truncated segments still needed by lagging readers from physical cleanup, the protection silently switched off for exactly the readers it exists for. See #265 for the full analysis; observed in the field as part of milvus-io/milvus#52341.

## Key changes

- **`CreateReaderTempInfo`** now guards the key with a `concurrency.Session` (automatic keepalive at ~TTL/3, mirroring the existing writer-lock pattern). Liveness is decoupled from read progress, so idle/slow readers keep their metadata; a crashed process still auto-expires within the TTL. Reopening the same reader name swaps in a fresh session and closes the stale one.
- **`UpdateReaderTempInfo`** is now an unconditional upsert rebuilt from the reader's cached open position, gated on the in-process session entry:
  - an owned reader self-heals its key after lease loss (e.g. an etcd partition outliving the TTL), rebuilding the session and retrying once;
  - a reader without an in-process session (closed, or a key written by someone else) is rejected with the existing `reader temp info not found` error — blind adoption would resurrect a key with a live keepalive after close and pin the cleanup low-watermark forever.
- **`DeleteReaderTempInfo` / provider `Close()`** close the session first (revoking the lease deletes the key immediately), keeping the explicit etcd delete as a fallback for keys not owned by this process.
- The TTL is now the package variable `readerTempInfoSessionTTLSeconds` (default unchanged at 60) so tests can shrink it.

No key layout or value format changes; mixed-version deployments keep working (keys written by old clients keep the old 60s-expiry behavior).

## Testing

Written test-first (both regressions verified failing on master before the fix):

- `test reader temp info outlives lease ttl while idle` — an idle reader's metadata survives 2.5× the lease TTL and stays attached to a live lease (fails on master: the key expires).
- `test update reader temp info self heals after lease loss` — after the lease is revoked and the key removed, the next update recreates the key, preserves the open position, and re-attaches a live lease (fails on master: `reader temp info not found`).
- `test update after delete does not resurrect reader temp info` — a late update after close is rejected and the key stays deleted (guards the deliberate non-resurrection gate).
- Existing `testUpdateReaderTempInfoWithoutLease` renamed to `testUpdateReaderTempInfoWithoutSession` with inverted semantics to pin the new gate.
- `go test ./meta/` and `go test ./woodpecker/log/` pass; `golangci-lint run ./meta/...` reports 0 issues.

Known boundaries (deliberately out of scope): the stuck-retry behavior on `no record extract` from milvus-io/milvus#52341 is a separate defect; "process crash → key expires within TTL" relies on etcd session semantics and is not separately unit-tested.

Fixes #265


czs007:
You've hit your session limit · resets 11:20am (America/Los_Angeles)

czs007:
## Re-review of `97c15df802df`

No findings from this mechanism were published on this commit previously, so there is nothing from an earlier round to confirm or retract. Everything below is from this round.

---

This round's review found lifecycle problems in the new reader temp-info session handling in `meta/metadata_etcd.go` — failed writes and post-close updates can both leave a reader without metadata protection or leave an untracked session behind — plus one missing-timeout regression and one test-hygiene item.

## High

- **`meta/metadata_etcd.go:1368` — any failed `Put` revokes the session, deleting the live reader's temp info key.** The retry loop calls `session.Close()` unconditionally when `e.client.Put` returns an error, without distinguishing the error source. `concurrency.Session.Close()` is `Orphan()` + `Revoke(leaseID)`, and revoking makes etcd delete every key attached to that lease — including this reader's own temp info key. The most common `Put` failures have nothing to do with lease health: `ctx1` comes from `e.getContextWithTimeout(ctx)` (line 1358), so a cancelled caller context or an etcd slowdown past `e.requestTimeout` yields a context error while the lease is perfectly healthy. Both attempts derive from the same (already dead) parent context, so the second `Put` fails identically. *Failure: a `Put` fails because the caller's context was cancelled or etcd exceeded `e.requestTimeout` while the lease is healthy → the healthy lease is revoked and the reader temp info key is deleted; the reader keeps reading with no metadata protection, so concurrent truncated-segment cleanup can remove segments it has not yet consumed, and the key is only rebuilt on the next successful update — potentially never for an idle or lagging reader.* The caller at `woodpecker/log/log_reader.go:160-163` only logs a warning and keeps reading. Suggestion: rotate the session only when `session.Done()` is already closed or etcd explicitly reports the lease is gone (`rpctypes.ErrLeaseNotFound`); return context errors and other unrelated write errors without touching the existing lease. (raised by nsj)

- **`meta/metadata_etcd.go:1301` — TOCTOU between `Update`'s pre-lock `Load` and `Delete`'s `LoadAndDelete`: a closed reader's key can be resurrected onto an untracked session.** `UpdateReaderTempInfo` loads the entry pointer at line 1301 and only locks it at line 1309, while `DeleteReaderTempInfo` removes the entry from the map first (line 1402) and then locks it to close the session (lines 1404-1409). With the interleaving `Update.Load` → `Delete` completes fully (including the etcd delete at line 1415) → `Update` acquires the lock, the update sees a closed `session.Done()` (lines 1333-1337), builds a fresh session (lines 1343-1352), and re-`Put`s the key (line 1359) for a reader that is already closed. That new session hangs off an entry no longer in `readerTempSessions`, so neither `DeleteReaderTempInfo` nor the `readerTempSessions.Range` in `provider.Close()` (lines 1499-1509) can ever reach it. *Failure: one goroutine triggers `UpdateReaderTempInfo` from `ReadNext` (entry loaded, lock not yet taken) while another calls `reader.Close()` → `DeleteReaderTempInfo` runs to completion → the update rewrites the deleted key with a new session that nothing tracks; its keepalive renews until the process or etcd client exits, so the zombie key permanently pins the truncated-segment cleanup low-water mark and storage is never reclaimed.* This interleaving is reachable in supported usage: `logBatchReaderImpl.Close` (`woodpecker/log/log_reader.go:240-257`) only reads fields fixed at construction (`logHandle`, `logId`, `logName`, `readerName`, `logIdStr`, `logNs`), while `ReadNext` writes a disjoint set (`batch`, `next`, `lastRead`, `pendingRead*`, `currentSegmentHandle`, lines 128-229 and 299-357), so "read in one goroutine, close from another" is not itself a data race at the reader layer. Suggestion: set a `closed` flag under the entry mutex, and after acquiring the lock re-validate that the map still maps this key to the same, non-closed entry before recreating a session or writing; the existing test at `meta/metadata_etcd_reader_session_test.go:141-153` is sequential and should be extended with a deterministic concurrent interleaving. (raised by nsj)

## Medium

- **`meta/metadata_etcd.go:1161` — `concurrency.NewSession` gets no bounded context, so `CreateReaderTempInfo` ignores the caller's deadline when etcd is unreachable.** `concurrency.NewSession(e.client, concurrency.WithTTL(...))` is called without `concurrency.WithContext`, so etcd's `NewSession` performs its `Grant` on `client.Ctx()`, which is cancelled only when the etcd client itself is closed; combined with clientv3's default `grpc.WaitForReady(true)`, the RPC queues instead of failing fast. *Failure: etcd is unreachable or partitioned and the application calls `OpenLogReader` → `CreateReaderTempInfo` with a bounded deadline → `Grant` blocks indefinitely on `client.Ctx()`, the caller's deadline and cancellation are ignored, and `OpenLogReader` hangs until etcd recovers or the client is closed.* This is a regression relative to the previous `lease, err := e.client.Grant(ctx1, 60)` under `e.getContextWithTimeout(ctx)`, which returned `ErrMetadataWrite` after `e.requestTimeout` and let the caller retry or degrade. The same pattern appears in the update path at line 1343. The pre-existing writer-lock call at line 686 shares the shape, but that is a separate path and does not offset losing the timeout here. Suggestion: pass `concurrency.WithContext(ctx1)` (or at least a context bounded by `e.requestTimeout`) at both 1161 and 1343. (raised by xiaocai)

## Low

- **`meta/metadata_etcd_reader_session_test.go:84` — 8.5s unconditional sleep with no `testing.Short()` skip, in a package that `make unit-test` runs with `-short`.** `testReaderTempInfoOutlivesLeaseTTLWhileIdle` lowers the TTL to 3s and then sleeps `TTL*2.5 + 1s` = 8.5s, with no `if testing.Short() { t.Skip(...) }` guard. It is registered under `TestAll` at `meta/metadata_etcd_test.go:105` and uses embedded etcd (`common/etcd/etcd_util.go:44-46`), so it is inside the `make unit-test` (`go test -race -short -cover -failfast`) package set, which excludes only `tests/{integration,benchmark,stability,docker}`. Per the repo house rules, long tests should honour `testing.Short()`. A 3s TTL with a 1s heartbeat also leaves thin margin on a loaded `-race` runner, and with `-failfast` a flake truncates the rest of the suite. Suggestion: add the `testing.Short()` skip and raise the TTL to 5s or more. (surfaced during verification)

tinswzy:
@czs007 Replying to the [re-review of 97c15df802df](https://github.com/zilliztech/woodpecker/pull/266#issuecomment-5264739921):

All four findings confirmed and addressed in c2874da, each with a regression test that was verified failing on the previous commit first:

- **Failed Put revoking the healthy lease (High)** — the retry loop now rotates the session only when the lease itself is gone (`rpctypes.ErrLeaseNotFound` on the Put, or the session's `Done()` channel already closed). Unrelated failures (caller cancelled, etcd slow) return the error without touching the session. Regression test: `test update put failure does not revoke reader temp info lease` (update with a dead caller context must leave the key attached to a live lease).

- **Update/Delete TOCTOU resurrection (High)** — entries are now marked `closed` under their mutex on every retirement path (`DeleteReaderTempInfo`, same-name reopen in `CreateReaderTempInfo`, provider `Close()`), and the update re-validates after acquiring the lock that the entry is non-closed and still the map's current entry for that reader. The post-lookup body was extracted to `updateOwnedReaderTempInfo` so the test drives the exact interleaving deterministically: `test update racing with close does not resurrect reader temp info`.

- **Unbounded `concurrency.NewSession` (Medium)** — fixed, but deliberately not with `concurrency.WithContext(ctx1)`: the session's keepalive context derives from the option context, so a request-scoped context would kill the keepalive the moment the request ends and quietly reintroduce the original expiry bug (`WithContext`'s own doc notes the lease is then abandoned to expire). Instead, `newReaderTempSession` grants the lease under the request-scoped/bounded context and attaches the session to it via `WithLease`, keeping the keepalive on the client's lifetime context. Regression test: `test create reader temp info honors caller deadline when etcd unreachable` (previously hung indefinitely; now fails within the deadline).

- **Missing `testing.Short()` guard (Low)** — the idle-reader test now skips in short mode and uses a 5s TTL for more margin under `-race`.

Verified: `go test ./meta/` (full `TestAll`, including the pre-existing self-heal test which now exercises the `ErrLeaseNotFound` rotation path), `go test -race -short ./meta/`, `go test ./woodpecker/log/`, `golangci-lint run ./meta/...` — all green.

Known boundary unchanged: a concurrent same-name reopen while an update is mid-flight can still transiently drop the key until the next update (real reader names are unique per open, so this is not reachable in supported usage); it is narrowed but not eliminated by the closed-flag gate.

czs007:
## Overview

The reader temp-info lease rework is sound on the happy path, but keeping the lease alive indefinitely removes the only time bound on reader-held truncation protection, and the new `session.Close()` calls on the close paths can stall for a full TTL while ignoring the caller's context.

## Medium

**`meta/metadata_etcd.go:1151` — keepalive removes the only time bound on reader temp keys, so an unclosed reader pins physical cleanup forever**
The reader temp key moves from a dead 60s lease to a `concurrency.Session` lease renewed indefinitely on `client.Ctx()` (`:1151`, key bound via `session.Lease()` at `:1203`), and renewal is independent of read progress — only `DeleteReaderTempInfo` (`:1471`), a same-name reopen, or provider `Close` (`:1569`) stops it. The cleanup side applies no freshness test: `woodpecker/log/log_writer.go:524-527` and `woodpecker/log/internal_log_writer.go:448-451` fold every reader's `RecentReadSegmentId` into `minTruncatedSegmentId` unconditionally, and nothing in the repo ever reads `ReaderTempInfo.RecentReadTimestamp` (only writes, at `metadata_etcd.go:1179` and `:1377`); the `segId >= minTruncatedSegmentId` guard at `log_writer.go:554` runs ahead of the `ttlCutoffMs` check at `:571`. Failure: a reader whose `LogReader.Close()` is never called (dropped reader object, or a consumer stuck in a retry loop that no longer advances) keeps its key alive for the process lifetime → `minTruncatedSegmentId` stays pinned at that reader's stale position, all later segments are skipped by `log_writer.go:554`, already-truncated segments are never physically deleted, and object storage grows without bound (on master the pin self-releases after 60s). Suggestion: add a freshness threshold on `RecentReadTimestamp` in the cleanup paths (comfortably above `UpdateReaderInfoIntervalMs = 30s`), or give session renewal an upper bound on time-since-last-update, so a slow reader stays distinguishable from an abandoned one. *(raised by zack)*

**`meta/metadata_etcd.go:1475` — `session.Close()` on the reader-close paths ignores the caller context and can block a full TTL (60s)**
Three new `entry.session.Close()` calls were added: `DeleteReaderTempInfo` (`:1471-1480`, under `entry.Lock()`), the `CreateReaderTempInfo` Put-failure rollback (`:1205`), and provider `Close` (`:1569-1581`, under `e.Lock()`, iterating readers serially). `concurrency.Session.Close()` calls `Orphan()` and then `Revoke` under `context.WithTimeout(s.opts.ctx, ttl)`; since the session is built with `concurrency.NewSession(e.client, WithLease, WithTTL)` at `:1151` with no `WithContext`, `s.opts.ctx` is `client.Ctx()` and the caller's context never participates, while `ttl` is `readerTempInfoSessionTTLSeconds = 60`. clientv3's default `FailFast(false)` (wait-for-ready) means the Revoke does not fail fast when etcd is unreachable. Failure: etcd is down and the application closes readers (or shuts the provider down) with N readers open → each Close burns the full 60s regardless of the ctx passed to `log_reader.go:245`, holds `entry.Lock()` and blocks concurrent updates for that reader, and provider shutdown degrades to N×60s — well past any upper-layer shutdown timeout (before this PR the path was one `Delete` bounded by `requestTimeout`, default 10s). Suggestion: on all three paths use `entry.session.Orphan()` (stops keepalive immediately, no RPC) followed by `e.client.Revoke` under `e.getContextWithTimeout(ctx)`; the explicit `Delete` at `:1485` already removes the key, and an unrevoked lease expires within the TTL. *(raised by zack)*

## Low

**`meta/metadata_etcd.go:1388` — discarded dead session is never `Orphan()`ed, leaking a cancel node on the client's lifetime context**
In the `<-session.Done()` pre-check at `:1387-1393` the session is dropped via `session = nil` and later overwritten by `entry.session = newSession` (`:1406`) without any `Orphan()`/`Close()`. `concurrency.NewSession` creates `ctx, cancel := context.WithCancel(ops.ctx)` with `ops.ctx = client.Ctx()`, and that `cancel` runs only from `Orphan()`/`Close()`. Failure: repeated etcd unavailability exceeding the 60s TTL in a long-lived process drives active readers through this rotation branch repeatedly → one derived `cancelCtx` stranded in the client lifetime context's children map per rotation, accumulating without bound until the etcd client is closed. The `leaseGone` path at `:1437` does call `Close()`, so only this pre-check branch is affected. Suggestion: call `session.Orphan()` before dropping it — not `Close()`, which would pull in the 60s Revoke described above while holding `entry.Lock()`. *(raised by zack)*

- `meta/metadata_etcd.go:1387` — lease-loss recovery is triggered only by updates, so a reader that stops consuming never recovers: the `<-session.Done()` check at `:1387-1393` is the only dead-session detector in the repo, and the sole production caller of `UpdateReaderTempInfo` (`woodpecker/log/log_reader.go:160`) is gated at `:158` on being inside `ReadNext` with either a segment advance or 30s since the last successful read (`UpdateReaderInfoIntervalMs`), while the `ErrSegmentNotFound` branch at `:144-151` `continue`s before reaching it. Failure: a lagging reader stops being consumed (backpressure, paused subscription, or stuck in the `ErrSegmentNotFound` wait) while etcd is unreachable beyond the 60s TTL → the lease expires and the temp key is deleted with no `ReadNext` call to rebuild it and no watchdog on `entry.session.Done()` (`CheckSessionLockAlive` at `:1510-1522` covers writer locks only), so `log_writer.go:524-527` computes a larger `minTruncatedSegmentId` and physically removes segments the reader has not consumed. This is unrelated to this PR because the same outcome occurs more readily on master, where reader keys have no keepalive at all and expire after 60s regardless of etcd health — this PR only narrows the trigger to "etcd unavailable beyond the TTL", and the missing watchdog lives in code this PR does not touch (`log_reader.go:144-164`, `CheckSessionLockAlive`). A watchdog goroutine selecting on `entry.session.Done()` that performs the rotate-and-re-put itself would close it; the payload fields (`openTimestamp`/`openSegmentId`/`openEntryId`) are already held on `entry`. Recommend landing that separate fix on its own. *(raised by zack)*

czs007:

The reader temp-info lease/session rework is sound on the normal path, but provider shutdown and the new 5s reporting cadence introduce lifecycle and throughput hazards worth addressing before merge.

## High

- **`meta/metadata_etcd.go:1269`** — `CreateReaderTempInfo` starts the session keepalive at line 1258 and performs the `Put` at 1269, but only registers the session in `e.readerTempSessions` at line 1290; throughout that window the session is invisible to the registry. `Close()` (1582) simply ranges the registry (1629-1632) without first transitioning the provider to a closed state, and the creation path never takes `e.Lock()`, so the two are not mutually exclusive. Failure: a process reusing one etcd client calls `metadataProviderEtcd.Close()` while another goroutine is inside `OpenLogReader` past the `Put` but before the `Store`, and `Close()`'s `Range` completes first → the session is never retired, and because `Close()` never calls `client.Close()`, its keepalive runs on `e.client.Ctx()` and renews the lease until process exit; the orphaned reader key then pins `minTruncatedSegmentId` in `woodpecker/log/log_writer.go:525-529` (which applies no staleness filter), so truncated segments are never physically cleaned and object storage for that log grows without bound. Suggestion: set a provider-closed flag under `e.Lock()` before `Close()` scans, and make the creation path register atomically under that same flag — if the provider is already closed, immediately `closeReaderTempSession` and fail the create. *(raised by sijie-ni-0214)*

## Medium

- **`meta/metadata_etcd.go:1629`** — `Close()` takes `e.Lock()`/`defer e.Unlock()` for its entire body and then serially ranges reader sessions (1629-1632); each `retireReaderTempSession` → `closeReaderTempSession` (1196) issues a synchronous `Revoke` (1200) under a fresh `getContextWithTimeout`, i.e. a full `etcd.requestTimeout` of 10000ms (`config/woodpecker.yaml:220`). Failure: etcd is partitioned and the process receives SIGTERM with N open reader sessions → shutdown blocks for roughly N × 10s while holding the provider lock (N=60 is ~10 minutes), exceeding any container termination grace period and getting SIGKILLed mid-shutdown — even though these revocations carry no correctness dependency, since `session.Orphan()` (1197) has already stopped the keepalive and the lease expires within one TTL. Suggestion: in the `Range`, only orphan and mark sessions closed (non-blocking), then perform best-effort revocations concurrently under a single shared shutdown deadline — or skip `Revoke` entirely on the provider shutdown path and let the leases expire. *(raised by sijie-ni-0214)*

- **`woodpecker/log/log_reader.go:43`** — `UpdateReaderInfoIntervalMs` drops from 30000 to 5000, a 6× increase in etcd write rate (N readers go from N/30 to N/5 writes per second). The reporting check at `log_reader.go:186` sits ahead of `ReadBatchAdv`, and a caught-up reader loops through `ErrEntryNotFound` → `waitWithContext(200ms)` → `continue`, so the threshold is re-evaluated every ~200ms; the `Put` itself is synchronous on the `ReadNext` call chain under `entry.Lock()` (`metadata_etcd.go:1467-1468`) with a 10s timeout budget. Failure: etcd is degraded such that each `Put` burns the full 10s `requestTimeout` before failing → since 10s > the 5s threshold, the condition at line 186 is true again on the very next iteration, so every batch boundary blocks ~10s and reader throughput collapses to near zero, whereas the old 30s threshold lost only 10s out of every 30s window. The 6× amplification buys only a reduction of the post-lease-loss key-absence window from 30s to 5s, and lease loss itself requires 60s (one TTL) of continuous etcd unavailability. Suggestion: keep 30000, or move reporting off the `ReadNext` path into the reader's own background goroutine with backoff on failure. *(surfaced during verification)*

## Low

- **`meta/metadata_etcd.go:1290`** — The PR description states "Reopening the same reader name swaps in a fresh session and closes the stale one", but no such logic exists: `readerTempSessions` is a `sync.Map` keyed by `*readerTempSession` pointer (`metadata_etcd.go:115`), `CreateReaderTempInfo` unconditionally `Store`s a new object (1290), and there is no lookup by `(logId, readerName)` anywhere (grep for `readerTempSessions` yields only 115/1225/1290/1629). A genuine same-name reopen would leave two independently keepalived sessions, and the unconditional name-keyed `Delete` in `DeleteReaderTempInfo` (1542) would remove the key of a still-live reader. The only production caller, `log_handle.go:725-727`, embeds `time.Now().UnixNano()` in the reader name, so this path is currently unreachable — but the mismatch could mislead a future maintainer into relaxing the naming rule. Suggestion: either implement the name-keyed swap or remove that sentence from the PR body/comments. *(surfaced during verification)*

- **`meta/metadata_etcd_reader_session_test.go:85`** — `testReaderTempInfoOutlivesLeaseTTLWhileIdle` is the only case covering the actual defect from #265 (an idle reader's temp info surviving a full lease TTL), but it is skipped under `testing.Short()` at lines 85-86 because it waits 13.5s. The `unit-test` target in `Makefile:125` runs with `-short`, so CI never executes it and a regression of the exact bug this PR fixes would still go green. The other cases (self-heal, non-resurrection, foreign session) are unaffected and remain covered. Suggestion: reduce the TTL to 2s so total wait lands under ~6s and drop the short-mode skip, or run this case explicitly from an integration test target. *(surfaced during verification)*

tinswzy:
@czs007 Replying to the [review of `54d58cb`](https://github.com/zilliztech/woodpecker/pull/266#issuecomment-5422611865):

All five findings confirmed against the merged code — no false positives. This review landed nine minutes after #266 merged, so the fixes go to a follow-up PR rather than this one. Notes below where verification sharpened or corrected the finding.

- **Provider shutdown races session creation (High)** — confirmed, and the mechanics all check out: `CreateReaderTempInfo` never takes `e.Lock()` (the whole function is lock-free), `Close()` takes it at `metadata_etcd.go:1584` but transitions no state, and there is no `closed` field on `metadataProviderEtcd` at all. `Close()` also never calls `e.client.Close()`, and `newReaderTempSession` deliberately builds the session without `concurrency.WithContext`, so the keepalive runs on `e.client.Ctx()` and renews until process exit. The orphaned key then pins `minTruncatedSegmentId` in `log_writer.go:525-529`, which applies no staleness filter — `RecentReadTimestamp` is written but never read by any production caller.

  One correction that makes this broader than described: the interleaving window is not required. Any reader opened *after* `Close()` returns leaks identically, because the `Range` at `:1629` has already completed and nothing prevents a later `Store`. So the fix has to be a provider-closed state checked on the creation path, not merely moving the `Store` ahead of the `Put`.

- **Serial blocking revocation on shutdown (Medium)** — confirmed. `requestTimeout` is `10000` (`config/woodpecker.yaml:220`, and the in-code fallback at `metadata_etcd.go:121` is also 10000), `Close()` holds `e.Lock()` across its whole body, and the `Range` calls `retireReaderTempSession` → `closeReaderTempSession` → a synchronous `Revoke` one session at a time. Agreed the revocation carries no correctness dependency: `session.Orphan()` at `:1197` has already stopped the keepalive, so the lease expires within one TTL regardless.

  Adjacent issue on the same path, not introduced by this PR: the `logWriterLocks` `Range` immediately above uses `sessionLock.session.Close()` at `:1616`, which revokes under a deadline of one full session TTL — the exact API `closeReaderTempSession` was written to avoid, and its comment says why. That path has the same stall, with a longer per-item budget. Will fix both together.

- **`UpdateReaderInfoIntervalMs` 30000 → 5000 (Medium)** — confirmed, and this one is a regression I introduced in `54d58cb`; the analysis is right. The ordering detail that makes it bite: `l.lastReported = now` is assigned at `log_reader.go:187` *before* the blocking Put at `:189`, so `now` is loop-entry time. After a Put that burns the full 10s timeout plus the 200ms `waitWithContext`, the next iteration sees ~10.2s elapsed — over the 5s threshold, under the old 30s one. So the old value lost 10s out of every 30s; the new one stalls continuously.

  The cost/benefit point stands too: 5s only shortens the post-lease-loss key-absence window, and opening that window requires 60s of continuous etcd unavailability — the same condition under which the 5s cadence does the most damage. I sized the change on steady-state write volume and missed the degraded-path amplification.

- **PR description claims a same-name session swap that does not exist (Low)** — confirmed, and the provenance is a stale description rather than an unimplemented intent. In `c2874da` the registry was `sync.Map` keyed by `readerTempSessionKey(logId, readerName)` with a `Swap` at `:1220`, which is what the description referred to. The follow-up refactor in `7bb283c` re-keyed the map by `*readerTempSession` pointer and dropped the name-keyed lookup along with it; the description was not updated. Agreed the mismatch is worth closing, since the unconditional name-derived `Delete` in `DeleteReaderTempInfo` would remove a live same-name reader's key if the naming rule were ever relaxed. Will add the invariant as a comment on `CreateReaderTempInfo`.

- **Regression test for #265 never runs in CI (Low)** — confirmed, and slightly worse than stated. `unit-test` (`Makefile:125`) runs `go list ./...` minus the `tests/*` trees, so `./meta/` is included and runs with `-short`, skipping the case at `:85-86`. But `integration-test` (`Makefile:130`) runs only `./tests/integration/...`, which does not include `./meta/` — so `testReaderTempInfoOutlivesLeaseTTLWhileIdle` executes in no CI target at all. The other six cases in the file are unaffected. Will drop the short-mode skip and shrink the TTL so the wait fits a unit-test budget.

Follow-up PR will cover the shutdown lifecycle (findings 1 and 2, plus the writer-lock path), the interval regression, the naming invariant comment, and the test coverage gap, each with a regression test verified failing first.

PR #268, opened 2026-08-20
enhance(monitor): scope dashboards by a woodpecker_id instance label

### LINKED ISSUE #267: enhance(monitor): add a woodpecker_id instance label so one namespace can host multiple woodpecker clusters
## Motivation

The monitor dashboards assume a namespace holds exactly one woodpecker cluster. In a pooled deployment that no longer holds: a single namespace can host several woodpecker StatefulSets (`wp-<instance-id>-N`), all matched by one PodMonitor. None of the dimensions the dashboards expose can tell those instances apart:

- `namespace` — shared by every instance in the pool;
- `server_job` — one value for the whole pool, because prometheus-operator derives `job` from the PodMonitor (`<namespace>/<podmonitor-name>`), not from the scraped pod;
- `log_ns` / `log_id` — per business log, orthogonal to which cluster serves it.

That leaves the pod name as the only discriminator, so an operator has to hand-type a pod-name regex (`wp-<id>-.*`) into a free-text variable. It is also silently wrong when one Prometheus aggregates several k8s clusters that each have the namespace: same-named pods from two clusters are summed with no indication.

The scrape side can supply a proper identity for free — every woodpecker pod already carries `app.kubernetes.io/instance`.

## Proposed changes

- **PodMonitors** (`tests/k8s/monitor/manifests/podmonitor-{server,client}.yaml`): relabel `__meta_kubernetes_pod_label_app_kubernetes_io_instance` into a `woodpecker_id` target label, stripping the `wp-` prefix when present.
- **docker monitor** (`tests/docker/monitor/prometheus.yml`): stamp the same `woodpecker_id` label on the static targets, so both suites share one label scheme.
- **Dashboards** (all four templates under `tests/k8s/monitor/grafana/templates/` and `tests/docker/monitor/grafana/templates/`): add a multi-select `woodpecker_id` variable and thread `woodpecker_id=~"$woodpecker_id"` through every panel expression and variable query. `allValue` must be `.*` rather than the generated value list, so series scraped by a deployment that predates the relabel — where the label is absent — still match under `All` (a missing label equals the empty string, which `.*` matches).
- **Regenerate** `tests/k8s/monitor/manifests/dashboard-configmaps.yaml` from the templates and add a drift check. The generated file is currently stale: it lacks the `Disk Watermark — Backpressure` and `Compacted Cleanup & Decommission` rows of the server dashboard and the `Progress Frontiers` / `Segment Frontiers` / `Entry Frontiers` / `Compaction Failure Rate by Reason` panels of the client one, so the k8s suite applies an outdated dashboard.
- **Tests** (`tests/k8s/monitor/dashboards_test.go`): require the `woodpecker_id` template variable, and assert the configmap content still matches the templates so the two cannot drift again.

## Non-goals

- No metric or label changes inside the woodpecker binaries — this is scrape configuration plus dashboards only.
- Per-replica / cross-AZ traffic visibility (which replica served a read, how much traffic crossed an availability zone) is a separate gap; it needs new client metrics and is tracked in #269.

tinswzy:
Fixes #267

## What

A namespace can host several woodpecker clusters — a pooled deployment runs one StatefulSet per instance, named `wp-<id>`. prometheus-operator derives `job` from the PodMonitor, so every instance in the pool reports the same `job` value, and none of the dashboard dimensions could separate them (`namespace` is shared, `log_ns`/`log_id` are per business log). The only remaining discriminator was a hand-typed pod-name regex, which is also silently wrong when one Prometheus aggregates several k8s clusters that each have the namespace.

This adds a proper instance identity to the scrape config and threads it through the dashboards.

## Changes

- **PodMonitors** — relabel `__meta_kubernetes_pod_label_app_kubernetes_io_instance` into a `woodpecker_id` target label, stripping the `wp-` prefix (`(?:wp-)?(.+)`), so both the test suite's `my-woodpecker` and production's `wp-<id>` produce a clean value.
- **docker suite** — stamp the same `woodpecker_id` on the static targets, so one dashboard layout serves both suites.
- **Dashboards** (4 templates) — new multi-select `woodpecker_id` variable, and `woodpecker_id=~"$woodpecker_id"` added to every panel expression and variable query. Each dashboard discovers the values from the metric its existing variables already use (`go_info` for k8s, `woodpecker_server_logstore_instances_total` / `woodpecker_client_append_requests_total` for docker).
- **`allValue` is `.*`**, not the generated value list: a missing label equals the empty string in PromQL, so `All` keeps matching targets scraped by a deployment that predates the relabel. With the default value list they would vanish from every panel.
- **`dashboard-configmaps.yaml`** — regenerated, and a new test keeps it in sync. It had drifted: the committed manifest was missing the `Disk Watermark — Backpressure` and `Compacted Cleanup & Decommission` rows of the server dashboard and the `Progress Frontiers` / `Segment Frontiers` / `Entry Frontiers` / `Compaction Failure Rate by Reason` panels of the client one, so `run_monitor_tests.sh` was applying an outdated dashboard.
- **Tests** — `dashboards_test.go` requires the `woodpecker_id` variable and asserts the ConfigMap matches the templates (`-update` regenerates); the live `LabelScheme` subtest asserts the relabeled value equals `$CR_NAME`.

## Verification

- `go test ./tests/k8s/monitor/` — pass; the sync test was confirmed to fail against the stale manifest before regenerating, so it does catch drift.
- `golangci-lint run --config .golangci.yml ./tests/k8s/monitor/...` — 0 issues; `gofmt`/`gofumpt` clean.
- All 224 distinct PromQL expressions from the four patched templates were executed against a live Prometheus with the variables substituted — 0 parse or evaluation errors. This caught one real bug during development: the first pass injected the filter into `by (grpc_method)` grouping clauses, since a grouping label looks exactly like a bare metric selector.
- YAML of both PodMonitors, `prometheus.yml` and the regenerated ConfigMaps re-parsed; the embedded dashboards compare equal to the templates.

## Notes

- No metric or label changes inside the woodpecker binaries — scrape configuration plus dashboards only.
- The test suite deploys a single release, so `woodpecker_id` has one value there; the multi-instance case it targets is the pooled production deployment. Widening the suite to two releases would exercise the split, but that is a bigger change to `run_monitor_tests.sh` and is left out.
- Per-replica / cross-AZ traffic visibility (which replica served a read, how much traffic crossed an AZ) is a separate gap that needs new client metrics; it is tracked in #269.

### DIFF
--- a/tests/docker/monitor/grafana/templates/dashboard_client.json
+++ b/tests/docker/monitor/grafana/templates/dashboard_client.json
@@ -55,7 +55,7 @@
       },
       "targets": [
         {
-          "expr": "sum(rate(woodpecker_client_append_requests_total{log_ns=~\"$log_ns\"}[1m])) by (log_id)",
+          "expr": "sum(rate(woodpecker_client_append_requests_total{woodpecker_id=~\"$woodpecker_id\", log_ns=~\"$log_ns\"}[1m])) by (log_id)",
           "legendFormat": "logID={{log_id}}",
           "refId": "A"
         }
@@ -106,12 +106,12 @@
       },
       "targets": [
         {
-          "expr": "histogram_quantile(0.5, sum(rate(woodpecker_client_append_latency_bucket{log_ns=~\"$log_ns\"}[1m])) by (log_id, le))",
+          "expr": "histogram_quantile(0.5, sum(rate(woodpecker_client_append_latency_bucket{woodpecker_id=~\"$woodpecker_id\", log_ns=~\"$log_ns\"}[1m])) by (log_id, le))",
           "legendFormat": "logID={{log_id}} Append latency P50",
           "refId": "A"
         },
         {
-          "expr": "histogram_quantile(0.99, sum(rate(woodpecker_client_append_latency_bucket{log_ns=~\"$log_ns\"}[1m])) by (log_id, le))",
+          "expr": "histogram_quantile(0.99, sum(rate(woodpecker_client_append_latency_bucket{woodpecker_id=~\"$woodpecker_id\", log_ns=~\"$log_ns\"}[1m])) by (log_id, le))",
           "legendFormat": "logID={{log_id}} Append latency P99",
           "refId": "B"
         }
@@ -162,12 +162,12 @@
       },
       "targets": [
         {
-          "expr": "histogram_quantile(0.5, sum(rate(woodpecker_client_append_bytes_bucket{log_ns=~\"$log_ns\"}[1m])) by (log_id, le))",
+          "expr": "histogram_quantile(0.5, sum(rate(woodpecker_client_append_bytes_bucket{woodpecker_id=~\"$woodpecker_id\", log_ns=~\"$log_ns\"}[1m])) by (log_id, le))",
           "legendFormat": "logID={{log_id}} Append bytes P50",
           "refId": "A"
         },
         {
-          "expr": "histogram_quantile(0.99, sum(rate(woodpecker_client_append_bytes_bucket{log_ns=~\"$log_ns\"}[1m])) by (log_id, le))",
+          "expr": "histogram_quantile(0.99, sum(rate(woodpecker_client_append_bytes_bucket{woodpecker_id=~\"$woodpecker_id\", log_ns=~\"$log_ns\"}[1m])) by (log_id, le))",
           "legendFormat": "logID={{log_id}} Append bytes P99",
           "refId": "B"
         }
@@ -218,7 +218,7 @@
       },
       "targets": [
         {
-          "expr": "sum(rate(woodpecker_client_read_requests_total{log_ns=~\"$log_ns\"}[1m])) by (log_id)",
+          "expr": "sum(rate(woodpecker_client_read_requests_total{woodpecker_id=~\"$woodpecker_id\", log_ns=~\"$log_ns\"}[1m])) by (log_id)",
           "legendFormat": "logID={{log_id}}",
           "refId": "A"
         }
@@ -269,12 +269,12 @@
       },
       "targets": [
         {
-          "expr": "histogram_quantile(0.5, sum(rate(woodpecker_client_read_latency_bucket{log_ns=~\"$log_ns\"}[1m])) by (log_id, le))",
+          "expr": "histogram_quantile(0.5, sum(rate(woodpecker_client_read_latency_bucket{woodpecker_id=~\"$woodpecker_id\", log_ns=~\"$log_ns\"}[1m])) by (log_id, le))",
           "legendFormat": "logID={{log_id}} Read latency P50",
           "refId": "A"
         },
         {
-          "expr": "histogram_quantile(0.99, sum(rate(woodpecker_client_read_latency_bucket{log_ns=~\"$log_ns\"}[1m])) by (log_id, le))",
+          "expr": "histogram_quantile(0.99, sum(rate(woodpecker_client_read_latency_bucket{woodpecker_id=~\"$woodpecker_id\", log_ns=~\"$log_ns\"}[1m])) by (log_id, le))",
           "legendFormat": "logID={{log_id}} Read latency P99",
           "refId": "B"
         }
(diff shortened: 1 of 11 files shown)


czs007:

This PR's dashboard filtering work is structurally sound (all 220+ patched expressions carry the `woodpecker_id` filter inside a metric selector and the generated ConfigMap matches the templates byte-for-byte); the findings below concern the new drift guard not being enforced and two label-plumbing gaps.

## Medium

**`.github/workflows/unit-test.yaml:46` — the new ConfigMap sync test never runs in CI.**
The unit-test job filters out every package under `tests/` (`go list ./... | grep -v -E '(tests/)'`), so `TestK8sDashboardConfigMaps_InSync` (`tests/k8s/monitor/dashboards_test.go:105`) — the "keeps it in sync" guarantee cited in the PR description — is never executed by PR checks. All 16 workflows were read individually; only `unit-test.yaml` invokes `go test`, and none reference `tests/k8s/monitor`. Only a local `make unit-test` reaches it, since the Makefile excludes just `tests/integration|tests/benchmark|tests/stability|tests/docker` (`Makefile:126`). Failure: a contributor edits `grafana/templates/dashboard_server_k8s.json` and forgets `go test ./tests/k8s/monitor -run TestK8sDashboardConfigMaps_InSync -update` → all checks go green, `manifests/dashboard-configmaps.yaml` stays stale, and `run_monitor_tests.sh:78` applies an outdated dashboard, reproducing exactly the drift this PR just repaired. Suggestion: align the CI filter with the Makefile's (`grep -v -E '(tests/integration|tests/benchmark|tests/stability|tests/docker)'`), or add an explicit `go test ./tests/k8s/monitor/...` step. Note the workflow's `paths` filter is `**.go`, so a JSON-only template change still will not trigger this job — the filter change alone is not sufficient. (provenance unclear — please confirm whether this PR introduces it) (raised by czs007)

**`tests/k8s/monitor/dashboards_test.go:57` — nothing asserts that panel expressions actually carry the `woodpecker_id` filter.**
`TestK8sDashboards_Valid` only checks that a template *variable* named `woodpecker_id` exists, while the substance of the PR is threading `woodpecker_id=~"$woodpecker_id"` into every panel expression and variable query. The current state is correct — all four templates were checked mechanically (83 expressions per server template, 27 per client template, all filtered; only the upstream `label_values(go_info, namespace)` query legitimately omits it) — so this is a missing regression guard, not a live defect. Two gaps: the expression-level invariant has no automated assertion at all (it was validated once by hand against a live Prometheus, which is how the "filter injected into the `by (grpc_method)` grouping clause" bug was caught during development), and the two templates under `tests/docker/monitor/grafana/templates/` have no test coverage whatsoever, since the file list at `dashboards_test.go:31-34` contains only the two `_k8s.json` files. Suggestion: walk `panels[*]` (including nested panels inside rows) and their `targets[*].expr`, require `woodpecker_id` in each and require it to appear only inside a `{...}` selector — the same invariant that caught the injection bug — and include the two docker templates in the same loop. (raised by czs007)

## Low

**`tests/k8s/monitor/grafana/templates/dashboard_client_k8s.json:487` — two client table panels gain an unhidden `woodpecker_id` column.**
`Log Name → Log ID Mapping` (id 14) and `Active Segment → Nodes (by Log)` (id 16) are `format: table` instant queries, so the `organize` transform renders every series label as a column unless listed in `excludeByName`. Both exclude maps currently list `Time`/`__name__`/`instance`/`job`/`namespace` (plus `Value` in the second) but not `woodpecker_id`, which this PR now stamps on every one of these series. Failure: deploy this PR's dashboards and open either panel → the table renders a raw, un-renamed `woodpecker_id` column beside the business columns, taking column width and contradicting the deliberate hiding of `instance`/`job`/`namespace` in the same table. Suggestion: add `"woodpecker_id": true` to all four exclude maps — `dashboard_client_k8s.json:487` and `:596`, plus `tests/docker/monitor/grafana/templates/dashboard_client.json:487` and `:596` (line numbers verified identical across the two templates) — then re-run with `-update` to regenerate `manifests/dashboard-configmaps.yaml`. (raised by czs007)

**`tests/k8s/monitor/monitor_test.go:77` — the `wp-` prefix stripping has no effective test coverage.**
The `LabelScheme` subtest reads `CR_NAME` and applies `strings.TrimPrefix(wantID, "wp-")` (`monitor_test.go:77-81`), which reads as if the suite were parameterized by `CR_NAME`. But both PodMonitors hardcode `app.kubernetes.io/instance: my-woodpecker` in their selectors (`podmonitor-server.yaml:16`, `podmonitor-client.yaml:16`) and the loadgen pod label is likewise hardcoded (`client-loadgen.yaml:9`), while the operator stamps the real pod label from `$CR_NAME` (`deployments/operator/test/lib.sh:221`). Failure: run the documented parameterized form `CR_NAME=wp-1 ./tests/k8s/monitor/run_monitor_tests.sh` → both PodMonitors select zero pods, Prometheus has no woodpecker targets, and the suite dies at `monitor_test.go:73` with `no logstore_active_logs series` before the new `woodpecker_id` assertion or the stripping branch is ever reached. So in any run that gets far enough, `TrimPrefix` is a no-op and `(?:wp-)?(.+)` is only ever exercised on its non-stripping branch; the production `wp-<id>` shape is never validated. The regex semantics were checked by hand and are correct (Prometheus fully anchors, Go leftmost-first: `wp-1` → `1`, `my-woodpecker` unchanged), so there is no known wrong value — only absent coverage. Suggestion: template the instance name in the three manifests off `$CR_NAME` (they are already applied by a script that exports it), or drop the `CR_NAME` lookup and assert the constant, so the test does not imply coverage it does not have. (raised by czs007)

PR #270, opened 2026-08-20
enhance(metrics): make per-replica and cross-AZ client traffic observable

### LINKED ISSUE #269: enhance(metrics): no per-replica or cross-AZ traffic visibility on the client
## Motivation

The client has no visibility into *which replica* served a request or *how far the traffic travelled*. Every client metric in `common/metrics/client_metrics.go` is aggregated by `log_ns` / `log_id` (plus `operation` / `status` / `reason`), so none of these questions can be answered today:

- when an entry is written to a Wq quorum, how did each individual replica do — success, latency, bytes?
- when a batch is read from an active segment, did it come from a replica in the local availability zone, or did it cross an AZ (or a region)?
- how often does a local replica fail and force a fallback to a remote one?

The only metric carrying a node dimension is `WpActiveSegmentNode` (`client_metrics.go:209`), a membership gauge for the current writable segment: labels `log_ns / log_id / segment_id / node`, value always 1. It answers "which three endpoints form the quorum", never "how much traffic went where". Cross-AZ traffic is a real cost and latency factor in a multi-AZ deployment, and right now it is entirely unobservable from the metrics.

The foundations are already in place — this is instrumentation, not new plumbing:

- `proto.QuorumNode` (`proto/meta.pb.go:562`) carries `Endpoint / NodeId / ClusterName / Region / Az / ResourceGroup` for every replica, persisted in the quorum metadata.
- The client knows its own placement via `topology.GetCurrentRegion()` / `GetCurrentAvailabilityZone()` (`common/topology/topology_utils.go`, env `REGION` / `AVAILABILITY_ZONE`).
- The read path **already uses this**: `orderedQuorumReadCandidates` (`woodpecker/segment/segment_handle.go:599`) sorts replicas whose `Region`/`Az` match the client's ahead of the rest. Whether the node that actually served a read was local is known at the moment of success in `quorumReadBatch` (`segment_handle.go:654`) — it is simply never recorded.
- The write path fans out per node in `append_op.go` (`nodeCount := len(op.quorumInfo.Nodes)` at :135, `serverAddr` at :158) and already keeps per-replica `channelAttempts` / `channelErrors`; the existing latency/bytes observations at :257 and :294 are quorum-level only.

## Proposed changes

Add an `az_scope` dimension — a normalized placement of the *peer* relative to the client — with a small, fixed set of values: `local`, `cross_az`, `cross_region`, `unknown` (client placement not configured). Deliberately **not** a `node` / `endpoint` label: that multiplies by `log_id` and blows up cardinality, while `az_scope` stays at four values.

New client metrics:

```
woodpecker_client_replica_append_total{log_ns, log_id, az_scope, status}
woodpecker_client_replica_append_bytes_total{log_ns, log_id, az_scope}
woodpecker_client_quorum_read_total{log_ns, log_id, az_scope, status}
woodpecker_client_quorum_read_bytes_total{log_ns, log_id, az_scope}
woodpecker_client_read_failover_total{log_ns, log_id, from_scope, to_scope}
```

Instrumentation points:

- **write** — the per-node result path in `append_op.go`, so each replica of a quorum append is counted once with its own scope and status.
- **read** — the candidate loop in `quorumReadBatch`; count the scope of the candidate that succeeded, and count a failover when an earlier candidate of a different scope failed first.
- **scope resolution** — `orderedQuorumReadCandidates` already builds a `replicasByEndpoint` map; carrying the resolved scope on `quorumReadCandidate` (`segment_handle.go:594`) avoids a second lookup at record time.
- Direct reads from object storage (`WpClientDirectReadRequestsTotal`, `client_metrics.go:217`) are not replica traffic; keep them as their own series rather than folding them into a scope.

Also worth doing, cheaply: add an `az` label to `WpActiveSegmentNode`. It is a membership gauge (series count = active logs × quorum size), so cardinality is bounded, and it turns the existing "Active Segment → Nodes" panel into an AZ-distribution view of the current write quorum.

Dashboards: a new `Cross-AZ / Replica Traffic` row in both client dashboards (`tests/k8s/monitor/grafana/templates/dashboard_client_k8s.json`, `tests/docker/monitor/grafana/templates/dashboard_client.json`) — write traffic and success rate by scope, read traffic by scope, a cross-AZ read ratio stat, and the failover rate.

## Prerequisites and validation

Two things that must be handled, or the feature ships blind:

1. **The client must have `AVAILABILITY_ZONE` / `REGION` in its environment.** `orderedQuorumReadCandidates` short-circuits when either is empty (`segment_handle.go:626`) — so on a client without them the local-preference logic is *already* inert today, and the new metrics would report `unknown` for everything. Whatever deployment embeds the client needs to inject these the same way the server side does.
2. **A single-AZ deployment never exercises the cross-* paths.** The k8s monitor suite should inject a different `AVAILABILITY_ZONE` per server pod to fake a multi-AZ topology; that makes all four scopes reachable on a single-node minikube and promotes the existing `TestReadBatchAdv_PrefersLocalRegionAZReplica` unit test into an end-to-end check.

## Non-goals

- No server-side metrics; the placement question is a client-side routing question.
- No `node` / `endpoint` label on any counter — see the cardinality note above.
- Not changing the read-preference policy itself, only observing it.

tinswzy:
Fixes #269

## What

Client metrics were all aggregated by log, so nothing could say *which replica* served a request or *how far the traffic travelled*. This adds an `az_scope` dimension — where the peer sits relative to the client — to the write and read paths, and surfaces it on both client dashboards.

The placement data was already available: `proto.QuorumNode` carries each replica's region/AZ, and `orderedQuorumReadCandidates` already uses it to prefer a local replica. This PR records what that machinery was already deciding.

## New metrics

| Metric | Labels |
|---|---|
| `woodpecker_client_replica_append_total` | log_ns, log_id, az_scope, status |
| `woodpecker_client_replica_append_bytes_total` | log_ns, log_id, az_scope |
| `woodpecker_client_quorum_read_total` | log_ns, log_id, az_scope, status |
| `woodpecker_client_quorum_read_bytes_total` | log_ns, log_id, az_scope |
| `woodpecker_client_read_failover_total` | log_ns, log_id, from_scope, to_scope |

`az_scope` ∈ `local` / `cross_az` / `cross_region` / `unknown`. The replica endpoint is deliberately **not** a label — it would multiply by `log_id`, while `az_scope` stays at four values and answers the question that costs money.

`woodpecker_client_active_segment_node` gains an `az` label, so the AZ spread of the current write quorum is visible from the membership gauge. Its clear path now matches on the segment rather than the exact label set, so a quorum whose placement changed since it was published is still removed in full (covered by a new test).

## Instrumentation points

- **Write** — every point where a replica resolves: the single-append ack callback, the group-commit drain (`applyNodeAck`), and the send-failure path (`failNode`, client-unavailable). `FastFail`-completed ops are skipped, so a fast-failed entry is not double counted.
- **Read** — the candidate loop in `quorumReadBatch`. `ErrEntryNotFound` and end-of-file are **not** counted: a caught-up tail reader hits `ErrEntryNotFound` on every poll, so counting it would swamp the error rate and add per-poll counter churn for traffic that never moved. Logging behaviour is unchanged.
- **Scope resolution** — a segment's quorum is fixed, so scopes are resolved once per segment handle and passed to each `AppendOp`, rather than rebuilding an endpoint map per append.

`topology.Scope` returns `unknown` unless *both* sides know their placement, which keeps the existing read-preference semantics exactly: a client without `REGION` / `AVAILABILITY_ZONE` sees no local replica and keeps the original candidate order.

## Dashboards

A `Cross-AZ / Replica Traffic` row on both client dashboards (k8s + docker): replica write bytes by scope, replica write rate by scope/status, quorum read bytes by scope, cross-zone traffic ratio (read and write), read failover rate, and active segment replicas by AZ. `dashboard-configmaps.yaml` regenerated via the sync test added in #268.

## Verification

- `go test ./woodpecker/segment/ ./common/metrics/ ./common/topology/ ./tests/k8s/monitor/` — all pass (segment suite 159s).
- New tests: `topology.Scope` truth table; `quorumNodeScopes` / `activeSegmentNodes` incl. a replica missing from the quorum's replica list; candidates carrying their scope; a read that fails over from local to a remote replica asserting the exact counter values (error on local, success + bytes on cross_region, one failover local→cross_region); and a local read asserting **no** failover is recorded.
- `golangci-lint run --config .golangci.yml` on the touched packages — 0 issues; `gofmt` clean.
- All 14 new dashboard expressions executed against a live Prometheus with variables substituted — 0 parse/eval errors.

## Notes / not done

- **No live multi-AZ validation.** Verification is unit-level. The prerequisite from #269 — faking a per-pod `AVAILABILITY_ZONE` in the k8s monitor suite so the `cross_*` branches are exercised end to end — is **not** in this PR: a StatefulSet shares one pod template, so per-pod placement needs an init container or downward-API wiring, which is its own change.
- Whatever embeds the client still has to inject `REGION` / `AVAILABILITY_ZONE`; without them every sample is `unknown` — and note the read path's local-replica preference is already inert in that case today. This is documented alongside the new metrics table in `tests/docker/monitor/README.md`.

### DIFF
--- a/common/metrics/active_segment_node_test.go
+++ b/common/metrics/active_segment_node_test.go
@@ -31,7 +31,11 @@ func TestActiveSegmentNodes_SetAndClear(t *testing.T) {
 	reg.MustRegister(WpActiveSegmentNode)
 
 	ns, logId, segId := "bucket/root", "42", "7"
-	nodes := []string{"10.0.0.1:8000", "10.0.0.2:8000", "10.0.0.3:8000"}
+	nodes := []ActiveSegmentNode{
+		{Node: "10.0.0.1:8000", AZ: "us-west-2a"},
+		{Node: "10.0.0.2:8000", AZ: "us-west-2b"},
+		{Node: "10.0.0.3:8000", AZ: ""}, // placement unknown for this replica
+	}
 
 	SetActiveSegmentNodes(ns, logId, segId, nodes)
 
@@ -39,12 +43,34 @@ func TestActiveSegmentNodes_SetAndClear(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, len(nodes), count, "one series per node after set")
 	for _, n := range nodes {
-		v := testutil.ToFloat64(WpActiveSegmentNode.WithLabelValues(ns, logId, segId, n))
+		v := testutil.ToFloat64(WpActiveSegmentNode.WithLabelValues(ns, logId, segId, n.Node, n.AZ))
 		assert.Equal(t, float64(1), v)
 	}
 
-	ClearActiveSegmentNodes(ns, logId, segId, nodes)
+	ClearActiveSegmentNodes(ns, logId, segId)
 	count, err = testutil.GatherAndCount(reg, "woodpecker_client_active_segment_node")
 	assert.NoError(t, err)
 	assert.Equal(t, 0, count, "series removed after clear")
+}
+
+// A quorum whose placement changed since it was published must still be
+// removed in full: the clear matches on the segment, not on the exact labels.
+func TestActiveSegmentNodes_ClearAfterPlacementChange(t *testing.T) {
+	WpActiveSegmentNode.Reset()
+
+	reg := prometheus.NewRegistry()
+	reg.MustRegister(WpActiveSegmentNode)
+
+	ns, logId, segId := "bucket/root", "42", "7"
+	SetActiveSegmentNodes(ns, logId, segId, []ActiveSegmentNode{{Node: "10.0.0.1:8000", AZ: "us-west-2a"}})
+	SetActiveSegmentNodes(ns, logId, segId, []ActiveSegmentNode{{Node: "10.0.0.1:8000", AZ: "us-west-2c"}})
+
+	count, err := testutil.GatherAndCount(reg, "woodpecker_client_active_segment_node")
+	assert.NoError(t, err)
+	assert.Equal(t, 2, count)
+
+	ClearActiveSegmentNodes(ns, logId, segId)
+	count, err = testutil.GatherAndCount(reg, "woodpecker_client_active_segment_node")
+	assert.NoError(t, err)
+	assert.Equal(t, 0, count, "all series for the segment removed")
 }
--- a/common/metrics/client_metrics.go
+++ b/common/metrics/client_metrics.go
@@ -229,7 +229,7 @@ var (
 		Subsystem: clientRole,
 		Name:      "active_segment_node",
 		Help:      "Quorum node membership of each log's current active (writable) segment (value always 1, one series per node)",
-	}, []string{"log_ns", "log_id", "segment_id", "node"})
+	}, []string{"log_ns", "log_id", "segment_id", "node", "az"})
 
 	// Direct read metrics (client reads sealed segments directly from object storage)
 	WpClientDirectReadRequestsTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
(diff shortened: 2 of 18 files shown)


czs007:

Rendering the two surviving findings (one medium, one low) into a single English PR review comment.

Adversarial review surfaced one dashboard-semantics issue (the new Cross-Zone Traffic Ratio panel counts `unknown` as cross-zone) and one minor failover-metric accounting nuance in the read path; the core write/read instrumentation, double-count paths, and clear logic were traced and hold.

## Medium

- **`tests/docker/monitor/grafana/templates/dashboard_client.json:1054`** — The Cross-Zone Traffic Ratio panel (id 109) uses `az_scope!="local"` as the numerator for both the read and write expressions, which counts `az_scope="unknown"` as cross-zone traffic. `topology.Scope()` returns `ScopeUnknown` whenever `REGION`/`AVAILABILITY_ZONE` is unset on either side, and the README states that without them every sample is `unknown`. Client without `REGION`/`AVAILABILITY_ZONE` (the default deployment described in the PR README), or replicas with empty Region/Az in quorum metadata → numerator equals denominator and the panel shows a flat 100% cross-zone ratio for both curves, misread as all traffic crossing zones when placement is simply unconfigured (and the ratio is overstated in mixed deployments where some replicas are `unknown`). The same expression appears in all three copies: `tests/docker/monitor/grafana/templates/dashboard_client.json:1054,1059`, `tests/k8s/monitor/grafana/templates/dashboard_client_k8s.json:1054,1059`, `tests/k8s/monitor/manifests/dashboard-configmaps.yaml:4523,4528`. Suggestion: change the numerator to `az_scope=~"cross_az|cross_region"` (or explicitly document that `unknown` is included in the panel description), apply to all three files, and regenerate the configmap. (raised by sijie-ni-0214, xiaocai2333)

## Low

- **`woodpecker/segment/segment_handle.go:737`** — `quorumReadBatch` sets `preferredScope = candidates[0].azScope`, but `orderedQuorumReadCandidates` only promotes `ScopeLocal` candidates (`:706`); when no local replica exists, `candidates[0]` is just the round-robin head starting from `lastReadState.Node` (`:675-691`), whose scope may be `cross_az`, `cross_region`, or `unknown`. Client with `REGION`/`AVAILABILITY_ZONE` set, segment replicas all in other AZs/regions, round-robin head fails and a replica of a different scope serves the read → `woodpecker_client_read_failover_total` records `from_scope=cross_az,to_scope=cross_region` (or the reverse) even though no local preference was ever expressed, and `from_scope` jitters as the round-robin start rotates, making the Read Failover panel show a "local-preference failure" that does not exist. Only affects the failover counter and its panel; bytes/total counters and the all-`unknown` default case (where `from == to` is not counted) are unaffected. Suggestion: record failover only when `candidates[0].azScope == ScopeLocal`, or fix `preferredScope` to `unknown` when no local replica is present. (raised by xiaocai2333)

PR #272, opened 2026-08-21
fix(membership): announce load over gossip so peers hold a fresh reading

### LINKED ISSUE #271: [bug] Load-aware quorum selection is blind at scale: a node's load reaches only ~3-4 peers regardless of cluster size
## Summary

Load-aware quorum selection (#114) silently degrades to near-uniform random as the cluster grows, because a node's load factor **never reaches most of its peers**. A node's freshly sampled load is written only to its own memory; the only way a peer learns it is a *direct* pairwise push/pull, which memberlist performs with **one** randomly chosen node every 15s.

Measured on a real cluster: a node holds a fresh (within `loadTTL`) load reading for **~3–4 peers, roughly regardless of cluster size**.

| Cluster size | Peers with fresh load | Fraction of cluster |
|---|---|---|
| 5 nodes | 2.4 – 2.8 of 4 | 60 – 70 % |
| 10 nodes | 3.2 of 9 | 36 % |
| 16 nodes | 3.8 of 15 | **25 %** |

Every peer whose reading has expired is scored at `unknownLoad = 0.5` (`service_discovery.go:996`), so at 16+ nodes three quarters of the weight vector is a constant and `selectLowestLoadNodes` is mostly picking at random. The feature works as designed only on very small clusters.

## Root cause

Three separate facts combine. Each is individually defensible; together they close every propagation path except the slowest one.

| # | Fact | Location |
|---|---|---|
| 1 | The load reporter deliberately does not call `memberlist.UpdateNode()`, so a load change never produces an `alive` broadcast | `common/membership/server_node.go:233` |
| 2 | `LocalState()` marshals **only the local node's own meta** — a node never forwards what it knows about others, so load has no transitivity | `common/membership/server_delegate.go:68` |
| 3 | memberlist's node-table `Meta` is populated only by `setAlive()` and `UpdateNode()`, so it stays frozen at its startup value (`load=0`) | memberlist `memberlist.go:457,516` |

`reportLoadOnce` itself runs correctly and on schedule — every `loadReportInterval` (default 10s), from startup until shutdown (`server_node.go:215`). But both of its steps are purely local:

```go
func (n *ServerNode) reportLoadOnce() {
	n.publishLoad()                                    // writes n.delegate.meta in memory
	snap := n.delegate.SnapshotMeta()
	n.discovery.UpdateServer(snap.GetNodeId(), snap)   // updates its own discovery entry
}
```

Neither sends a packet. The value sits in memory until a peer happens to come and fetch it.

That leaves exactly one live path: the user-state half of push/pull (`LocalState` → `MergeRemoteState`). memberlist's `pushPull()` selects `kRandomNodes(1, ...)` (memberlist `state.go:651`) every `PushPullInterval`, which `DefaultLocalConfig()` sets to **15s** and `server_node.go:119` does not override. For a specific ordered pair the expected wait is therefore `15s × (N-1)/2` — 67s at 10 nodes, 220s at 30 — against a `loadTTL` of 30s.

The comment at `server_node.go:237` states the intent behind fact 1:

> We deliberately do NOT call memberlist.UpdateNode here: load is a best-effort hint, not worth churning memberlist node state / bumping incarnations on our own cadence (which also raced with status readers). memberlist already gossips this meta on its own push/pull cadence via LocalState, and peers ingest it in MergeRemoteState.

The first half is a reasonable tradeoff. The second half is where it goes wrong: push/pull's *user state* carries only the sender's own meta, so it is a pairwise channel, not a gossip channel — it does not scale the way the comment assumes.

## Evidence

Reproduced with a temporary probe (3 real `ServerNode`s, loopback advertise addresses, `loadTTL` raised to 300s to isolate propagation from expiry). Node A's load is changed from `0.11` to `0.99` at t=0 and both of B's views are polled:

| t | B's memberlist node-table view of A | B's `discovery` view of A |
|---|---|---|
| 0s | 0.00 | 0.11 |
| 30s | 0.00 | 0.11 |
| **50s** | 0.00 | **0.99** |
| 90s | **0.00** | 0.99 |

The node-table `Meta` never moves — it still holds the startup value after 90s. The `discovery` entry updates once a direct push/pull happens to land.

Control, same run: calling `a.UpdateMeta(...)` (which does call `UpdateNode` → `alive` broadcast) makes B's node-table view show `0.99` **within 1 second**. The broadcast path works and is fast; the load reporter simply never uses it.

The freshness table at the top of this issue comes from a second probe: 5 / 10 / 16 real nodes, `loadReportInterval=2s`, `loadTTL=30s`, sampled every 6s for 60s after membership converged. The absolute count of fresh peers (2.4 → 3.2 → 3.8) is consistent with each node refreshing ~2 peers per 15s round and ~2 rounds fitting inside the 30s TTL — i.e. a constant, not a fraction of the cluster.

## Impact

- On clusters of ~10 nodes and up, `loadAwareEnabled: true` buys close to nothing; selection is effectively `SelectRandom`.
- `WpQuorumSelectionSkew` reports mode `weighted` (not `random_no_load`) the whole time, because the selecting node's *own* load is always fresh — so `anyKnown` is true and the metric looks healthy while 75 % of the weight vector is the `0.5` placeholder. The existing observability cannot detect this.
- Newly added capacity is not preferentially used, which is the main thing #114 set out to achieve.

## Why this should not be fixed in isolation

Simulation (not measurement) of the corrected system suggests that restoring load visibility *without* damping the selector introduces classic herd oscillation: with writers re-selecting at segment rollover and a propagation lag still on the order of the rollover interval, mean imbalance across nodes moves from ~1.06x (today's blind behaviour) to ~1.83x, peaking at ~2.29x — every writer chases the same idle nodes, overloads them, then flees together. Today's blindness is accidentally protecting the cluster.

Shortening the propagation lag well below the rollover interval removes the oscillation in the same simulation (~1.07x), so the fix direction is: make propagation fast **and** make the selector robust to correlated views. Suggested order:

1. Damp the selector so that many clients holding an identical load snapshot do not converge on the same nodes.
2. Then close the propagation gap described here.

## Suggested fix for this issue

Have `reportLoadOnce` trigger an `alive` broadcast when — and only when — the load has moved materially since the last broadcast (e.g. an absolute delta threshold, plus a minimum interval between broadcasts). This keeps the original intent (no incarnation churn on a fixed 10s cadence when nothing is happening) while giving the cluster a sub-second view of a node that actually became busy or idle.

Points to be careful about:

- `UpdateNode(timeout)` blocks waiting for the broadcast to be queued; it must not stall the reporter goroutine or race with `broadcastNodeStatus()` (`server/service.go:853`), which is the only current `UpdateNode` caller and uses the same meta.
- The status tags written by `broadcastNodeStatus` and the load field share one `NodeMeta`; both writers need to go through `ServerDelegate` so neither clobbers the other.
- The `loadTTL` / `loadReportInterval` / broadcast-threshold triple should stay mutually consistent — a threshold high enough to suppress all broadcasts would reintroduce this bug silently.

tinswzy:
Fixes #271.

## The defect

A node samples its load every `loadReportInterval` (10s) and writes it into its own meta — but both steps are purely local. Nothing is sent. The only path that carried the value to a peer was push/pull's *user state* (`LocalState` → `MergeRemoteState`), and memberlist's `pushPull()` contacts `kRandomNodes(1, ...)` per `PushPullInterval` (15s). The expected wait for one specific peer is therefore `15s × (N-1)/2` while `loadTTL` stays pinned at 30s.

Past a handful of nodes the refresh loses the race with the TTL and most peers score the node at the neutral `unknownLoad = 0.5` placeholder. Load-aware selection (#114) silently degenerates toward uniform random — and `WpQuorumSelectionSkew` still reports mode `weighted` throughout, because the *selecting* node's own load is always fresh, so the existing metric cannot see it.

## Why `NodeMeta` is the wrong carrier

The first attempt in this PR made `reportLoadOnce` call `memberlist.UpdateNode()`. That works, and **CI caught a data race on it** — which turned out to be the useful signal, not an accident.

`Members()` returns pointers into memberlist's internal `nodeState` and releases `nodeLock` before the caller reads them, so every field read through one is unsynchronised. That has always been true, but nothing called `UpdateNode` on a schedule: a node's own entry was written once at `Create`, a peer's only when its incarnation advanced. Announcing load that way made `aliveNode()` rewrite `state.Meta/Addr/Port` **every 10s, on every node in the cluster**, turning a latent race into a permanent one.

The race is a symptom. `NodeMeta` is republished *only* through `UpdateNode`, and `UpdateNode` bumps the node's incarnation and makes every peer rewrite that node's table entry. That is the right price for an identity change and the wrong price for a routine sample. The two groups of fields have opposite lifetimes:

| | `node_id`, `az`, `rg`, `endpoint`, `tags` | `load_factor` |
|---|---|---|
| Changes | on deploy/reconfigure | every 10s |
| Nature | event | sample |
| Missing one | wrong node dialled | harmless, next tick corrects |
| Expiry | never | `unknownLoad` after `loadTTL` |

memberlist makes the same split internally: `Node.State` is volatile and is **not** in `Meta` — it has its own message types, its own dedup and its own timers.

## Changes

Load now rides memberlist's user-level broadcast channel (`Delegate.GetBroadcasts`/`NotifyMsg`), which travels the same gossip loop and the same UDP packets as `alive` without touching membership state.

- **New `NodeRuntimeInfo` message** — the counterpart of `NodeMeta`, carrying a node's volatile runtime snapshot (`node_id` + `updated_at` + `load_factor`, ~30 bytes against a full `NodeMeta`), tagged by a leading byte so receivers can ignore kinds they don't know.
- **`runtimeBroadcast` implements `NamedBroadcast`**, so a newer snapshot replaces a node's pending one in O(1) rather than queueing behind it.
- **A snapshot that is new to a node is relayed onward.** This is not optional: memberlist cascades `alive` messages itself (`aliveNode` → `encodeBroadcastNotify`) but does nothing for user messages, so without a relay the originator's own budget — `RetransmitMult × ceil(log10(N+1))` deliveries, ~8 on 16 nodes — is the entire reach. Relaying only on first sight is what makes it terminate, the same role incarnation numbers play for `alive`.
- **`ServiceDiscovery.ApplyLoad`** ingests a reading without disturbing identity fields, and drops anything not newer than what it holds. That is what makes the path safe against gossip retransmits and reordering, and it is what the relay dedupes on. Comparison is always between two readings from the same node, so it relies only on that node's clock moving forward.
- **`UpdateServer` no longer lets a whole-meta update roll back a fresher reading.** push/pull ships the sender's meta as of that exchange and `alive` ships it as of the sender's last `UpdateNode`; both can carry a stale load.
- **`ClientDelegate` ingests the announce too.** A client keeps its own discovery and selects on load, but runs with `PushPullInterval 0` and a no-op `MergeRemoteState` — this is its only path to a server's current load. Ingest only, no relay: a client is a leaf in the gossip graph.

## Shaped to take more than load

`NodeRuntimeInfo` is deliberately not named after its payload — load is not the last volatile thing a node will want to publish. Three choices keep the path generic:

- **`updated_at` versions the whole message, not individual fields.** Every value in a snapshot comes from one sampling pass, so ordering, dedup and relay stay uniform. Per-field stamps would make the relay ambiguous the moment a message carried one fresh and one stale value.
- **`ApplyRuntimeInfo` takes the snapshot itself**, not unpacked arguments, so adding a signal changes the consumer and nothing else — encode, queue, spread, relay and dedup all treat it as opaque.
- **Dedup stays with the consumer**, which already holds the version of what it has; the transport relays whatever the consumer calls news. A second version map in the delegate would duplicate that state and need its own eviction.

Adding disk watermark or in-flight requests is then a proto field plus one line in the scoring path — the transport, the dedup, the relay and `NodeInfo` itself all carry it unchanged.

## Rolling upgrades

`SetLoadFactor` still stamps load into the node's own meta, so push/pull remains the slow fallback for a node running a build that ignores `NodeRuntimeInfo`. Unknown message tags are dropped rather than fatal, and the pre-change `NotifyMsg` was a no-op, so both directions are safe.

The `NodeMeta` load fields are now marked in the proto as retained for compatibility only. Selection results still cross the package boundary as `[]*proto.NodeMeta`, so quorum construction and every caller outside `common/membership` are untouched. #273 tracks dropping the `NodeMeta` load fields entirely once the fallback is unnecessary — a pure internal refactor, since `QuorumNode` has no load field and `quorumNodeFromMeta` already drops it.

## Results

Peers holding a fresh reading, 16-node in-process cluster, one announce:

| | Peers reached |
|---|---|
| Before (#271, push/pull only) | 3.6 / 15 |
| Originator fan-out alone, no relay | 8 / 15 |
| **This PR** | **15 / 15** |

End-to-end latency for a changed load to reach a peer's `discovery`: **under 1s** (measured 0.21s for a peer, 0.21s for a client).

Per announce the cluster now moves ~30 bytes per node instead of a full `NodeMeta`, and no node bumps its incarnation.

## Tests

- `TestReportLoadOnce_DoesNotRaceMemberlistReaders` and `TestReportLoadOnce_LeavesMemberlistNodeTableAlone` — the #272 regression pair. Both **fail on the parent commit** (4 data races) and pass here.
- `TestLoadBroadcast_ReachesEveryPeerInACluster` — 16 real nodes, every peer must pick up the reading. Fails without the relay (8/15).
- `TestLoadBroadcast_ReachesPeerDiscovery` / `..._ReachesClientDiscovery` — end-to-end through the wiring quorum selection actually reads. The client one fails if `ClientDelegate.discovery` is left unwired.
- `TestNotifyMsg_RelaysASnapshotOnlyOnce` — relay termination; fails if the first-sight guard is removed.
- `TestApplyRuntimeInfo_*`, `TestUpdateServer_DoesNotRollBackAFresherLoad`, `TestUpdateServer_AcceptsAFresherLoadFromMeta` — merge and ordering rules.
- `TestServerDelegate_NotifyMsg` / `TestClientDelegate_NotifyMsg` — unknown tags and malformed payloads must not disturb held state.

`go test ./common/membership/ ./common/config/ ./server/... -race` passes, `golangci-lint` reports 0 issues, and `TestStagedStorageService_Chaos_ReaderDuringServerCrash` — the e2e test CI failed on — runs clean locally under `-race`.

> `server/storage/disk` has a pre-existing flaky race unrelated to this PR (`writer_impl_test.go:1231` assigns `writer.file` while the writer's own flush goroutine reads it at `writer_impl.go:851`). Reproduced on `master` with `-count=20`.

## Note for reviewers — do not land a selector change on top of this without damping

Restoring load visibility makes the selector *act* on a signal it has effectively been ignoring. Simulation of the corrected system (described in #271) suggests that with writers re-selecting at segment rollover, mean cross-node imbalance moves from ~1.06x today to ~1.83x once every client can see the same accurate idle-node ranking: they converge on the same nodes, overload them, then flee together. Today's blindness has been accidentally damping that.

This PR is deliberately scoped to the propagation defect. The selector-side damping is tracked separately; the two should be evaluated together before the default is considered tuned.


czs007:

**Verified:**
- `common/membership/server_node.go:288` — `UpdateNode(loadBroadcastTimeout)` with a positive timeout always returns; on a single node `anyAlive()` is false and it returns immediately, so the single-node tests cannot block.
- `common/membership/server_node.go:301-307` — `Shutdown()` cancels the context, waits on `loadWG`, then calls `memberlist.Shutdown()`, so the reporter never touches a closed memberlist and `nodeMap[self]` always exists during `UpdateNode`.
- `common/membership/event_delegate.go:88` — lock order is `nodeLock` → `discovery.mu` only; no path holds `discovery.mu` and re-enters memberlist, so no deadlock.
- `common/membership/server_node.go:183` — concurrent `UpdateNode` from the reporter and `UpdateMeta` is safe: `nextIncarnation` is atomic and `aliveNode` holds the lock.
- `common/membership/server_node.go:163` — the `LoadAwareEnabled && LoadReportInterval > 0` gate is unchanged and the yaml change is comment-only, so there is no upgrade/compat impact.
- `common/membership/server_node_load_test.go` — `fakeSampler` stores values directly with no EWMA path and proto doubles round-trip exactly, so the 0.40/0.88 exact comparisons hold; ports 27810-27818 do not collide with existing package tests.

## Low (non-blocking)

- `common/membership/event_delegate.go:90` — `NotifyUpdate` logs unconditionally, and this PR makes it fire on every report tick: `NodeMeta()` (`server_delegate.go:47`) stamps a fresh `LastUpdate` on every call, so `broadcastLoad` (`server_node.go:288`) always produces changed meta bytes and memberlist invokes `NotifyUpdate` on every node, including self. Before this PR `UpdateNode` was called only once at join (`server_node.go:183`). With defaults (`loadReportInterval=10`, `loadAwareEnabled=true`) in a steady-state N-node cluster with unchanged load → every process (clients included) emits N `[SERVER-EVENT]/[CLIENT-WATCH] Server updated: ...` lines every 10s, growing linearly with cluster size and drowning out genuine membership changes. Suggestion: either skip the log when only `LastUpdate`/load changed (compare the relevant meta fields before logging), or downgrade this line to debug level / rate-limit it. (raised by bigsheeper)
- `common/membership/server_node_load_test.go:161` — `TestReportLoadOnce_NoSamplerDoesNotBroadcast` only asserts that load in meta is 0, which is always true: even if the nil early-return in `reportLoadOnce` (`server_node.go:239`) were removed, `publishLoad`'s own nil check (`server_node.go:253`) keeps load at 0 while `broadcastLoad` (`server_node.go:275`) would still bump the incarnation. The "no announce / no churn" property the test comment claims to guard is not actually exercised. Suggestion: assert that the incarnation is unchanged (or that no broadcast occurred) after `reportLoadOnce()`. (raised by bigsheeper)

## Nitpick (non-blocking)

- PR description — the "Changes" and "Tests" sections still describe the `shouldBroadcastLoad` gate, the new `nodeSelectionPolicy.loadBroadcastThreshold` config (default 0.15, validated to [0,1)), `TestShouldBroadcastLoad` / `TestReportLoadOnce_SmallDeltaDoesNotRebroadcast` / `TestReportLoadOnce_ThresholdZeroDisablesBroadcast`, and "Load-broadcast config is now applied whenever loadAwareEnabled is set". All of this was reverted by the second commit (ce3d5b1 "refactor(membership): drop the load-broadcast threshold, announce eve…"): the final diff contains only the `server_delegate.go` comment, `broadcastLoad` in `server_node.go` (unconditional announce on every tick, per the new comment at `server_node.go:271-272`), three tests, and a `woodpecker.yaml` comment — no config changes. The "Results" table (100% fresh) was also measured on the thresholded version and should be restated for the final behavior. Since squash-merge takes the commit message from the PR body, please rewrite the description before merging. (raised by bigsheeper)

czs007:

Adversarial review of the node-runtime gossip changes surfaced two low-severity concurrency issues, both introduced by this PR and both with straightforward fixes.

## Low

**`common/membership/service_discovery.go:198` — `ApplyRuntimeInfo` CASes onto a `NodeInfo` that `UpdateServer` may already have detached, losing the snapshot.**
`ApplyRuntimeInfo` takes `sd.mu.RLock()`, reads `node` out of `sd.Nodes`, immediately `RUnlock()`s, and then calls `applyRuntimeSnapshot(node, info)` outside the lock (lines 198-204). Meanwhile `UpdateServer` does not mutate the existing `NodeInfo` in place — under the write lock it allocates a fresh `&NodeInfo{Meta: meta}`, copies the old runtime pointer across, and replaces the map entry (lines 149-161). If the two interleave, the CAS lands on the now-orphaned object and still returns `true`, while the live entry keeps only whatever `UpdateServer` carried over. The trigger path is routine, not rare: `MergeRemoteState` (`server_delegate.go:176`) and the `EventDelegate.NotifyJoin/NotifyUpdate` callbacks both call `UpdateServer` concurrently with the `NotifyMsg` receive goroutine.
Failure: node C's receive goroutine handles peer A's *first* `NodeRuntimeInfo` (A's `runtime` still `nil` in C) while a concurrent `UpdateServer("A", metaA)` has already swapped in a new `NodeInfo` → C's live entry keeps `runtime == nil`, `loadOf` reports A as not-known, and A is scored at the neutral `unknownLoad` in C's weighted quorum selection until A's next announce (~10s); C also relays a reading it did not itself store, so the relay "first-sight" semantics and local state diverge at that moment.
Suggestion: hold the read lock across `applyRuntimeSnapshot` — the CAS does not need the write lock, and while the `RLock` is held `UpdateServer` cannot acquire the write lock to swap the map entry. Alternatively, have `UpdateServer` reuse the existing `NodeInfo` and replace only `Meta`.
*(raised by xiaocai2333, sijie-ni-0214)*

**`common/membership/server_node_load_test.go:282` — redundant outer `RLock` around `RuntimeOf`, which takes the same lock again; recursive read-locking can deadlock the test binary.**
The `waitFor` closure in `TestLoadBroadcast_ReachesPeerDiscovery` does `b.discovery.mu.RLock()` / `defer RUnlock()` and then calls `b.discovery.RuntimeOf("peer-a")`, which acquires the same `sd.mu` for reading internally (`service_discovery.go:928-930`). The outer lock is pure surplus: `RuntimeOf` locks for itself, and the `*proto.NodeRuntimeInfo` it returns is documented as immutable, so callers need no lock. Per the `sync.RWMutex` contract read locks are not recursive — once a writer has called `Lock()` and is queued, subsequent `RLock` calls block, so the inner `RLock` waits on the writer and the writer waits on the outer read lock. `b` is a joined, live node whose discovery has periodic writers (`EventDelegate.NotifyJoin/NotifyUpdate` and push/pull `MergeRemoteState`, all via `UpdateServer` holding `sd.mu.Lock()`). The window is narrow, but hitting it is a hard hang rather than a retryable failure. The other two call sites in the same file (`:312`, `:351`) already call `RuntimeOf` without an outer lock, so this is a missed edit.
Failure: during `make unit-test` (`go test -race -short -failfast -timeout=20m ./common/membership/`) an `UpdateServer` write lock lands between the outer `RLock` at line 282 and the inner `RLock` at `service_discovery.go:928` → the goroutines wait on each other permanently, `waitFor`'s 5s bound is never evaluated, and the package hangs until the 20-minute test timeout panics with a goroutine dump, reddening the Unit Test gate.
Suggestion: delete the `RLock`/`defer RUnlock()` on lines 282-283 and call `RuntimeOf` directly, matching `:312` and `:351`.
*(raised by xiaocai2333, sijie-ni-0214)*

PR #276, opened 2026-08-26
fix(meta): stop reader temp info sessions from outliving provider shutdown

tinswzy:
Follow-up to #266. The [adversarial review](https://github.com/zilliztech/woodpecker/pull/266#issuecomment-5422611865) landed nine minutes after that PR merged, so its five findings are fixed here. All five were verified against the merged code first — none were false positives. My [reply on #266](https://github.com/zilliztech/woodpecker/pull/266#issuecomment-5422905407) records the verification.

## Provider shutdown races session creation (High)

`CreateReaderTempInfo` took no provider lock and registered its session in `readerTempSessions` only after the Put. `Close()` took `e.Lock()` but transitioned no state — there was no `closed` field on `metadataProviderEtcd` at all. So nothing made the two mutually exclusive, and nothing stopped a create that began *after* `Close()` had already drained the registry.

Either way the session ends up tracked by nothing. `Close()` does not close the etcd client, and `newReaderTempSession` deliberately builds the session without `concurrency.WithContext`, so the keepalive runs on `e.client.Ctx()` and renews the lease until the process exits. The orphaned reader key then pins `minTruncatedSegmentId` in `log_writer.go:525-529`, which applies no staleness filter — `RecentReadTimestamp` is written but never read by any production caller — so truncated segments are never physically cleaned and object storage for that log grows without bound.

Registration is now gated on a provider-closed flag set under the same mutex `Close()` takes. A create that loses the race revokes its own session and fails; one that wins is in the registry before the drain. There is no third outcome where a live keepalive ends up untracked.

## Serial blocking revocation on shutdown (Medium)

`Close()` held `e.Lock()` across its whole body and revoked one session at a time, each under its own full `requestTimeout` (10000ms, `config/woodpecker.yaml:220`). A SIGTERM against a partitioned etcd therefore cost one timeout per open session — 60 readers is ~10 minutes — and blew past any container termination grace period.

`Close()` is now three stages:

1. Under the lock: mark closed, drain both registries. No network I/O.
2. Outside the lock: stop every keepalive. This is the only stage with a correctness dependency — once nothing renews, every lease expires within one TTL regardless of what follows.
3. Revoke concurrently under a single shared request timeout. Revocation only makes keys disappear sooner than the TTL, so it must never hold shutdown hostage.

Writer locks move onto the same path. They used `concurrency.Session.Close()`, which revokes under a deadline of one full session TTL — the exact API `closeReaderTempSession` was written to avoid, with a comment saying why. Same stall, longer per-item budget. Pre-existing, not from #266, but it is the same shutdown path.

## `UpdateReaderInfoIntervalMs` 5000 → 30000 (Medium)

A regression I introduced in 54d58cb. At 5000 the interval sat below the etcd request timeout, and the report is synchronous on the `ReadNext` path. When a write burned its full 10s timeout, the interval had already elapsed again by the next 200ms poll (`l.lastReported = now` is assigned before the blocking Put, so `now` is loop-entry time), and every batch boundary stalled for a full timeout instead of one stall per interval.

The tighter interval only shortened the key-absence window after a lease loss — which itself requires a full session TTL of continuous etcd unavailability, i.e. exactly the conditions under which it does the most damage. Reverted, with the lower bound now documented and pinned by a test.

## Reader name uniqueness was undocumented (Low)

The #266 description claimed a same-name reopen swaps in a fresh session. `c2874da` did key the registry by `(logId, readerName)` and `Swap` on reopen; `7bb283c` re-keyed it by `*readerTempSession` pointer and dropped that logic, and the description was not updated. Not reachable today — `OpenLogReader` appends `time.Now().UnixNano()` to every reader name — but `DeleteReaderTempInfo` deletes by name, so a same-name reopen would leave two independently keepalived sessions and closing either would remove the survivor's key. The invariant is now stated on `CreateReaderTempInfo` rather than left implicit.

## Regression test for #265 ran in no CI target (Low)

`testReaderTempInfoOutlivesLeaseTTLWhileIdle` is the only case covering the original defect, but it skipped under `-short`. `unit-test` (`Makefile:125`) passes `-short` and includes `./meta/`; `integration-test` (`Makefile:130`) runs only `./tests/integration/...`, which excludes `./meta/`. So it executed nowhere, and a regression of the exact bug #266 fixed would have gone green. Skip removed, TTL reduced to 2s so the wait fits a unit-test budget.

## Testing

- `go build ./...`, `go vet ./meta/ ./woodpecker/log/` — clean
- `golangci-lint run ./meta/... ./woodpecker/log/...` — 0 issues
- `go test -race -short ./meta/ ./woodpecker/log/ ./common/metrics/` — all pass
- `testCreateReaderTempInfoAfterProviderCloseIsRejected` verified failing on the parent commit first (`An error is expected but got nil`)
- `testReaderTempInfoOutlivesLeaseTTLWhileIdle` confirmed executing under `-race -short` (6.62s), not skipped

`TestNewEmbedClientFromConfig_MinioObjectStorageError` fails in `./woodpecker/`, but it fails identically on a clean worktree at `7d1b3b0` (506.66s vs 535.18s) — it dials Azure blob storage with an empty account name and retries to exhaustion. Pre-existing and unrelated; `./common/objectstorage/` has a sibling failure with the same cause.

## Not covered by a test

The shutdown stall duration itself. Reproducing it needs live `concurrency.Session`s — which require a reachable etcd — together with revocations that hang, which require an unreachable one. Closing the client instead makes `Revoke` fail fast rather than time out, so it exercises neither the old nor the new path. The bound is structural instead: stage 1 does no I/O, stage 2 is non-blocking and alone sufficient for correctness, and stage 3 shares one `getContextWithTimeout` across all revocations, so total shutdown cost is one request timeout regardless of session count.

### DIFF
--- a/meta/metadata_etcd.go
+++ b/meta/metadata_etcd.go
@@ -113,6 +113,12 @@ type metadataProviderEtcd struct {
 	// CreateReaderTempInfo handed it to; this is only a registry so provider
 	// Close() can stop every keepalive that would otherwise outlive it.
 	readerTempSessions sync.Map // map[*readerTempSession]struct{}
+
+	// closed is set by Close() under the provider mutex. It is what makes
+	// Close() and CreateReaderTempInfo mutually exclusive: the registry alone
+	// cannot express "this session was created too late to be retired", so
+	// every registration is gated on this flag under the same lock.
+	closed bool
 }
 
 func NewMetadataProvider(ctx context.Context, client *clientv3.Client, cfg *config.Configuration) MetadataProvider {
@@ -1225,10 +1231,33 @@ func (e *metadataProviderEtcd) retireReaderTempSession(ctx context.Context, entr
 	e.readerTempSessions.Delete(entry)
 }
 
+// CreateReaderTempInfo registers one open reader and returns the session that
+// keeps its temp info alive.
+//
+// readerName must be unique among live readers of the same log. Sessions are
+// tracked per session object, not per name, so reopening a name that is still
+// open does NOT swap out the old session: both would keep their own lease
+// alive, and DeleteReaderTempInfo deletes by name, so closing either one would
+// remove the surviving reader's key. Callers satisfy this by making the name
+// unique per open (see OpenLogReader, which appends a nanosecond timestamp).
 func (e *metadataProviderEtcd) CreateReaderTempInfo(ctx context.Context, readerName string, logId int64, fromSegmentId int64, fromEntryId int64) (ReaderTempInfoSession, error) {
 	ctx, sp := otel.Tracer(CurrentScopeName).Start(ctx, "CreateReaderTempInfo")
 	defer sp.End()
 	startTime := time.Now()
+
+	// Fast path: don't grant a lease and write a key we would only have to
+	// revoke again. The authoritative check is at registration time below.
+	e.Lock()
+	alreadyClosed := e.closed
+	e.Unlock()
+	if alreadyClosed {
+		metrics.WpEtcdMetaOperationsTotal.WithLabelValues(e.logNs, "create_reader_temp_info", "error").Inc()
+		metrics.WpEtcdMetaOperationLatency.WithLabelValues(e.logNs, "create_reader_temp_info", "error").Observe(float64(time.Since(startTime).Milliseconds()))
+		logger.Ctx(ctx).Warn("create reader temp info on a closed metadata provider",
+			zap.String("readerName", readerName), zap.Int64("logId", logId))
+		return nil, werr.ErrMetadataWrite.WithCauseErrMsg("metadata provider is closed")
+	}
+
 	// Create a key path for the reader temporary information
 	readerKey := e.keyBuilder.BuildLogReaderTempInfoKey(logId, readerName)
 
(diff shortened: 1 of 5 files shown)


czs007:
> Also reported: 1 pre-existing hazard (real, but not introduced here — follow-up only, does **not** block this PR).

This follow-up to #266 is structurally sound in the create/close race, idempotency, and interval-vs-TTL dimensions; one finding in the new `Close()` code needs attention, and one real hazard in the writer-lock path is pre-existing and should be tracked separately.

## Medium

- **`meta/metadata_etcd.go:1679`** — Stage two is documented as "Stage two: stop every keepalive. Non-blocking, and enough on its own." (metadata_etcd.go:1677), and that claim is the structural argument for the whole `Close()` refactor's cost bound of one shared `getContextWithTimeout`. But stage two's first act is `entry.Lock()`, and that entry mutex is held by the reader side across full etcd round-trips: `UpdateReaderTempInfo` takes `entry.Lock(); defer entry.Unlock()` at 1452-1453 and holds it through up to two `client.Put` calls (1516-1518, one full request timeout each), a possible `newReaderTempSession` (`client.Grant`, 1178-1181), and `closeReaderTempSession` (`client.Revoke`, 1206) on lease loss. `DeleteReaderTempInfo` → `retireReaderTempSession` (1219-1230) likewise revokes under `entry.Lock()`. Since `UpdateReaderTempInfo` is called periodically from the `ReadNext` hot path (woodpecker/log/log_reader.go:194-197), "some entry is doing network I/O under its lock" is the steady state under etcd degradation, not a rare interleaving. Failure: an active reader holds `entry.Lock()` and stalls in `client.Put` under an etcd partition (10s default, config/woodpecker.yaml:220) when the process gets SIGTERM → `Close()` waits a full request timeout before stage three's shared deadline even starts, and staggered lock holders across entries make that wait accumulate serially, breaking the bound this PR's own comment promises and pushing shutdown back toward the container termination grace period. Suggestion: narrow the entry-mutex critical section to "mark closed + `Orphan()`" and move `Put`/`Grant`/`Revoke` outside the mutex; alternatively give `readerTempSession` a lightweight mutex covering only the `closed`/`session` fields so stage two is genuinely non-blocking. (raised by sijie-ni-0214)

- **`meta/metadata_etcd.go:721`** — `AcquireLogWriterLock` registers a newly acquired lock via `e.logWriterLocks.Store(logName, sessionLock)` without ever taking the provider mutex or checking `e.closed`, while `Close()` drains `logWriterLocks` under `e.Lock()` (1660-1668); a `Store` that lands after the drain produces a `SessionLock` that is neither orphaned in stage two (1686-1697) nor unlocked/revoked in stage three (1710-1727), and since keepalive runs on the etcd client context and `woodpeckerEmbedClient.Close` only closes the client when `managedCli == true` (woodpecker/woodpecker_embed_client.go:437-443), an externally injected client keeps renewing the lease so the `concurrency.Mutex` key never disappears — the log becomes unwritable by any takeover process until the holding process exits. The same path also bypasses the embed client's `c.mu`/`closeState` entirely via `logHandleImpl.openLogWriterWithDistributedLocks` (woodpecker/log/log_handle.go:259), so a stale `LogHandle` can acquire a fresh unreclaimable lock after provider close. This is unrelated to this PR: the only edits in `meta/metadata_etcd.go` here are the `closed` struct field (113-121), `CreateReaderTempInfo` (1243-1345), and `Close` (1631-1734) — `AcquireLogWriterLock` (655-724) is untouched, and at base commit a659f4d `Close` already drained `logWriterLocks` under `e.Lock()` while `AcquireLogWriterLock` already took no provider mutex, so the identical race predates this change. No tracking issue yet; recommend gating writer-lock registration (including the fast path at 664-676 that returns an existing lock) on `e.closed` under the same mutex `Close()` uses, with `MarkInvalid` + `mutex.Unlock` + `Orphan`/`Revoke` and an error return for the losing racer. Recommend merging that separate fix before this PR. (raised by sijie-ni-0214, xiaocai2333)

PR #281, opened 2026-08-28
enhance: make log deletion complete and synchronous, add ClearMeta

### LINKED ISSUE #279: Make log deletion complete and synchronous, and add ClearMeta
# Make log deletion complete and synchronous, and add ClearMeta

## Background

Milvus operators need to switch a stopped instance from one WAL backend to another
(e.g. pulsar → woodpecker), and to clear an instance's WAL before a cold restore.
Both flows assume a documented precondition:

> **Before stop / cold backup, the MQ's data and metadata have been cleaned,
> with `logidgen` preserved.**

Today no tool can fulfil that precondition reliably. `DeleteLog` leaves object
storage untouched — harmless for an instance being destroyed outright, but not for
one that is cleared and then reused — its local cleanup is asynchronous with no
completion signal, and its node fan-out is a no-op outside `service` mode. This
issue asks for the capabilities that make the precondition actually achievable.

## What exists today

`woodpecker/delete_log.go` already has the right shape, and its idempotency
contract is correct:

```
1. GetLogMeta → resolve logId (not-found = successful no-op)
2. collect quorum nodes from segment metadata
3. fan out MarkLogDeleted to each node
4. DeleteLogMetadata last, so partial failure is retry-safe
```

`DeleteLog` / `DeleteAllLogs` are exposed on both clients. The gaps are in what
step 3 actually cleans.

### G1 — Object storage is not cleaned per log

`MarkLogDeleted → EvictLog` (`server/logstore.go:886`) writes a delete marker and
closes segment processors. It deletes no files. The background reclaim task
(`server/maintenance.go:197`, `localLogDataDir` at `:116`) then does:

```go
dir = localLogDataDir(cfg, m.Bucket, m.RootPath, m.LogId)
os.RemoveAll(dir)      // local only
```

Object deletion exists (`CleanSegment`, per-segment) but is **not on the delete-log
path**, so everything under `{baseDir}/{logId}/` survives `DeleteLog`.

For an instance that is being **permanently destroyed** this is only wasted space
in the interim: the objects go away when the instance's whole `bucket`/`rootPath`
is torn down.

**The problem is instance reuse.** When an instance is cleared and then restored or
restarted rather than destroyed — which is exactly the flow this issue is about —
those objects stay for the remaining lifetime of the instance. And because
`DeleteLogMetadata` removes the segment metadata that would be needed to enumerate
them, nothing can reclaim them selectively afterwards. They are orphans until the
entire instance is eventually deleted.

### G2 — No way to know deletion finished

Local reclaim is `deleteGracePeriod` (5s) + `deleteReclaimInterval` (2s) behind the
call. The API returns before anything is deleted. A caller that must guarantee
"the WAL is now empty" before restarting Milvus has no way to wait for that.

### G3 — Only `service` mode is handled

`nodeSet` is built from `seg.Metadata.GetQuorum().GetNodes()`. In `minio` / `local`
mode there are no quorum nodes, so step 3 is a no-op and no delete marker is ever
written.

### G4 — No instance-level metadata clear

There is no `ClearMeta`. Deleting the metadata prefix by hand hits the trap in G5.

### G5 — `InitIfNecessary` cannot tolerate a partially preserved instance

`meta/metadata_etcd.go:273`:

```go
} else if len(initOps) != len(keys) {
    // cluster already initialized partially, but not all
    return werr.ErrMetadataInit.WithCauseErrMsg("some keys already exists")
}
```

The four instance-level keys (`instance`, `version`, `logidgen`, `quorumidgen`)
must be all-present or all-absent. Any clear operation that preserves `logidgen`
(see below for why that is required) leaves exactly one key, and startup fails.

## Why `logidgen` must be preserved

`logId` appears in the data paths of both storage tiers:

```
object storage   {minio.rootPath}/{logId}/{segmentId}/…      getSegmentFileKey()
node local disk  {storage.root}/{bucket}/{rootPath}/{logId}/  localLogDataDir()
```

`InitIfNecessary` seeds `logidgen` to `"0"`, so a wiped instance restarts at
`logId=1`. If any object from a previous `logId=1` survives, new writes land in the
same directory.

**And residue cannot be ruled out.** Three paths leave files that nothing knows
about:

1. **A node is permanently down.** `deleteLogUnsafe` is fail-closed today, so the
   delete never completes; in practice an operator ends up forcing past it or
   clearing etcd by hand. Either way that node keeps its files.
2. **`nodeSet` is structurally incomplete.** It is built from *surviving* segment
   metadata. Truncated segments have their metadata deleted
   (`log_auditor.go:210`: *"segments are cleaned in ascending order and their metadata
   is deleted afterwards"*). A node holding files for such a segment — whose own
   cleanup failed, which `segment_impl.go:145` tolerates with a warn-and-continue —
   is not in `nodeSet` at all. **`DeleteLog` returns success while leaving files on
   nodes it never contacted.**
3. **Node replacement.** A node decommissioned between write and delete keeps its
   disk; it may be re-added later.

So "all historical files are gone" is not a guarantee this system can offer.
Correctness has to rest on an invariant instead: **`logId` never goes backwards.**

Only `logidgen` needs this treatment. The test is whether the ID leaks into a data
path:

| key | in a data path? | conclusion |
|---|---|---|
| `logidgen` | yes, both tiers | **must be preserved** |
| `quorumidgen` | no, etcd only (`{prefix}/quorums/{id}`) | safe to clear |
| `instance` | no — only read by legacy-prefix detection and `InitIfNecessary` itself | safe to clear |
| `version` | no, re-initialised | safe to clear |

### W1 — Enumerate by prefix and delete what matches, rather than per segment

Enumerate everything under `{baseDir}/{logId}/`, **filter to woodpecker's own
object patterns**, then delete the survivors concurrently.

Driving the deletion from the prefix rather than from segment metadata is what
makes it complete. Per-segment enumeration misses anything the metadata does not
know about: segments whose metadata write failed after the objects landed,
`write.lock` objects, partial blocks from an aborted flush, and the
warn-and-continue orphans described in G1. Prefix enumeration also reclaims
orphans that have already accumulated. `logId` is never reused, so the prefix is
exclusive to that log — including against any future one.

**Filter rather than delete the prefix wholesale.** A caller with a wrong
`bucket` / `rootPath` / `baseDir` would otherwise turn one bad argument into a mass
deletion of somebody else's data. Matching against the known layout keeps the blast
radius bounded to objects woodpecker could plausibly have written:

```
{baseDir}/{logId}/{segmentId}/{N}.blk        data block
{baseDir}/{logId}/{segmentId}/m_{N}.blk      merged block
{baseDir}/{logId}/{segmentId}/footer.blk     segment footer
{baseDir}/{logId}/{segmentId}/write.lock     writer lock
```

`{logId}` and `{segmentId}` must both be numeric, which is a second guard on the
path shape.

**Anything under the prefix that does not match should be counted and reported,
not silently skipped.** A non-empty residue means either a new object type was
introduced without updating the matcher, or the caller aimed at the wrong prefix.
Both are worth surfacing rather than swallowing.

This additionally resolves an ordering conflict in the current design: "delete
metadata last (for idempotency)" versus "enumerating objects requires metadata".
Driving from the prefix removes the dependency on metadata entirely.

### W2 — Synchronous variant

Return only when all three tiers are clean. Suggested internal ordering:

```
1. MarkLogDeleted fan-out                 ← must be first, and must all succeed
2. concurrently {
     2a. wait for each node's local reclaim to finish
     2b. delete object storage by prefix
   }
3. DeleteLogMetadata                      ← last, keeps the retry-safe contract
```

Step 1 must precede step 2: `EvictLog` is what closes the segment processors and
sets the `deletingLogs` gate, i.e. what actually stops the node from writing.
Deleting objects while a node is still compacting would let it write a new footer
after the delete and recreate the orphan. The two *waits* still run concurrently,
which is where the latency is.

### W3 — Handle `local` / `minio` / `service` separately

In `minio` / `local` mode the client should clean object storage directly rather
than relying on a node fan-out that has no nodes to reach.

### W4 — `ClearMeta`

```go
ClearMeta(ctx, clearLogIdGen bool) error
ClearMetaExceptLogIdGen(ctx) error   // convenience wrapper, clearLogIdGen=false
```

Clears `logs/`, `quorums/`, `logstores/`, `readers/`, `cleaning/`, `marking/`,
`conditionwrite`. Instance-level keys are handled per the flag.

Callers that intend to reuse the instance must use the `ExceptLogIdGen` form.
A small API note: `ClearMeta(ctx, true)` is unreadable at the call site and that
bool is exactly the safe/dangerous boundary — an option (`WithClearLogIdGen()`) or
an explicitly named dangerous variant may read better.

### W5 — `InitIfNecessary` tolerates `logidgen` alone

Relax the invariant to: `instance` / `version` / `quorumidgen` all-present or
all-absent; `logidgen` may exist independently (keep its value if present, seed to
`0` if not).

**W4 and W5 must ship together.** The moment `ClearMeta(false)` exists without W5,
the next startup fails with `some keys already exists`.

Two ways to implement, and doing both is cheapest:
- `ClearMeta` *overwrites* the other three instead of deleting them, so all four
  keys always exist and `InitIfNecessary` sees an initialised cluster;
- `InitIfNecessary` tolerates any subset, as a backstop for interrupted clears and
  for historical half-initialised state (today that state is unrecoverable without
  hand-editing etcd — and seeding `logidgen` to `0` by hand is exactly the mistake
  that causes collisions).

### W8 — Concurrency in `DeleteAllLogs`

Different logs have disjoint object prefixes and disjoint metadata keys, so
per-log concurrency is safe. It should be bounded, so a few hundred logs do not
fan out at once against etcd and object storage.

## Explicitly rejected

**Raising `logidgen` at startup from the object store** (`max(logidgen,
maxObservedLogId)`). It is a read-modify-write across two storage systems and
races with concurrent `CreateLog`: A scans and sees 42 while B advances the
counter to 43, then A writes 42 back. The mechanism meant to prevent regression
would itself cause one.

The backup constraint is documentation instead: a woodpecker metadata backup must
include `{prefix}/logidgen`, and a restore must not lower it.

## Dependency note for callers

birdwatcher intends to call this from an offline ops command. It builds with
`CGO_ENABLED=0` and currently depends on woodpecker only for
`woodpecker/woodpecker/log` (the message-id type). If the delete/clear path can be
reached without pulling in the logstore server half of the tree, that would keep
the integration light.

## Summary

| # | change | priority |
|---|---|---|
| W1 | enumerate `{baseDir}/{logId}/`, delete objects matching woodpecker's layout | required |
| W2 | synchronous variant; fan-out → concurrent waits → metadata | required |
| W3 | handle local / minio / service separately | required |
| W4 | `ClearMeta(clearLogIdGen bool)` + `ClearMetaExceptLogIdGen()` | required |
| W5 | `InitIfNecessary` tolerates `logidgen` alone | required, ships with W4 |
| W8 | bounded per-log concurrency in `DeleteAllLogs` | performance |

tinswzy:
Closes #279.

Makes log deletion actually remove everything, adds a synchronous form, and adds
`ClearMeta` so an instance can be emptied and reused.

## Why

Milvus operators switch a stopped instance between WAL backends and clear an
instance's WAL before a cold restore. Both flows assume the MQ's data and metadata
were cleaned first, with `logidgen` preserved. Nothing could fulfil that: `DeleteLog`
left object storage untouched, its local cleanup was asynchronous with no completion
signal, and its node fan-out was a no-op outside `service` mode.

The reclaim task already flagged the object-storage half as pending:

```go
// Object storage is never touched here (that is an explicit CleanData operation, a later plan).
```

This is that operation, plus what turned out to be needed around it.

## What changed

**Object cleanup driven by the log's prefix** (`woodpecker/log_object_cleanup.go`)

`{minio.rootPath}/{logId}/` is enumerated, objects matching woodpecker's own layout
are deleted concurrently, and anything else is counted and reported.

Driving from the prefix rather than from segment metadata is what makes it complete:
metadata cannot describe objects it never learned about — segments whose metadata
write failed after the objects landed, write locks, partial blocks from an aborted
flush, and blocks whose per-segment cleanup failed and was tolerated with a warn.
It also reclaims orphans that accumulated before this existed.

Matching rather than deleting the prefix wholesale is a safety bound: a caller with a
wrong bucket/rootPath would otherwise turn one bad argument into a mass deletion.
Only these are removed, and `{logId}`/`{segmentId}` must be numeric:

```
{segmentId}/{N}.blk        {segmentId}/m_{N}.blk
{segmentId}/footer.blk     {segmentId}/write.lock
```

Verified against both writers — `objectstorage` and `stagedstorage` (service mode)
build the same `{rootPath}/{logId}/{segmentId}/…` layout.

**Three-phase deletion** (`woodpecker/delete_log.go`)

```
1. MarkLogDeleted fan-out          fence: close processors, raise the gate
2. concurrently { local reclaim | object storage }
3. DeleteLogMetadata               last, so any failure above stays retry-safe
```

Step 1 must precede step 2: `EvictLog` is what stops a node writing, and deleting
objects while a node is still compacting would let it write a fresh footer afterwards
and recreate the orphan.

Two behaviours worth calling out:

- **A missing object-storage client is now an error** rather than a silent skip.
  Deleting the metadata anyway would strand the objects with nothing left to
  enumerate them by.
- **Embedded deployments are fenced even with no segment metadata.** The in-process
  logstore may still hold an open processor. Service mode with an empty quorum
  genuinely has nothing to contact and still short-circuits.

**Synchronous variant** — `MarkLogDeletedRequest.sync` (new field) makes `EvictLog`
reclaim the local directory before replying, instead of deferring to the grace-based
task. Exposed as `DeleteLogSync` / `DeleteAllLogsSync`. Both paths write the marker
first and go through the same `reclaimMarker`, so a crash mid-way is still picked up
by the task.

**`ClearMeta(ctx, clearLogIdGen bool)`** + `ClearMetaExceptLogIdGen(ctx)`

Content metadata is removed; the four instance-level keys are re-seeded in one
transaction so `InitIfNecessary` never sees a partially initialised cluster.

`logidgen` is the one key worth preserving, and the test is whether the ID leaks into
a data path:

| key | in a data path? | |
|---|---|---|
| `logidgen` | yes — both storage tiers | preserved by default |
| `quorumidgen` | no, etcd only | cleared |
| `instance` | no — only read by legacy detection and `InitIfNecessary` | cleared |
| `version` | no, re-initialised | cleared |

Restarting `logId` lets a new log reuse a directory an old log may still have objects
in, and residue cannot be ruled out: a node may be permanently down when its log is
deleted, a node holding data for an already-truncated segment is not discoverable
from metadata at all, and nodes get replaced. Keeping the counter monotonic makes the
collision structurally impossible rather than contingent on a deletion having been
complete.

**`InitIfNecessary` tolerates `logidgen` alone**

The four instance-level keys were all-or-nothing, so a legitimate
`ClearMetaExceptLogIdGen` would leave exactly one and the node would refuse to start.
Only that one combination is now accepted; every other partial state is still an
error, because it means something outside `ClearMeta` touched these keys.

**Bounded concurrency** — `DeleteAllLogs` runs 8 logs at a time (disjoint prefixes and
metadata keys), object deletion 16 at a time. Bounded so a few hundred logs do not
fan out at once against etcd and the object store.

## Not in this PR

- **No `wp` CLI commands.** The consumer is birdwatcher via the Go API. Adding CLI
  surface can follow once the API settles.
- **No startup reconciliation of `logidgen` against object storage.** It was
  considered and rejected: a read-modify-write across two storage systems races with
  concurrent `CreateLog` (scan sees 42, another caller advances to 43, the scan writes
  42 back). The mechanism meant to prevent regression would cause one. The backup
  constraint belongs in documentation instead — a metadata backup must include
  `{prefix}/logidgen` and a restore must not lower it.

## Testing

`go test ./meta/... ./woodpecker/... ./common/... ./server/...` and
`golangci-lint run` pass.

New tests cover the matcher (every object woodpecker writes matches; foreign files,
near misses, non-numeric segment dirs and traversal do not), idempotent re-runs
(empty prefix, object already gone), a genuine storage error surfacing rather than
being mistaken for a clean run, the local-mode nil-storage path, the refusal when
object storage is unavailable, both no-quorum branches, and the `InitIfNecessary`
exception.

Existing `delete_log` tests were updated for the new signatures. One assertion
changed meaning: `TestDeleteLog_NoSegments_JustDeletesMetadata` previously expected no
node contact; embedded deployments are now fenced first, and the service-mode
counterpart that genuinely contacts nobody was added alongside it.

Protobuf was regenerated with the repo's pinned plugins; regenerating without the
schema change produces no diff, so the churn in the generated files is only the new
field and the descriptor shift.

---

## Follow-up: reporting what a delete actually removed

An end-to-end run on minikube (milvus v3.0.0, three-node woodpecker service
deployment) surfaced a gap this PR had left open, and a second commit closes it.

A caller pointed at the wrong `rootPath` — Milvus appends a `/wp` segment when it
builds woodpecker's config, so the value in milvus.yaml is wrong by exactly that
suffix — got a confident success while nothing was deleted:

```
deleted 16 log(s)
```

All 16 `data.log` files survived on all three nodes. Nothing in the return value
could have revealed it: `os.RemoveAll` succeeds on a directory that was never
there, and the object prefix listed empty. Every tier reported done.

### The change

`MarkLogDeletedResponse` gains `local_data_found`, which `EvictLog` fills from a
stat taken before the removal. `deleteLogUnsafe` returns `DeleteStats` — logs,
objects deleted, nodes fenced, and how many of those fences found local data —
and `deleteAllLogsUnsafe` accumulates across the fan-out.

These are facts, not verdicts. A log whose data is already compacted legitimately
has no local directory; an already-empty instance legitimately has no objects. So
nothing here fails — a whole-instance sync clear that deleted metadata while
finding no data anywhere logs a warning naming the bucket and rootPath it used.
Failing instead would break idempotency, since the second run of a successful
clear finds exactly nothing.

`DeleteLogSync` and `DeleteAllLogsSync` now return the stats. Both are new in this
branch, so no released caller changes. The asynchronous `DeleteLog` /
`DeleteAllLogs` keep their signatures and discard the counts, which for them are
meaningless — they return before the reclaim task has run.

### Verified against a populated instance

Getting real data onto all three tiers took some setup worth recording. Segments
only reach object storage after the client auditor compacts them, and milvus
v3.0.0 embeds woodpecker client v0.1.33, which calls the legacy `CompactSegment`
RPC that current servers reject:

```
"auditor maintain the log segment failed"
error="legacy CompactSegment RPC is disabled; upgrade the client to use CompactSegmentWithExpected"
segmentsProcessed=3  segmentsCompacted=0  segmentsFailed=3
```

So with a stock milvus image, **object storage stays empty and that tier is never
exercised at all**. Data was instead written through this branch's own client,
producing genuine `footer.blk` / `m_0.blk` objects in the real layout.

With all three tiers populated, both directions were confirmed:

| run | output | reality |
|---|---|---|
| wrong `rootPath` | `deleted 17 log(s): 0 object(s), 0 of 51 node fence(s) had local data` + warning | 46 objects and 339 `data.log` files untouched |
| correct `rootPath` | `deleted 1 log(s): 11 object(s), 3 of 3 node fence(s) had local data` | objects and local data gone |

The misdirected run also left a concrete artifact of the damage: 46 objects under a
`logId` whose metadata it had deleted, so nothing can enumerate them again. That is
the cost the warning exists to make visible.

A full clear on the same instance took etcd from 119 keys to 4, object storage from
46 to 0, and node-local staging from 68 to 16 `data.log` per node across three
nodes, with `logidgen` held at 49. The synchronous path was exercised for real —
`evicted log ... closedProcessors=2` followed 15ms later by
`reclaim: removed local data directory`, versus the asynchronous path's 5s grace
plus 2s poll.

`go test ./woodpecker/... ./server/ ./meta/...` passes; a new
`TestReclaimMarkerReportsWhetherLocalDataExisted` pins both the found and not-found
cases. Mocks were regenerated with the repo's pinned mockery v2.53.5.

---

### `logObjectPrefix` normalised rootPath while every writer uses it verbatim (High)

`logObjectPrefix` trimmed a trailing slash; `getSegmentFileKey` and
`getFooterObjectKey` concatenate the configured value as-is. That asymmetry is
exactly what `ValidateMinioConfig`'s own comment warns about when it explains why a
non-canonical rootPath is only warned about in minio/local mode — *"every
key/dir/label-building site uses the raw value verbatim, so even a non-canonical
value is self-consistent end-to-end"*. This was the one link that disagreed. With
`rootPath = "root/"` the objects live under `root//{logId}/` while cleanup
enumerated `root/{logId}/`, found nothing, returned `(0, nil, nil)`, and let the
caller delete the metadata that was their last handle.

Fixed by building the prefix verbatim. The test was rewritten too: it had asserted
the buggy behaviour ("must not produce a doubled separator"). It now pins the
property that actually matters — the enumeration prefix must cover the keys the
writers produce, for every rootPath the config layer tolerates (`root`, `root/`,
`wp/data`, `""`, `a//b`) — and must not reach a neighbouring logId.

### The synchronous delete dropped the deleting gate before the caller was done (Medium)

`EvictLog` called `reclaimMarker` inline, and that clears the marker *and* the gate.
So by the time `EvictLog` returned, the gate was gone with object deletion and
metadata deletion still to run. An arriving append would rebuild a segment
processor and write fresh blocks into the prefix the caller was about to enumerate
— invisible to that enumeration, unreferenced afterwards. This violated the
ordering invariant this PR itself documents in `delete_log.go`.

`reclaimMarker` is now split. `removeLocalData` does the removal and nothing else;
the sync path calls only that, leaving the marker and the gate for the grace-based
task to retire. The gate's lifetime therefore matches the asynchronous path instead
of collapsing to zero. `TestSyncEvictKeepsTheDeletingGate` asserts the gate and the
marker both survive a sync evict.

### A transient object-storage failure was permanent (Medium)

`getOrCreateCleanupStorage` used `sync.Once`, and `NewObjectStorage` makes
`BucketExists`/`MakeBucket` calls with retries. One brief outage spent the Once,
left the field nil, and every later delete refused — because deleting metadata
while objects remain would strand them — recoverable only by restarting a client
Milvus keeps for its lifetime. Replaced with a mutex-guarded initializer that
retries on the next call. (The contrast with `getOrCreateObjectStorageClient` is
real: there a spent Once only disables direct read, a performance fallback.)

### Old nodes silently ignored the sync flag (Low)

proto3 drops unknown fields, so a node on an older binary ran an ordinary async
delete and still answered Success with `local_data_found` false — which the client
reported as a completed synchronous delete. Added `sync_applied` to the response,
echoed by any server that understands the flag; a client that asked for sync and
did not get confirmation now fails closed, matching how the legacy `CompactSegment`
RPC treats old clients. Async requests are unaffected: they promise nothing about
local data.

### Parked logs lost their last handle (Low)

Soft deletes copy metadata to `logs-deleted/`, nothing reads that prefix, `ListLogs`
is scoped to `logs/`, and `ClearMeta` deleted the prefix outright. For a log deleted
by an older client — one that did not clean object storage — that parked record was
the only thing that still knew the objects existed.

Rather than documenting the residue, `ClearMeta` now reclaims it. A new
`ListParkedLogIds` harvests the ids (the LogMeta key has exactly one segment after
the prefix; parked segment records sit a level deeper and are skipped by that shape
check rather than by trusting protobuf to reject them), the client sweeps those
objects, and a sweep failure aborts the clear — the same principle as
`deleteLogUnsafe`: never destroy the handle on data you have not managed to delete.
This also reclaims orphans that predate this PR.

### Not addressed: the empty-node-set short-circuit

The review's own analysis is right that this is pre-existing — base commit `ccb3d57`
fences the same zero nodes when every segment has been truncated away, and there is
no node registry to fan out to (`logstores/` has no writer). The residue is already
documented in `meta/clear_meta.go` and backstopped by preserving `logidgen`.

### Verification

`go test ./meta/... ./woodpecker/... ./server/ ./common/...` passes and
`golangci-lint run` reports 0 issues. (Embedded-etcd tests conflict on ports when
several packages run in parallel; they pass with `-p 1`.)

Re-verified on the minikube deployment with all three tiers populated — 100 etcd
keys, 92 objects, 146 `data.log` per node across three nodes:

```
deleted 17 log(s): 46 object(s), 51 of 51 node fence(s) had local data
```

etcd dropped to the 4 instance keys, `logidgen` held at 84, and every object and
local directory this instance still owned was removed. The parked-log sweep is
wired and runs in the real flow:

```
sweepParkedLogObjects: reclaimed objects of soft-deleted logs  parkedLogs=1  objectsDeleted=0
```

(zero because `deleteLogUnsafe` had already cleaned that log's objects — the sweep
exists for records left by clients that did not).


czs007:

This review covers the new log-object cleanup path, the synchronous delete flow, and the metadata-clearing helpers introduced by this PR.

## High

**`woodpecker/log_object_cleanup.go:58` — cleanup prefix is normalized while writers build keys verbatim.** `logObjectPrefix` builds the enumeration prefix with `strings.TrimSuffix(rootPath, "/")`, but both writers concatenate the raw value: `getSegmentFileKey` uses `fmt.Sprintf("%s/%d/%d", baseDir, logId, segmentId)` (`server/storage/objectstorage/writer_impl.go:2476-2478`) and `getFooterObjectKey` uses `fmt.Sprintf("%s/%d/%d/footer.blk", w.rootPath, ...)` (`server/storage/stagedstorage/writer_impl.go:2134`). A trailing-slash `rootPath` is an explicitly supported runtime configuration: `ValidateRootPathValue` flags it as non-canonical, but `ValidateMinioConfig` only prints a stderr WARN for minio/local modes precisely because every key-building site consumes the value verbatim and therefore stays self-consistent (`common/config/configuration.go:521-566`). This new code is the only link in that chain that normalizes. Failure: with `minio.rootPath = "root/"`, objects for `logId=7` live under `root//7/...` while cleanup enumerates `root/7/` → `WalkWithObjects` returns nothing, `deleteLogObjects` returns `(0, nil, nil)`, `deleteLogUnsafe` proceeds to `DeleteLogMetadata` and reports success with `ObjectsDeleted=0`, leaving every WAL object permanently stranded with no metadata left to enumerate it. Suggestion: build the prefix from the unmodified `rootPath` so it matches the writers, and update the expectation at `woodpecker/log_object_cleanup_test.go:38` from `"root/7/"` to `"root//7/"`. (raised by sijie-ni-0214)

## Medium

**`woodpecker/woodpecker_client.go:334` — `sync.Once` makes a transient object-storage init failure permanent for the process.** `getOrCreateCleanupStorage` lazily constructs the client inside `cleanupStorageOnce.Do`; on error it logs a warning and returns, leaving `cleanupStorageClient` nil with the `Once` already consumed. `NewObjectStorage` is not a local-only construction — `NewMinioHandler` performs `BucketExists`/`MakeBucket` network calls with retries (`common/minio/minio_client_helper.go:148-163`), so a transient outage, a credential-service hiccup, or a caller `ctx` expiring during those retries is enough to fail it. Failure: the first `DeleteLog`/`DeleteLogSync` on a client happens while MinIO/S3 is briefly unreachable → every subsequent `DeleteLog`/`DeleteAllLogs`/`DeleteLogSync` on that client returns `ErrMetadataWrite` "object storage client unavailable; refusing to delete metadata while objects remain" (`woodpecker/delete_log.go:127-130`), with no self-healing after storage recovers — only a client rebuild or process restart. Note the contrast with `getOrCreateObjectStorageClient`, where a `Once` failure merely disables direct read; here it disables deletion entirely, and Milvus-side woodpecker clients are long-lived. Suggestion: use a retryable lazy initializer (mutex plus clearing the cached value on failure) instead of `sync.Once`. (raised by sijie-ni-0214, bigsheeper)

**`server/maintenance.go:230` — the sync delete path drops the deletion gate inside `EvictLog`, before any object or metadata deletion has run.** `EvictLog` raises `deletingLogs[logKey]` and detaches the processors (`server/logstore.go:906-913`), then for `sync=true` calls `reclaimMarker` inline (`server/logstore.go:919-926`), which deletes `deletingLogs[key]` at `server/maintenance.go:230`. At that point steps 3 (prefix enumeration and object deletion) and 4 (`DeleteLogMetadata`) have not started. With the gate gone, the rejection check at `server/logstore.go:359-364` no longer applies, and an arriving `AppendEntry`/`Compact` RPC recreates a segment processor and writer — `Close` only nils `currentSegmentWriter` (`server/processor/segment_processor.go:819`) with no `closed` flag, so `getOrCreateSegmentWriter` rebuilds it. Failure: a `DeleteLogSync` (or a direct `POST /admin/log/delete` with `{"sync":true}`, `common/http/management/delete_handler.go:41-42`) against a node still serving the log, with appends or compaction arriving → new blocks/footers are written into `{rootPath}/{logId}/` and the just-removed local directory after the gate is dropped, invisible to the prefix enumeration that already ran and unreferenced by any metadata afterward — permanent orphans. This is exactly the ordering invariant this PR states at `woodpecker/delete_log.go:47-51`; the async path has no such window because the gate survives until the grace-period reclaim task. Suggestion: have the sync path reclaim only the local directory and defer clearing the gate until the client-side deletion completes (or leave it to the grace task). (raised by sijie-ni-0214, bigsheeper)

## Low

**`woodpecker/client/logstore_client_remote.go:367` — old logstore nodes silently ignore the new `sync` flag.** `MarkLogDeleted` sets the new `Sync` field with no version negotiation; proto3 discards unknown fields, so an old node treats the call as an ordinary async delete and returns `Success` with `local_data_found` defaulting to false, which the client reports as fact at `:374`. Failure: during a rolling upgrade, a new client calls `DeleteLogSync`/`DeleteAllLogsSync` against a node still running an old binary → the call returns success with `NodesWithLocalData=0` (and can even trigger the "found no data on any tier" warning at `woodpecker/delete_log.go:315-320`), while that node's `data.log` remains on disk until the grace-period reclaim runs — an operator reading this as "WAL cleared everywhere" may proceed with a backend switch or cold recovery. This also diverges from the repository's existing mixed-version convention of failing closed (`server/service.go:608-613` returns `ErrUnsupportedVersionError` for the legacy `CompactSegment` RPC). Suggestion: when `sync=true`, fail explicitly against nodes whose support cannot be confirmed, or add an explicit `sync_applied` response field and error when it is absent. (raised by bigsheeper, sijie-ni-0214)

**`woodpecker/delete_log.go:266` — logs parked by a previous async delete are never enumerated for cleanup, and `ClearMeta` removes their last metadata handle.** `parkLogInBatches` copies `LogMeta` plus segment records to `logs-deleted/<name>-<ts>/` before removing the active keys (`meta/metadata_etcd.go:2410-2435`), and nothing in the repository consumes that prefix — `BuildLogDeletedKey`/`BuildLogDeletedSegmentKey` are only ever written there. `DeleteAllLogsSync` enumerates via `ListLogs` → `ListLogsWithPrefix("")`, whose prefix is `BuildLogKey("") = "{prefix}/logs/"` (`meta/constant.go:97-98`), which does not cover parked records; the new `meta/clear_meta.go:86` then prefix-deletes `LogDeletedPrefix()` wholesale. Failure: an instance that called `DeleteLog` before this PR (when objects were not deleted, so the log was parked) is upgraded and then runs `DeleteAllLogsSync` followed by `ClearMetaExceptLogIdGen` → both steps succeed with no warning, but those logs' objects remain under `{minio.rootPath}/{logId}/` and, with `logs-deleted/` gone, no path remains to enumerate or reclaim them. Correctness is unaffected (`logidgen` is preserved, so prefixes cannot collide), but the PR's "known-empty state" guarantee does not hold for the bucket. Suggestion: enumerate the `logs-deleted/` prefix as part of `DeleteAllLogsSync` (or before `ClearMeta` drops it), or document the residue explicitly. (raised by bigsheeper)

- **`woodpecker/delete_log.go:177`** — In service mode, `markLogDeletedOnNodes` derives its node set solely from existing segment metadata (`logQuorumNodes`, `:203-228`); when that set is empty it short-circuits to `(0, 0, nil)` and fences nothing. Failure: a log whose segments have all been truncated away is deleted via `DeleteLogSync` → no delete marker is written and no node is fenced, so a node still holding staged local data neither stops serving nor reclaims its directory, while `DeleteLogSync` returns success with `NodesFenced=0`. Unrelated to this PR: the diff only rewrites the previous `for node := range nodeSet` loop (which iterated zero times on an empty set) into an explicit short-circuit, so base commit `ccb3d57` fences the same zero nodes in this scenario. Note also that the suggested "fan out to all registered logstore nodes" has no implementation basis today — `BuildNodeKey` (`meta/constant.go:116-118`) has no non-test callers and nothing writes the `logstores/` prefix, so there is no node registry; the author already documents this residue at `meta/clear_meta.go:47-52` and backstops it by preserving `logidgen`. (raised by sijie-ni-0214, bigsheeper)

czs007:

**Verified:**

- `woodpecker/log_object_cleanup.go:67-70` — the cleanup prefix is now built verbatim as `fmt.Sprintf("%s/%d/", rootPath, logId)` with no `TrimSuffix`, so it aligns exactly with `getSegmentFileKey` / `getFooterObjectKey` key construction, including the `rootPath=""` and `rootPath="root/"` edge cases.
- `woodpecker/log_object_cleanup_test.go:64-80` — `TestLogObjectPrefixMatchesWriterKeys` asserts writer-produced keys start with the computed prefix across five rootPath shapes (`root`, `root/`, `wp/data`, `""`, `a//b`) and that logId 7's prefix does not capture logId 8; the earlier test that encoded the bug in its expectations was removed.
- `server/maintenance.go:189-245` — `removeLocalData` deletes only the log directory while the marker (stored under `rootPath/.deleted/`) and the `deletingLogs` gate are reclaimed atomically by `reclaimMarker`, so the gate stays valid through the object and metadata deletion steps required by the ordering invariant in `delete_log.go:70-83`.
- `woodpecker/client/logstore_client_remote.go:351-354` — the `sync_applied` path is fail-closed: an old node returning Success without the field yields `ErrUnsupportedVersionError`, `markLogDeletedOnNodes` aborts at the fence stage, and metadata is left intact, making the operation retry-safe.
- `woodpecker/woodpecker_client.go:347-370` — `getOrCreateCleanupStorage` takes no nested locks and resets the field to nil on failure so a later call retries, while `deleteLogUnsafe:130-133` still refuses to delete metadata when storage is nil.
- `woodpecker/delete_log.go:338-370` — `sweepParkedLogObjects` runs before `Metadata.ClearMeta` and aborts on failure, and its `logs-deleted/` prefix parsing matches the layout written by `parkLogInBatches` without introducing a new constraint on log names.

Issue #229, opened 2026-07-14
audit: coherent, configurable timeout budget for the wp client → logstore path


tinswzy:
## Summary

Timeouts on the wp-client → logstore path have grown ad hoc: several are hardcoded `30s // TODO configurable`, several control-plane RPCs have **no** deadline at all, and none are documented as a coherent budget. Three findings in #225 (#2 LAC syncer, #3 stream leak, #4 shared batch read timeout) are all symptoms of the same underlying gap — missing or incoherent timeouts. This issue is an **audit + hardening pass**: inventory every timeout, verify each value is sensible for what it bounds, and ensure timeouts nest coherently along each call chain (an inner step must not be able to outlive its caller; a caller must not give up before the work it waits on can reasonably finish).

This is not a single bug fix; it's the systematic follow-up that the scattered `// TODO configurable` markers point at. It subsumes the `lacSyncTimeout` package var added as a hotfix in #228.

## The coherence principle (the core of this issue)

Beyond "make each timeout configurable", the values must be **consistent across upstream/downstream steps**:

1. **Nesting** — an inner step's deadline derives from (and is ≤) its enclosing operation's deadline, so an inner RPC can't outlive the operation waiting on it.
2. **Budget** — for a chain A → B → C, `timeout(A) ≥ timeout(B) ≥ timeout(C)` plus overhead; and where a step retries, `caller patience ≥ maxRetries × per-attempt timeout`.
3. **Semantic fit** — a *durability wait* must exceed `flush interval + object-storage write latency` (otherwise it false-fails already-persisted entries — this is exactly #4); a *control-plane RPC* should be a few seconds; an *advisory push* (LAC) should be short and best-effort.
4. **No unbounded step** on any path a caller can block on.

## Current inventory (client path — to be completed, incl. server + read path)

| Site | Bounds | Current | Bounded? |
|---|---|---|---|
| `woodpecker/segment/append_op.go:214` | wait for single-entry durability ack | `30s` hardcoded (`// TODO configurable`) | yes, per-entry |
| `woodpecker/segment/batch_append_op.go:136` | wait for a whole batch's durability acks | `30s` hardcoded, **shared across the batch** | yes but shared → #225 #4 |
| `woodpecker/segment/segment_handle.go:1237` | LAC fan-out (`UpdateLastAddConfirmed` to quorum) | `lacSyncTimeout` var `30s` (#228 hotfix) | now yes (was unbounded) |
[...]

PR #284, opened 2026-09-02
enhance: bound and retry log deletion, allow an explicit skip for unreachable nodes

### LINKED ISSUE #283: Make log deletion resilient: bound it, retry it, and allow an explicit skip for unreachable nodes
## Background

zilliztech/woodpecker#281 made log deletion complete and synchronous, and
birdwatcher's `reset wal` command now drives it to empty a stopped instance's WAL
before a backend switch or a cold restore.

Operating it surfaced a gap that is not about correctness of the delete itself, but
about what happens when one of the three storage tiers cannot be reached — and the
three tiers are not equally reachable.

## The tiers are not equivalent, and the current code treats them as if they were

| Tier | Reachability | Can the delete always succeed? |
|---|---|---|
| etcd metadata | one endpoint set, always addressable | **yes** — it must succeed |
| object storage | one endpoint, addressable independently of any node | **yes** — it must succeed |
| node-local staged data | one RPC per node in the log's recorded quorum | **no** — a node can be permanently gone |

Object deletion is prefix-driven and client-side (`deleteLogObjects`), so it depends
only on object storage being reachable — never on a log-store node. Metadata
deletion depends only on etcd. Neither has a "the peer no longer exists" failure
mode.

The node fence does. `logQuorumNodes` reads the quorum recorded in each segment's
metadata, and nothing in the repository ever rewrites that record — `Decommission()`
is a server-side operation that runs *on* the node being retired and only affects
quorum selection for new segments. So a node that is permanently gone stays in the
node set of every log it ever served, forever.

Today `markLogDeletedOnNodes` returns on the first node error, which is correct as a
default but is the only behaviour available.

### G1 — No timeout anywhere except the etcd tier

`MarkLogDeleted` passes the caller's context straight through with no deadline of
its own, and the caller's context has none either (birdwatcher builds commands on
`context.WithCancel(context.Background())`).

This is a known, documented hole. `logstore_client_remote.go` says so directly:

> `appendFirstResponseTimeout` bounds the synchronous wait for the first (Buffered)
> response of AddEntry / AddEntries. **The caller passes a deadline-less context**, so
> without this bound a server that accepts the stream but never responds blocks the
> send forever […] TODO make configurable; tracked with the timeout audit (#229).

Only `AddEntry` got that bound. `MarkLogDeleted` did not.

`grpc.WaitForReady(false)` in the pool means an *unconnectable* peer fails fast, but
it does nothing for a peer that accepts the connection and then never answers — a
hung process, a black-holing network, or a node whose synchronous reclaim is stuck.
That wedges the whole delete with no way out but Ctrl-C.

The object-storage tier is unbounded too. `minio.requestTimeoutMs` exists in the
configuration and is given a 1000ms default, but no call site reads it.

### G2 — No retry, so a momentary blip aborts the whole run

There is no retry on the fence path. A pod mid-rolling-restart, a DNS record that
has not re-resolved yet, or a sub-second network hiccup fails the entire
`DeleteAllLogsSync`.

This is the common case, and it is worse than it looks: the pool evicts the cached
connection on transport failure precisely so the *next* call re-resolves DNS, but
there is no next call — the error has already aborted the run. The recovery
mechanism exists and is never given the chance to work.

### G3 — A permanently dead node is an operational dead end

When a node is genuinely gone, the fence can never succeed. `reset wal` then:

- deletes nothing for any log whose quorum includes that node — objects and
  metadata are both preserved, correctly, since metadata is deleted last;
- fully deletes every log whose quorum does not include it;
- aborts before the instance metadata clear;
- and tells the operator `re-run until it succeeds`, which will never happen.

The only escape that works today is undocumented: bring up a replacement node
reachable at the same `host:port`. In a Kubernetes StatefulSet the address is
stable, so recreating the pod restores the fence — the new node reports no local
data, and the objects are deleted by the client regardless. Nothing tells the
operator this.

Note that the safety argument for allowing progress already exists in this
codebase. Preserving `logidgen` was justified by the fact that "all historical files
are gone" is not a guarantee a distributed system can offer, and **"a node
permanently down when its log is deleted" was residue path #1** in that reasoning.
Monotonic `logId` makes leftover files on a dead node structurally unable to collide
with a future log. The residue is bounded and harmless; what is missing is a way to
accept it deliberately.

## Proposed changes

All client-side. No proto change, no server change, and therefore no mixed-version
concern.

### W1 — Bound the object-storage tier

Apply `minio.requestTimeoutMs` where object-storage calls are made, so the tier that
*must* succeed fails in bounded time instead of hanging.

### W2 — Retry the node fence

Wrap the per-node `MarkLogDeleted` in `common/retry` with bounded exponential
backoff, retrying **only** on transport failures — `werr.IsTransportError` already
encodes exactly that predicate and is what the pool uses to evict connections.

A logical error from a reachable node is not a connectivity problem and must not be
retried into a timeout.

### W3 — Bound the node fence

Give the fence a per-attempt deadline so a hung peer cannot wedge the run. This is a
narrow, targeted instance of the timeout audit in #229, not a repo-wide sweep.

The bound has to be generous: a synchronous reclaim does `os.RemoveAll` over a
node's staged directory, which is legitimately slow for a large log. Aborting a
reclaim that is making progress would be worse than the hang it prevents.

### W4 — Optional, explicit skip for unreachable nodes

Add an opt-in delete option (default off) that, **after retries and the timeout are
exhausted and only when the final error is a transport error**, proceeds without
that node instead of failing.

It must not be automatic. Retries cannot distinguish "gone forever" from "down for a
ten-minute maintenance window" — both fail every attempt inside any sane budget, and
only the operator knows which it is. Worse, a network partition is indistinguishable
from a dead node at the client: auto-skipping would delete objects out from under a
node that is still alive and possibly still writing, since the fence is also what
closes the segment processors and raises the deleting gate. That is precisely the
ordering hazard #281 fixed, re-entering through a different door.

The consequence of skipping is deliberate and must be visible, not inferred.

### W5 — Report what was skipped, in enough detail to act on

`DeleteStats` should carry the skipped nodes, the logs affected, and enough
information to locate the residue, so an operator can decide between "the node is
scrap, leave it" and "the node is coming back, reclaim the space by hand".

birdwatcher will render it as, for example:

```
skipped 2 node(s) that could not be fenced:
  milvus-woodpecker-1.milvus-woodpecker-headless:18080
    logs 17,19,23  ->  {storage.rootPath}/{bucket}/{rootPath}/{17,19,23}/
```

Skipping applies to the node-local tier only. Object storage and metadata keep the
current all-or-nothing contract: if either fails, the delete fails and the log stays
enumerable for a retry.

## Summary

| # | change | tier | priority |
|---|---|---|---|
| W1 | apply `minio.requestTimeoutMs` to object-storage calls | object storage | required |
| W2 | retry the fence on transport errors, bounded backoff | node-local | required |
| W3 | per-attempt deadline on the fence | node-local | required |
| W4 | opt-in `SkipUnreachableNodes`, transport errors only | node-local | required |
| W5 | report skipped nodes / logs / residue paths in `DeleteStats` | node-local | required, ships with W4 |

W1–W3 are improvements on their own and carry no policy decision: they fix "one
network blip destroys the run" and "a hung peer wedges it forever". W4–W5 are what
turn a permanently dead node from a dead end into a documented, auditable choice.

## Consumer

birdwatcher's `reset wal` (milvus-io/birdwatcher#535) is the caller. It will expose
W4 as an explicit flag and render W5. `DeleteLogSync` / `DeleteAllLogsSync` /
`ClearMetaExceptLogIdGen` have no callers inside woodpecker and none in Milvus, so
extending their signatures is free.

tinswzy:
Closes #283.

Follows zilliztech/woodpecker#281, which made log deletion complete and synchronous.
Operating that through birdwatcher's `reset wal` surfaced a gap that is not about the
correctness of the delete but about what happens when a tier cannot be reached — and
the three tiers are not equally reachable.

| Tier | Can the delete always succeed? | Why |
|---|---|---|
| etcd metadata | **yes** | one endpoint set, always addressable |
| object storage | **yes** | prefix-driven and client-side; never depends on a node |
| node-local staged data | **no** | one RPC per node in the log's recorded quorum |

`logQuorumNodes` reads the quorum recorded in each segment's metadata, and nothing in
the repository ever rewrites that record — `Decommission()` runs *on* the node being
retired and only affects quorum selection for new segments. A node that is
permanently gone therefore stays in the node set of every log it ever served.

So only the node tier needs retry / timeout / skip. The other two keep the
all-or-nothing contract: if either fails, the delete fails and the log stays
enumerable for a retry.

## Bounded

`MarkLogDeleted` passed the caller's context straight through with no deadline of its
own, and the caller had none either — birdwatcher builds commands on
`context.WithCancel(context.Background())`. The client file already documents this as
an open hole:

> `appendFirstResponseTimeout` bounds the synchronous wait for the first (Buffered)
> response of AddEntry / AddEntries. **The caller passes a deadline-less context**, so
> without this bound a server that accepts the stream but never responds blocks the
> send forever […] TODO make configurable; tracked with the timeout audit (#229).

Only `AddEntry` ever got that bound. `grpc.WaitForReady(false)` makes an
*unconnectable* peer fail fast, but does nothing for one that accepts the connection
and then never answers — a hung process or a black-holing network — which wedged the
whole delete with no way out but Ctrl-C.

The delete mark now has a per-attempt deadline. It is deliberately generous: a synchronous
mark has the node `os.RemoveAll` its staged directory for that log, which is
legitimately slow for a large un-compacted tail, and cutting off a reclaim that is
making progress would be worse than the hang it prevents.

Object deletes are bounded per request for the same reason. The prefix walk is not:
one delete's duration does not scale with anything, so a fixed ceiling is safe, while
a walk's does scale with the object count — a ceiling there would fail exactly the
largest logs it is meant to protect.

## Retried

There was no retry at all, so a pod mid-rolling-restart, a DNS record that had not
re-resolved, or a sub-second hiccup aborted the entire `DeleteAllLogsSync`.

That is worse than it looks. The pool evicts its cached connection on transport
failure precisely so the *next* call re-resolves DNS — but the error had already
aborted the run, so the recovery mechanism never got a turn.

The delete mark now retries with bounded exponential backoff, on transport failures only.
`werr.IsTransportError` already encodes that predicate and is what the pool uses to
evict connections, so the classification is shared rather than re-invented. A
reachable node that rejects the mark is reporting a real problem, not a connectivity
one; retrying it would turn a clear error into a timeout.

## Skippable, but only on purpose

`WithSkipUnreachableNodes()` lets a delete finish without a node that stayed
unreachable through every attempt. **Off by default**, and deliberately not something
retries decide on their own:

- Retrying cannot distinguish "this node is scrap" from "this node is down for a
  ten-minute maintenance window". Both fail every attempt inside any sane budget, and
  only the operator knows which it is.
- A network partition is indistinguishable from a dead node at the client. Auto-skipping
  would delete objects out from under a node that is still alive and possibly still
  writing — marking a log deleted is also what closes its segment processors and raises
  the deleting gate. That is the ordering hazard #281 fixed, re-entering through a different door.

The safety argument for allowing it at all is the one that already justified
preserving `logidgen`: "all historical files are gone" is not a guarantee a distributed
system can offer, and **"a node permanently down when its log is deleted" was residue
path #1** in that reasoning. Monotonic `logId` makes leftover files structurally unable
to collide with a future log, so the residue is bounded and harmless.

Only transport failures are ever skipped, so the option stays honest to its name.

## Reported

`DeleteStats.SkippedNodes` maps node address to the log ids whose delete mark it never
received. Each entry is staged data left on that node forever — nothing references it
once the metadata is gone, and no reclaim runs because the node never got a delete
marker. An operator needs the node and the log ids to tell "scrap hardware, leave it"
from "coming back, reclaim the space by hand", so the detail is reported rather than
summarised into a count.

birdwatcher (milvus-io/birdwatcher#535) will render it as:

```
skipped 2 node(s) that did not accept the delete mark:
  milvus-woodpecker-1.milvus-woodpecker-headless:18080
    logs 17,19,23  ->  {storage.rootPath}/{bucket}/{rootPath}/{17,19,23}/
```

## Scope note

`minio.requestTimeoutMs` is deliberately **not** wired up, despite being the obvious
candidate. It has no reader anywhere in the repository, so its 1s default has never
been exercised; activating it from the delete path would silently impose a tight bound
on every object delete that no operator opted into, and would break slow object stores
that are fine today. Making it live belongs to #229, which can weigh it across all
object-storage calls at once. The delete path uses its own generous bound instead, in
a package var, and says why in the code.

## API

`DeleteLogSync` / `DeleteAllLogsSync` take variadic `DeleteOption`, so existing calls
compile unchanged. Both are new in v0.1.39, have no callers inside woodpecker, and
none in Milvus — birdwatcher is the only consumer.

No proto change, no server change, and therefore no mixed-version concern.

## Testing

`go test ./woodpecker/... ./server/ ./meta/` passes; `golangci-lint run` reports 0
issues. (Embedded-etcd tests contend for ports when packages run in parallel; the run
above is `-p 1`.)

Four new tests, each pinning one decision rather than the mechanism:

- a node Unavailable twice then answering completes the delete — the case that used to
  abort the whole run
- retries exhausted with no option: fails, and `DeleteLogMetadata` is deliberately not
  expected, so reaching it would fail the test
- retries exhausted with the option: the reachable node is marked, the dead one is
  recorded as `{"n1": [7]}`, and the delete completes
- a reachable node returning a logical error is neither retried nor skipped, even with
  the option set

Errors are built as real `codes.Unavailable` statuses, since both the retry and skip
predicates key off `werr.IsTransportError` — a hand-rolled error would exercise
neither path.

## Naming

An earlier revision of this branch called the per-node operation a "fence". That word
already means something specific in this repository — `FenceSegment`, `FencePolicy`,
`ErrSegmentFenced`, `ErrSegmentProcessorFenced`, and the `wpcli logstore intervene
fence` command all refer to sealing a segment against further writes. Using it for
"tell each node to mark this log deleted" put two unrelated mechanisms behind one term
in the same package.

Everything now follows the RPC's own verb, `MarkLogDeleted`, so the code and the wire
protocol agree. `DeleteStats.NodesFenced` becomes `NodesMarked`; that field shipped in
v0.1.39, so it is a breaking change to a released name, taken deliberately rather than
leave the ambiguity permanently in a public struct. birdwatcher is the only consumer,
there are none inside woodpecker, and Milvus does not reference it.

The pre-existing fencing in `server/logstore.go` and `logstore.proto` is untouched.

### DIFF
--- a/server/maintenance.go
+++ b/server/maintenance.go
@@ -182,7 +182,7 @@ func (i *idleProcessorCleanupTask) Run(ctx context.Context) error {
 //
 // It deliberately does NOT touch the marker or the deleting gate. The gate is what makes
 // getOrCreateSegmentProcessor reject an arriving append, so it has to outlive the client-side
-// work that follows a fence — object deletion and metadata deletion. Dropping it here would
+// work that follows the delete mark — object deletion and metadata deletion. Dropping it here would
 // let a new writer rebuild a processor and write fresh blocks into the prefix that was just
 // enumerated, invisible to the enumeration that already ran and unreferenced by any metadata
 // afterwards.
@@ -222,7 +222,7 @@ func removeLocalData(ctx context.Context, store *logStore, m deleteMarker) (bool
 // and prunes the in-memory deleting-set entry.
 //
 // This is the grace-based path. By the time it runs, the grace period has elapsed since the
-// fence, so the client-side deletion it was racing has had its window and the gate can go.
+// mark, so the client-side deletion it was racing has had its window and the gate can go.
 // The synchronous path calls removeLocalData directly and leaves the marker in place, so this
 // task is what eventually retires it.
 //
--- a/woodpecker/delete_log.go
+++ b/woodpecker/delete_log.go
@@ -53,12 +53,12 @@ type DeleteStats struct {
 	Logs int
 	// Objects removed from object storage.
 	ObjectsDeleted int
-	// Successful fence calls, counted per node per log.
-	NodesFenced int
+	// Nodes that accepted the delete mark, counted per node per log.
+	NodesMarked int
 	// Of those, the ones that found a local directory to remove. Always 0 for an
 	// asynchronous delete, which returns before the reclaim task runs.
 	NodesWithLocalData int
-	// SkippedNodes maps a node address to the log ids whose fence it never received,
+	// SkippedNodes maps a node address to the log ids whose delete mark it never received,
 	// because it stayed unreachable through every retry and the caller opted into
 	// WithSkipUnreachableNodes. Empty otherwise — without that option an unreachable node
 	// fails the delete instead.
@@ -71,7 +71,7 @@ type DeleteStats struct {
 	SkippedNodes map[string][]int64
 }
 
-// recordSkippedNode notes that a log's fence never reached a node.
+// recordSkippedNode notes that a log's delete mark never reached a node.
 func (d *DeleteStats) recordSkippedNode(node string, logId int64) {
 	if d.SkippedNodes == nil {
 		d.SkippedNodes = make(map[string][]int64, 1)
@@ -83,7 +83,7 @@ func (d *DeleteStats) recordSkippedNode(node string, logId int64) {
 func (d *DeleteStats) Add(other DeleteStats) {
 	d.Logs += other.Logs
 	d.ObjectsDeleted += other.ObjectsDeleted
-	d.NodesFenced += other.NodesFenced
+	d.NodesMarked += other.NodesMarked
 	d.NodesWithLocalData += other.NodesWithLocalData
 	for node, logIds := range other.SkippedNodes {
 		for _, logId := range logIds {
@@ -115,10 +115,10 @@ func newDeleteOptions(opts []DeleteOption) deleteOptions {
 // window" — both fail every attempt inside any sane budget, and only the operator knows
 // which it is. A network partition is worse still: from the client it is indistinguishable
 // from a dead node, and skipping one would delete objects out from under a node that is
-// still alive and possibly still writing, since the fence is also what closes the segment
+// still alive and possibly still writing, since marking a log deleted is also what closes its segment
 // processors and raises the deleting gate.
 //
-// Only transport failures are ever skipped. A reachable node that rejects the fence is
+// Only transport failures are ever skipped. A reachable node that rejects the mark is
 // reporting a real problem, not a connectivity one, and still fails the delete.
 //
 // What was skipped comes back in DeleteStats.SkippedNodes so the decision leaves a trace.
@@ -172,8 +172,8 @@ func deleteLogUnsafe(
 	logId := logMeta.Metadata.GetLogId()
 
 	// Step 2: stop every node that could still be writing this log.
-	fenceStats, err := markLogDeletedOnNodes(ctx, md, pool, cfg, logName, logId, sync, opts)
-	stats.Add(fenceStats)
+	markStats, err := markLogDeletedOnNodes(ctx, md, pool, cfg, logName, logId, sync, opts)
+	stats.Add(markStats)
 	if err != nil {
 		return stats, err
 	}
@@ -208,12 +208,12 @@ func deleteLogUnsafe(
 	logger.Ctx(ctx).Info("deleteLogUnsafe: log deleted successfully",
 		zap.String("logName", logName), zap.Int64("logId", logId), zap.Bool("sync", sync),
 		zap.Int("objectsDeleted", stats.ObjectsDeleted),
-		zap.Int("nodesFenced", stats.NodesFenced),
+		zap.Int("nodesMarked", stats.NodesMarked),
 		zap.Int("nodesWithLocalData", stats.NodesWithLocalData))
 	return stats, nil
 }
 
-// markLogDeletedOnNodes fences the log on every node that could hold data for it.
+// markLogDeletedOnNodes marks the log deleted on every node that could hold data for it.
 //
 // In service mode the node set comes from the quorum recorded in segment metadata. In the
 // object-storage and local modes there is no quorum to read, and the embedded logstore is
@@ -238,19 +238,19 @@ func markLogDeletedOnNodes(
 	if len(nodes) == 0 {
 		if cfg.Woodpecker.Storage.IsStorageService() {
 			// Service mode with no quorum recorded: every segment was already truncated
-			// away, so no node is serving this log. Nothing to fence.
-			logger.Ctx(ctx).Info("deleteLogUnsafe: no quorum nodes recorded, nothing to fence",
+			// away, so no node is serving this log. Nothing to mark.
+			logger.Ctx(ctx).Info("deleteLogUnsafe: no quorum nodes recorded, nothing to mark",
 				zap.String("logName", logName), zap.Int64("logId", logId))
 			return stats, nil
 		}
 		// Embedded deployments (minio / local) run a single in-process logstore and the
-		// local pool ignores the target, so any target reaches it. Fence it so it stops
+		// local pool ignores the target, so any target reaches it. Mark it so it stops
 		// serving and — with sync — reclaims its directory before we touch anything else.
-		hadData, fenceErr := fenceLogOnNode(ctx, pool, "", bucketName, rootPath, logId, sync)
-		if fenceErr != nil {
-			return stats, fenceErr
+		hadData, deleteMarkErr := markLogDeletedOnNode(ctx, pool, "", bucketName, rootPath, logId, sync)
+		if deleteMarkErr != nil {
+			return stats, deleteMarkErr
 		}
-		stats.NodesFenced = 1
+		stats.NodesMarked = 1
 		if hadData {
 			stats.NodesWithLocalData = 1
 		}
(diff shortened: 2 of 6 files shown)


tinswzy:
Note on the one red check: `codecov/project/Server` reports `-0.05%`, but this PR
touches no server code. The diff is five files, all under `woodpecker/`:

```
woodpecker/delete_log.go
woodpecker/delete_log_test.go
woodpecker/log_object_cleanup.go
woodpecker/woodpecker_client.go
woodpecker/woodpecker_embed_client.go
```

Codecov's own component table agrees — `Server | 83.46% <ø> (-0.05%)`, where `<ø>`
means no lines from this diff land in that component. The drift is measurement noise
between runs, not a regression from these changes.

The component this PR does touch went up: `Client | 90.47% <83.56%> (+0.08%)`, and
patch coverage is 83.56%, above the 80% target — hence `codecov/patch` passing.

Everything else is green: Unit Test, Lint, Go Mod Tidy, Chaos, Component Integration,
E2E Local / Object Storage / Service, and Upgrade Compat Local / Object Storage /
Service.

czs007:

Adversarial review surfaced one issue, in the retry/skip classification of the new per-attempt timeout path; the rest of the change (call sites, option defaults, fail-closed behaviour, concurrency, embedded path) checked out.

## Medium

- **`woodpecker/delete_log.go:368`** — The per-attempt `context.WithTimeout` turns a hung node into a gRPC `DeadlineExceeded`, which `werr.IsTransportError` does not recognise (it only matches `codes.Unavailable` plus the dial/reset substrings), so neither `retry.RetryErr(werr.IsTransportError)` at line 384 nor the skip predicate at line 313 ever fires for it. A quorum node whose process is hung or whose network black-holes replies (TCP/HTTP2 established, `MarkLogDeleted` never completes), called via `DeleteAllLogsSync(ctx, WithSkipUnreachableNodes(), WithMarkAttempts(5))` with an undated or longer parent ctx → after one `markAttemptTimeout` (2 min default) `retry.Do` returns without a second attempt, the skip branch is not taken, the delete hard-fails at line 324, and the cached `ClientConn` to that node is not evicted by `maybeDropCachedConn`. This is exactly the case the `markAttemptTimeout` comment (lines 341-344) and the `WithSkipUnreachableNodes` doc ("a node that stays unreachable through every retry") say the options handle. It is fail-closed, so there is no data-safety consequence, and killing the hung pod converts it into a skippable connection-refused. **Suggestion:** either (a) treat expiry of `attemptCtx` while the parent ctx is still live (e.g. `attemptCtx.Err() == context.DeadlineExceeded && ctx.Err() == nil`, or `werr.IsTimeoutError` guarded by parent-ctx liveness so a caller's own deadline still aborts) as an unreachable attempt in both the retry and skip predicates, and evict the cached conn; or (b) if a possibly-still-alive hung node is deliberately not skippable for safety, state that explicitly in the `WithSkipUnreachableNodes` doc and the `markAttemptTimeout` comment. In either case add a test whose `MarkLogDeleted` mock blocks until its ctx is done and returns `status.Error(codes.DeadlineExceeded, ...)`; the new tests only exercise `codes.Unavailable` (the `unavailableErr` helper) and a logical rejection, so this path is currently untested. (raised by sijie-ni-0214, xiaocai2333)

PR #286, opened 2026-09-02
fix(meta): stop the auditor from truncating a still-Active segment

### LINKED ISSUE #285: [bug] Truncation marking races segment completion: the auditor rewrites a still-Active segment and invalidates its own writer
## Summary

**Scope: this is a steady-state concurrency bug between truncation marking and segment completion, under a single healthy writer.** No second writer, no takeover and no split-brain is involved anywhere in this scenario.

`CheckAndSetSegmentTruncatedIfNeed` does not skip segments that are still `Active`. Under sustained write load a segment stays `Active` for a while after the writer has rolled away from it — its completion is deferred, see [Where the invariant breaks](#where-the-invariant-breaks) — and the truncation frontier can move past it inside that window. The auditor then marks it `Truncated` prematurely, which bumps the segment key's etcd revision. When that segment's perfectly ordinary completion subsequently runs its CAS, it fails with `ErrMetadataRevisionInvalid` — and completion reads that as "another writer has taken over" and calls `NotifyWriterInvalidation` on its own `LogWriter`.

Two things are wrong:

1. `Active → Truncated` is not a legal transition. `docs/compacted_file_cleanup.md` documents the state machine as `Active → Completed --Compact--> Sealed --truncate--> Truncated`.
2. The conflicting writer is the **same client's own auditor goroutine**, not a competing writer — so the resulting writer invalidation is entirely self-inflicted.

For an embedding application like Milvus this surfaces as the WAL being fenced, the PChannel going `UNAVAILABLE`, a term bump and a reassignment — roughly a 45–60s write outage on that channel, repeatedly.

### A note on the word "fence"

It is overloaded, and the two meanings are easy to conflate here:

* **Segment fencing** — `FenceSegment` / `fenceSegmentQuorum` / `fenceAllActiveSegments`. A segment-level operation against the quorum nodes. **This is not what fails in this bug.** In the field timeline below the fence RPCs all succeed.
* **Writer invalidation** — `NotifyWriterInvalidation` -> `onWriterInvalidated`, which is what the embedding application eventually observes as `wal is fenced`.

What actually fails is an ordinary `UpdateSegmentMetadata` CAS on the segment's metadata key in etcd, issued by completion. Nothing in this report is about a fence operation losing a CAS, and no `Fence*` call is a party to the race.

## Production evidence

A Milvus cluster on woodpecker v0.1.37, 32 pchannels, sustained ingest. Counts over one 6-hour window:

```
segment rolling completed successfully : 6897
  of which "will rolling later"        : 6412   (93%)
Marked segments as truncated           : 4848
  of which fromStates contains Active  :  103
meta update revision invalid           :  103
wal is fenced                          :  103
```

**103 : 103 : 103.** Every WAL fencing in that window traces back to an Active-segment truncation; there is no other source.

The truncated-while-Active segment is always `truncationPointSegmentId - 1`, i.e. the segment that has just been rolled away from — never an old one:

```
["Marked segments as truncated"] [logId=14] [segmentIds="[989]"] [fromStates="[Active]"] [truncationPointSegmentId=990]
```

### One instance, end to end

```
15:28:14.574  ["setting segment to rolling_ready state"] [segmentId=989] [queueSize=5]
15:28:14.574  ["Segment is not empty, will rolling later"] [segmentId=989]
15:28:14.587  ["new writable segment created"] [segmentId=990]     <- writes move on; 989 still Active
15:28:14.634  ["Successfully fenced segment on node"] [segmentId=989]  (node A)
15:28:14.634  ["Successfully fenced segment on node"] [segmentId=989]  (node B)
              ... 9.05s waiting on node C ...
15:28:22.669  ["Marked segments as truncated"] [segmentIds="[989]"] [fromStates="[Active]"]   <- revision bumped
15:28:23.681  ["Successfully fenced segment on node"] [segmentId=989]  (node C)
15:28:23.681  ["Calculated LAC from quorum fence responses"] [lac=177] [successCount=3] [ackQuorum=2]
15:28:23.884  ["Quorum complete responses collected"] [lac=177] [finalizedCount=3]
15:28:23.897  ["segment metadata revision is invalid or outdated"] [segmentId=989]
15:28:23.897  ["trigger writer lock session expired"] [reason="segment:989 meta update revision invalid"]
15:28:44      wal is fenced -> PChannel UNAVAILABLE -> term+1 -> reassign
```

The previous roll on the same log took the identical code path and survived only because its completion landed in 58 ms:

```
15:27:39.445  ["new writable segment created"] [segmentId=989]
15:27:39.503  ["segment closed"] [segId=988]                       <- 58ms
15:27:42.689  ["Marked segments as truncated"] [segmentIds="[988]"] [fromStates="[Completed]"]   OK
```

So this is not an exotic corner: **every high-load roll races**, and ~1.6% of them lose.

### The implicit invariant the auditor relies on

`logHandleImpl.CheckAndSetSegmentTruncatedIfNeed` guards purely by segment id — there is no state guard:

```go
if segId >  logMeta.Metadata.TruncatedSegmentId { continue }
if segId == logMeta.Metadata.TruncatedSegmentId { continue }
if segMetadata.Metadata.State == proto.SegmentState_Truncated { continue }  // idempotency only
oldState := segMetadata.Metadata.State
segMetadata.Metadata.State = proto.SegmentState_Truncated   // Active gets rewritten too
err = l.Metadata.UpdateSegmentMetadata(ctx, l.Name, l.Id, segMetadata, oldState)
```

That is sound only under the invariant **"anything before the truncation frontier is already Completed"**, which in turn holds only if segment `N-1` reaches `Completed` before segment `N` accepts its first write. The reasoning chain is:

```
frontier reaches segment N  <=  the application consumed a message in segment N
                            <=  segment N accepted a write
                            <=  segment N exists
```

### Where the invariant breaks

`segmentHandleImpl.SetRollingReady` has two paths and only one preserves it:

```go
s.rollingState.Store(true)
if s.appendOpsQueue.Len() > 0 {
    logger.Info("Segment is not empty, will rolling later")
    return nil                                   // (B) returns immediately, no completion
}
s.completionMgr.TriggerCompletion()
...
return s.completionMgr.WaitForCompletion()       // (A) blocks until Completed is published
```

`logHandleImpl.GetOrCreateWritableSegmentHandle` creates segment `N` unconditionally once this returns.

* **Path A** (empty queue, low load): `N-1` is `Completed` before `N` exists. Invariant holds.
* **Path B** (non-empty queue, high load): `N` is created and starts taking writes while `N-1` is still `Active`; its completion is deferred to whenever the queue drains. **Invariant broken.** This is 93% of rolls under load.

Once `N` takes a write, the application's checkpoint moves into `N` within sub-second, the frontier advances to `N`, and `N-1` becomes truncation-eligible while its completion is still doing fence/complete RPCs.

### Why completion always loses the CAS

The auditor reads segment metadata fresh (`GetSegments` -> `GetAllSegmentMetadata`) immediately before writing. Completion CASes against `segmentMetaCache`, i.e. the revision captured when the segment was **created** — held across the whole fence + complete round trip. The stale side loses by construction.

`metadataProviderEtcd.UpdateSegmentMetadata` compares `ModRevision(segmentKey) == segmentMeta.Revision`, and `segmentHandleImpl.doCloseWritingAndUpdateMetaIfNecessaryUnsafe` reacts:

```go
if err != nil && werr.ErrMetadataRevisionInvalid.Is(err) {
    // metadata revision is invalid or outdated, some one updated it, trigger fence ...
    s.NotifyWriterInvalidation(ctx, fmt.Sprintf("segment:%d meta update revision invalid", s.segmentId))
```

`some one updated it` is true; the inference that it was **another writer** is not.

### Timing

Three independent clocks have to line up, and only one of them is under woodpecker's control:

| clock | length | set by |
| --- | --- | --- |
| Active window (`N-1` stops taking writes -> `Completed` CAS) | 58 ms .. 9 s+ | the **slowest** quorum node — fence and complete both collect all `nodeCount` responses |
| frontier crossing `N-1` | sub-second | the application's checkpoint; writes move to `N` immediately after the roll |
| auditor sampling | 10 s | `client.auditor.maxInterval` |

Because the second is far shorter than the first, the collision condition reduces to *an auditor tick landing inside the Active window*: `P ~ window / 10s`. A 58 ms window is ~0.6%; a 9 s window is near-certain.

## Side effect: the truncated metadata is malformed

The truncation loop only rewrites `State`, so a segment truncated straight out of `Active` keeps the field values it was born with in `storeNewSegmentMeta`:

* `LastEntryId` stays `-1` — the segment's extent is permanently lost.
* `CompletionTime` stays `0`, so the GC TTL check in `cleanupTruncatedSegmentsIfNecessary` falls back to `CreateTime` and the segment is reclaimed **earlier** than a normally-completed one.

This is a second reason to prevent the transition rather than only soften the reaction to it.

## Not the cause

* **Retention TTL.** `Woodpecker.Logstore.RetentionPolicy.TTL` is consumed in exactly one place, the `ttlCutoffMs` cutoff that decides when an **already-`Truncated`** segment is physically reclaimed. It does not participate in frontier advancement or in state marking. Lowering it (72h -> 30m to recover disk) neither caused nor worsened this.
* **Disk pressure.** Pool disk was ~6% and pod CPU ~0.34-0.51 cores while reproducing.
* **Rolling upgrades / PChannel migration.** These have an operator action as their cause; this one fires with no external trigger.

## Proposed fix

Add a state guard to the truncation loop in `logHandleImpl.CheckAndSetSegmentTruncatedIfNeed`: only transition **`Completed` / `Sealed`** to `Truncated` (an allowlist rather than `!= Active`, so a future state cannot silently fall through). An `Active` segment before the frontier is simply skipped and picked up on a later auditor cycle, once its completion has published `Completed`.

Cost: a segment is marked `Truncated` up to one auditor interval (10s) later, delaying its GC by the same amount. No other behaviour changes.

### What this fix deliberately does not change

A completion CAS that loses to a **genuine** takeover — a new writer's `fenceAllActiveSegments` racing an old writer's still in-flight completion on the same `Active` segment — keeps fencing, and that stays correct.

That path is not a party to this bug. `fenceAllActiveSegments` runs at open time, *before* the new writer's auditor is started (`openInternalLogWriter` fences first, then `NewInternalLogWriter` -> `startAuditor`), so truncation marking can never collide with the open-time fence batch. The only interaction is completion vs. fence, and there losing the CAS genuinely does mean the writer was preempted.

Plus a deterministic regression test: roll with a non-empty append queue, stall the fence, drive an auditor cycle, and assert that no `fromStates=[Active]` transition and no `NotifyWriterInvalidation` occur.

## Considered and rejected

**Making completion re-read on CAS conflict and classify the winner.** The point of the table below is *not* that fence is involved in this bug — it is not. It answers a narrower question: once the state guard is in place, what can still make a completion CAS fail? The answer is only `fenceAllActiveSegments` on the reopen path — which *is* a genuine takeover, and where fencing is the correct response. Every other writer is excluded:

| writer of the segment key | can still make **completion's** CAS fail? |
| --- | --- |
| auditor truncate | no, after this fix |
| compaction (`Compact`) | no — it re-reads and returns `ErrSegmentHandleSegmentStateInvalid` unless the state is `Completed` |
| completion re-entry | no — `completionManager.run` is a single `select`, one-shot |
| reopen fence | **yes — and that is a real takeover** |

So the existing inference becomes sound rather than needing to be replaced, and adding a re-read would only cost a round trip while weakening a correctness signal.

**Marking the truncated segments in ascending order, or stopping at the first segment that is not yet eligible**, so that the truncated set is always a contiguous prefix. Two sweeps do assume a prefix — `cleanupTruncatedSegmentsIfNecessary` passes `segmentIdsToClean[0]` to `CleanupOrphanedStatuses`, and `orphanSweepBound` takes the smallest existing segment id — but a gap is self-healing: `CheckAndSetSegmentTruncatedIfNeed` runs every auditor tick against a fresh `GetSegments`, so a segment skipped for being `Active` (or missed by a transient update failure) is picked up on the next cycle. A gap therefore lives at most one auditor interval, and the only reachable consequence — a still-`Sealed` segment's `marking/` record being reaped ~10s before it would have been reaped anyway — is inert. Ordering the loop would also change nothing observable, since the log line is already sorted before it is emitted. Stopping at the first ineligible segment would additionally block GC behind a stalled segment, which is a liveness trade worth its own discussion rather than a rider on this fix.

**Returning from fence/complete as soon as `Aq` responses arrive**, to shrink the Active window. This is unsafe. `resolveFenceLAC` returns `sorted[len(sorted)-ackQuorum]`, the Aq-th highest response. With 3 nodes reporting `[100, 150, 150]`, waiting for all yields `150`; returning after the first two to arrive can yield `sorted[0] = 100` and complete the segment short of entries that were already acknowledged to the client. `preserveKnownCompletionLAC` only covers a writer that still remembers acking them, which the recovery path does not. Waiting for every node is a deliberate correctness choice; the 9 s wait is diagnostic colour explaining which roll lost, not a defect. After the state guard, a slow node costs only one extra auditor cycle of GC delay.

## Affected versions

v0.1.37, v0.1.38 and current `master` — `git diff v0.1.37..HEAD` shows no change to either the truncation loop or `segment_handle.go`. Upgrading does not help.

tinswzy:
Fixes #285.

## What

`CheckAndSetSegmentTruncatedIfNeed` guarded only by segment id — everything strictly below the truncation frontier that was not already `Truncated` got rewritten, `Active` included. This restricts the transition to `Completed` and `Sealed`.

## Why it is reachable in normal operation

Single healthy writer, no takeover, no split-brain.

A roll that finds a non-empty append queue returns from `SetRollingReady` immediately without triggering completion, so `GetOrCreateWritableSegmentHandle` creates segment `N` and it starts taking writes while `N-1` is still `Active` with its completion deferred. The frontier moves into `N` within sub-second; the auditor, sampling every 10s, can catch `N-1` mid-completion. Marking it bumps the segment key's etcd revision, so completion's CAS — made against the revision the segment was created with, held across the whole fence + complete round trip — fails with `ErrMetadataRevisionInvalid`. Completion reads that as a writer takeover and fences its own `LogWriter`.

The conflicting writer is the client's own auditor goroutine, so the fence is entirely self-inflicted. On a production cluster this accounted for **every** WAL fencing observed — over one 6-hour window: 103 Active truncations, 103 revision-invalid errors, 103 fencings, against 6897 rolls of which 6412 (93%) took the deferred-completion path.

The issue has the full causal chain, a field timeline, and the two alternatives that were considered and rejected.

## The change

```go
// Only a finalized segment may be truncated. [...]
if segMetadata.Metadata.State != proto.SegmentState_Completed &&
    segMetadata.Metadata.State != proto.SegmentState_Sealed {
    continue
}
```

An allowlist rather than `!= Active`, so a future state cannot silently fall through.

## Cost

An `Active` segment below the frontier is skipped and picked up on the next auditor cycle once completion has published `Completed` — up to one auditor interval (10s) of extra GC delay, nothing else. Gaps are self-healing: every tick re-evaluates a fresh `GetSegments`.

It also keeps the resulting metadata well-formed. A segment truncated straight out of `Active` kept `LastEntryId = -1` and `CompletionTime = 0` from `storeNewSegmentMeta`, permanently losing its extent and making the GC TTL count from `CreateTime` instead of `CompletionTime`.

## Tests

New `TestLogHandle_CheckAndSetSegmentTruncatedIfNeed_OnlyTruncatesFinalizedSegments`, three cases:

* `SkipsActiveSegmentBelowFrontier` — the regression. Asserts no update is issued for the `Active` segment and that no transition originates from `Active`. Since the fencing is caused precisely by that write bumping the revision, asserting the write does not happen is the same as asserting the completion cannot be made to lose.
* `PicksUpTheSegmentOnceCompletionPublishesCompleted` — two consecutive cycles, proving the skip costs exactly one cycle and loses nothing.
* `SkipsSegmentsAlreadyTruncated` — no `UpdateSegmentMetadata` expectation, so any call fails as unexpected.

Verified the new test fails on the unfixed code:

```
--- FAIL: .../SkipsActiveSegmentBelowFrontier
    Error: []int64{1, 2, 3} should not contain 2
    Error: []proto.SegmentState{2, 0, 3} should not contain 0     # 0 == Active
```

Two existing tests asserted the old behaviour and are updated to the intent they were actually covering:

* `_WithActiveSeg` -> `_SkipsFrontierAndLaterSegments`. Its real coverage is the id-based skips of the frontier segment and everything after it; its below-frontier segment moves from `Active` to `Sealed` so it no longer depends on the illegal transition.
* `_UpdateError` — its below-frontier segment moves from `Active` to `Completed` so the update-error branch is still exercised.

Local run (matching the CI unit-test invocation):

```
go test -race -short -failfast -timeout=20m ./woodpecker/log/... ./woodpecker/segment/... ./meta/...
ok  github.com/zilliztech/woodpecker/woodpecker/log      28.830s
ok  github.com/zilliztech/woodpecker/woodpecker/segment  41.493s
ok  github.com/zilliztech/woodpecker/meta                17.735s
```

`tests/integration/e2e_truncate_test.go` calls `CheckAndSetSegmentTruncatedIfNeed` directly to persist Truncated state, but writes synchronously with `logWriter.Write` and only checks after all writes are done, so every segment below its truncation point is long since `Completed`. It needs etcd/object storage and was not run locally — the e2e workflows cover it.

### DIFF
--- a/woodpecker/log/log_handle.go
+++ b/woodpecker/log/log_handle.go
@@ -907,8 +907,19 @@ func (l *logHandleImpl) CheckAndSetSegmentTruncatedIfNeed(ctx context.Context) e
 			continue
 		}
 
-		if segMetadata.Metadata.State == proto.SegmentState_Truncated {
-			// already truncated, skip
+		// Only a finalized segment may be truncated. Truncated is already done, and an
+		// Active segment still has an in-flight completion holding the revision the
+		// segment was created with: marking it here bumps that revision, so completion's
+		// CAS fails with ErrMetadataRevisionInvalid, which it reads as a writer takeover
+		// and answers by fencing its own LogWriter. Such a segment exists whenever a roll
+		// found a non-empty append queue -- the new segment starts taking writes, and the
+		// truncation frontier can reach past the old one, while its completion is still
+		// running fence/complete RPCs. Skipping it costs one auditor cycle: the next pass
+		// sees Completed and truncates it then. It also keeps the truncated metadata
+		// well-formed, since an Active segment still carries LastEntryId=-1 and
+		// CompletionTime=0 from storeNewSegmentMeta.
+		if segMetadata.Metadata.State != proto.SegmentState_Completed &&
+			segMetadata.Metadata.State != proto.SegmentState_Sealed {
 			continue
 		}
 
--- a/woodpecker/log/log_handle_test.go
+++ b/woodpecker/log/log_handle_test.go
@@ -1877,8 +1877,10 @@ func TestLogHandle_ShouldMoveToTruncationPoint(t *testing.T) {
 	})
 }
 
-// TestLogHandle_CheckAndSetSegmentTruncatedIfNeed_WithActiveSeg tests truncation with active segment needing update
-func TestLogHandle_CheckAndSetSegmentTruncatedIfNeed_WithActiveSeg(t *testing.T) {
+// TestLogHandle_CheckAndSetSegmentTruncatedIfNeed_SkipsFrontierAndLaterSegments tests that
+// only finalized segments strictly below the truncation point are marked: the frontier
+// segment itself and anything after it are left alone regardless of their state.
+func TestLogHandle_CheckAndSetSegmentTruncatedIfNeed_SkipsFrontierAndLaterSegments(t *testing.T) {
 	logHandle, mockMeta := createMockLogHandle(t)
 	ctx := context.Background()
 
@@ -1891,7 +1893,7 @@ func TestLogHandle_CheckAndSetSegmentTruncatedIfNeed_WithActiveSeg(t *testing.T)
 	}, nil)
 
 	segments := map[int64]*meta.SegmentMeta{
-		0: {Metadata: &proto.SegmentMetadata{SegNo: 0, State: proto.SegmentState_Active}, Revision: 1},
+		0: {Metadata: &proto.SegmentMetadata{SegNo: 0, State: proto.SegmentState_Sealed}, Revision: 1},
 		3: {Metadata: &proto.SegmentMetadata{SegNo: 3, State: proto.SegmentState_Completed}, Revision: 1},
 		5: {Metadata: &proto.SegmentMetadata{SegNo: 5, State: proto.SegmentState_Active}, Revision: 1}, // truncation segment - skip
 		8: {Metadata: &proto.SegmentMetadata{SegNo: 8, State: proto.SegmentState_Active}, Revision: 1}, // after truncation - skip
@@ -1901,7 +1903,7 @@ func TestLogHandle_CheckAndSetSegmentTruncatedIfNeed_WithActiveSeg(t *testing.T)
 	// Only segments 0 and 3 (before truncation seg 5) should be marked as truncated
 	mockMeta.EXPECT().UpdateSegmentMetadata(mock.Anything, "test-log", int64(1), mock.MatchedBy(func(sm *meta.SegmentMeta) bool {
 		return sm.Metadata.State == proto.SegmentState_Truncated && sm.Metadata.SegNo == 0
-	}), proto.SegmentState_Active).Return(nil)
+	}), proto.SegmentState_Sealed).Return(nil)
 	mockMeta.EXPECT().UpdateSegmentMetadata(mock.Anything, "test-log", int64(1), mock.MatchedBy(func(sm *meta.SegmentMeta) bool {
 		return sm.Metadata.State == proto.SegmentState_Truncated && sm.Metadata.SegNo == 3
 	}), proto.SegmentState_Completed).Return(nil)
@@ -1924,7 +1926,7 @@ func TestLogHandle_CheckAndSetSegmentTruncatedIfNeed_UpdateError(t *testing.T) {
 	}, nil)
 
 	segments := map[int64]*meta.SegmentMeta{
-		0: {Metadata: &proto.SegmentMetadata{SegNo: 0, State: proto.SegmentState_Active}, Revision: 1},
+		0: {Metadata: &proto.SegmentMetadata{SegNo: 0, State: proto.SegmentState_Completed}, Revision: 1},
 	}
 	mockMeta.EXPECT().GetAllSegmentMetadata(mock.Anything, "test-log").Return(segments, nil)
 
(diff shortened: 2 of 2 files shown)


czs007:

Verified:
- `woodpecker/log/log_handle.go:921-924` — the explicit `Completed`/`Sealed` allowlist covers every reachable state in the enum (Active/Completed/Sealed/Truncated, value 1 reserved), so `Active`, already-`Truncated`, and future states are never rewritten by the auditor.
- `woodpecker/log/log_handle.go:899-908` — segment-ID boundaries are unchanged; only segments strictly below the truncation frontier are candidates.
- `woodpecker/segment/segment_handle.go:897-927` — completion still performs its own `Active -> Completed` transition using the creation-time revision CAS, and this auditor path can no longer race it into a self-inflicted fence.
- `meta/metadata_etcd.go:852-871` — the revision CAS still rejects races with completion or compaction, so the new state check does not weaken concurrent-update safety.
- `woodpecker/segment/segment_handle.go:390-392` and `:491-493` — both success and failure callbacks trigger completion once the queue drains, so a skipped `Active` segment on the normal path is picked up on the next audit cycle.
- `woodpecker/log/log_handle_test.go:2432-2517` — new subtests cover skipping an `Active` segment below the frontier, picking it up on the next cycle after it becomes `Completed`, and issuing no redundant update for already-`Truncated` segments.

## Low (non-blocking)

- `woodpecker/segment/segment_handle.go:903` — An `Active` segment whose completion path never publishes `Completed` is now skipped by the auditor for the entire lifetime of the current writer, not merely delayed by one auditor interval as the PR description states. Failure scenario: all appends for a segment ultimately fail (`segment_handle.go:483-492` sets `rollingState` and triggers completion with `lastAddConfirmed = -1`), fence/complete resolve LAC to -1, completion exits early at `segment_handle.go:903` without writing `Completed`, writes roll to a new segment, and truncation advances the frontier past it while the same writer keeps running → the segment stays `Active` with `LastEntryId = -1`, `CheckAndSetSegmentTruncatedIfNeed` skips it every cycle, and its etcd metadata key and server-side directory are never reclaimed until the writer is reopened and `fenceAllActiveSegments` publishes `Completed` (`segment_handle.go:1765-1772`). Before this PR the parent's check at `log_handle.go:910-913` only skipped `Truncated`, so the same segment would have been truncated and cleaned up on the next cycle. Impact is limited to leaked metadata and storage for such empty segments during the writer's lifetime; reads are unaffected and cleanup of other `Truncated` segments is not blocked. Suggestion: either document this exception in the PR description (the "up to one auditor interval of extra GC delay" cost statement is currently inaccurate), or have the `lastFlushedEntryId == -1` path at `segment_handle.go:903` also publish `Completed` with `LastEntryId = -1`, matching the behavior of `FenceAndComplete`. (raised by bigsheeper)

PR #294, opened 2026-09-04
fix(topology): say when placement is unset, and never label it ""

### LINKED ISSUE #292: [bug] An unset REGION/AVAILABILITY_ZONE silently becomes an empty label instead of failing loudly
## Summary

`topology.GetCurrentRegion()` and `GetCurrentAvailabilityZone()` fall back to the **empty string** when their environment variables are unset (`common/topology/topology_utils.go:50-56`):

```go
func GetCurrentRegion() string           { return envOrDefault(RegionEnvKey, "") }
func GetCurrentAvailabilityZone() string { return envOrDefault(AvailabilityZoneEnvKey, "") }
```

Nothing checks the result. The server passes them straight into its membership registration (`server/service.go:105-107`) with no validation and no log line:

```go
ClusterName: topology.GetCurrentClusterName(),
Region:      topology.GetCurrentRegion(),
AZ:          topology.GetCurrentAvailabilityZone(),
```

From there the empty values propagate silently:

1. they are published to gossip and persisted into `QuorumInfo.replicas[].region/az` when a segment is created;
2. the client reads them back and computes `topology.Scope(...)`, whose first condition is `if localRegion == "" || localAZ == "" || peerRegion == "" || peerAZ == ""` → `ScopeUnknown` (`topology_utils.go:27-28`);
3. `activeSegmentNodes` leaves `AZ` at its zero value when a placement is missing, so `woodpecker_client_active_segment_node` is emitted with `az=""`.

**A misconfigured deployment therefore produces metrics that look structurally fine and are entirely uninformative.** The same applies to the client side, which calls the same two functions (`woodpecker/segment/segment_handle.go:633-634, 686-687`) — if the *client* is missing them, every scope is `unknown` no matter how well the servers are configured.

## Field evidence

On a UAT cluster this went unnoticed for **49 days**:

* every `az_scope` label on `woodpecker_client_replica_append_total` and `woodpecker_client_quorum_read_total` was `unknown`;
* decoding a live segment's `QuorumInfo.replicas` showed `region` and `az` **absent** on all three replicas, while `cluster_name` and `resource_group` read `default` — the giveaway, since `default` is `envOrDefault`'s fallback for `CLUSTER_NAME` (`topology_utils.go:47`);
* nothing in any log, metric or health check indicated that placement information was missing.

The metric family added by `#269` to make cross-AZ traffic observable was, in that deployment, structurally incapable of observing anything — and there was no way to tell from the outside.

## Why `""` is worse than `"unknown"`

`count by (az) (woodpecker_client_active_segment_node{...})` with an empty label renders in Prometheus/Grafana as:

```
{}   96
```

which reads like a perfectly normal aggregation with no `az` breakdown — not like a missing value. `topology.Scope` already defines `ScopeUnknown = "unknown"` for exactly this situation; the placement labels should use the same convention so that "we do not know" is *visible* rather than blank.

## Proposed change

Three parts, which belong together because the alert depends on the label value:

1. **Validate at startup and say so.** When `REGION` or `AVAILABILITY_ZONE` is unset, log a WARN naming the missing variable and mark the node's placement as unknown in the startup banner. Do the same on the client when it computes its own local placement. Failing hard is probably too strong — placement is genuinely optional for single-AZ deployments — but silence is not acceptable.
2. **Emit `"unknown"` rather than `""`** for `region`/`az` wherever they become a metric label, aligning with `ScopeUnknown`.
3. **Ship an alert rule** so a misconfiguration surfaces without anyone opening a dashboard:
   ```promql
   count(woodpecker_client_active_segment_node{az="unknown"}) > 0
   ```

The repo's client templates currently have `Active Segment → Nodes (by Log)` and `Active Segments per Node` (`count by (node)`), but **no placement panel** — nothing groups by `az` at all.

Add a **Placement by AZ** panel to both:

```promql
count by (az) (woodpecker_client_active_segment_node{<file's usual selectors>})
```

With change (2) in place, a misconfigured cluster shows a single `unknown` bucket instead of a blank one — self-evident on the panel.

## Scope note

The immediate cause of the observed UAT case was **deployment-side**: two older pools were created from a template that lacks the topology init container, so their `woodpecker server` processes ran with `--region "" --availability-zone ""`. A newer pool in the same cluster, on the same image, was configured correctly and reported its placement fine. Fixing those deployments is separate.

**This issue is about the code accepting that condition silently for 49 days.** A cluster that cannot determine its own placement should say so.

## Acceptance criteria

* Starting a server without `REGION`/`AVAILABILITY_ZONE` produces a WARN naming the missing variables; starting with them produces no warning and the values appear in the startup banner.
* The same detection exists on the client for its own local placement.
* `region`/`az` metric labels carry `unknown`, never `""`.
* The alert rule fires on a cluster with unset placement and stays silent on a correctly configured one.
* Both dashboards render the placement panel, showing real AZ buckets on a configured cluster and a single `unknown` bucket otherwise.

shaurya703:
Fixes #292

Two changes, both keeping placement optional — you were right that failing hard is too strong, and a single-AZ deployment legitimately sets neither variable.

## 1. Say it out loud, once

`MissingPlacementEnv()` reports which of `REGION` / `AVAILABILITY_ZONE` are unset. The **server** and the **client** each warn at startup naming them, or log the values when set:

```
WARN  topology placement is not configured; this node registers with no
      region/az and cannot classify peers as local or remote
      {"unset": ["REGION", "AVAILABILITY_ZONE"], "clusterName": "default"}
```

Warned separately on the two sides because they are configured independently and either can be the one that is wrong — your UAT case was server-side, but a client missing its own placement silently degrades every per-replica scope to `unknown` in exactly the same way.

**`CLUSTER_NAME` is deliberately excluded.** It has a real default (`"default"`), so including it would fire the warning on every correctly configured single-cluster deployment — which is how warnings get ignored. There's a test pinning that.

## 2. Never emit `az=""`

`LabelOrUnknown` renders an unset value as `"unknown"` — the word `Scope` already uses for the same condition, so a dashboard grouping by `az` and one reading scope agree rather than carrying two spellings of "we do not know".

Applied at the **metric boundary** in `SetActiveSegmentNodes`, not at each caller. An unset AZ arrives there from two directions — a server that registered with no placement, and a quorum entry with no placement recorded for that node — and normalising once means a later caller cannot reintroduce the blank by forgetting to.

This changes one existing assertion: `TestActiveSegmentNodes_SetAndClear` looked its third node up under `AZ: ""`, which is precisely the behaviour being fixed, so it now looks it up where the series actually is. Flagging it rather than letting it look like incidental churn.

## Two parts of the issue I did not implement, and why

**The dashboard panel already exists.** The issue says the client templates have *"no placement panel — nothing groups by `az` at all"*. Both `dashboard_client.json` and `dashboard_client_k8s.json` already carry **"Active Segment Replicas by AZ"** (panel id 111) running exactly the query proposed:

```promql
count by (az) (woodpecker_client_active_segment_node{...})
```

It landed in **#270** (`enhance(metrics): make per-replica and cross-AZ client traffic observable`), and `dashboard-configmaps.yaml` is already in sync with both. So there is nothing to add — and with change (2) above, that panel now shows a single `unknown` bucket instead of a blank one, which is what the issue wants it to show. I'd rather say this than add a duplicate panel.

**The alert rule needs infrastructure that isn't there yet.** No rule files exist in the repo, `tests/docker/monitor/prometheus.yml` has no `rule_files:`, and `kube-prometheus-values.yaml` sets `alertmanager: { enabled: false }`. Adding the rule means introducing alerting to both monitor stacks — a larger decision than this fix, and one I didn't want to make unilaterally inside a bug fix. Happy to take it as a follow-up if you want it.

## Verification

Both changes mutation-checked rather than only run green:

| mutation | result |
|---|---|
| `LabelOrUnknown` returns its argument unchanged | **2 red** — `TestLabelOrUnknown`, `TestActiveSegmentNodes_UnsetAZIsUnknownNeverBlank` |
| `MissingPlacementEnv` returns `nil` | **3 red** — both-unset, region-only, az-only; the "both set is silent" control stays green |

The metric test asserts the **gathered label values**, not the collector's keys, since the exposed label is what a dashboard groups by and that is what was wrong.

`go build ./...`, `go vet`, `gofmt -l` clean. `./common/topology` and `./common/metrics` pass. The dashboard validation and config-map sync tests (`./tests/k8s/monitor`) pass unchanged.

`./common/objectstorage` fails on this machine for want of a reachable MinIO — it does the same on `master`, so it is environmental rather than from this change.

### DIFF
--- a/common/metrics/active_segment_node_test.go
+++ b/common/metrics/active_segment_node_test.go
@@ -17,11 +17,14 @@
 package metrics
 
 import (
+	"sort"
 	"testing"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/testutil"
 	"github.com/stretchr/testify/assert"
+
+	"github.com/zilliztech/woodpecker/common/topology"
 )
 
 func TestActiveSegmentNodes_SetAndClear(t *testing.T) {
@@ -43,7 +46,9 @@ func TestActiveSegmentNodes_SetAndClear(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, len(nodes), count, "one series per node after set")
 	for _, n := range nodes {
-		v := testutil.ToFloat64(WpActiveSegmentNode.WithLabelValues(ns, logId, segId, n.Node, n.AZ))
+		// An unset AZ is rendered "unknown" at the metric boundary (issue #292),
+		// so that is the label the series is filed under.
+		v := testutil.ToFloat64(WpActiveSegmentNode.WithLabelValues(ns, logId, segId, n.Node, topology.LabelOrUnknown(n.AZ)))
 		assert.Equal(t, float64(1), v)
 	}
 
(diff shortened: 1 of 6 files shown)


tinswzy:
Thank you for your contribution.

Issue #290, opened 2026-09-04
enhance(metrics): no reader position visibility — all three frontiers are write-side


tinswzy:
## Summary

The client exposes three progress frontiers, and **all three describe the write side**:

| metric | meaning |
| --- | --- |
| `woodpecker_client_write_frontier_{segment,entry}` | how far the writer has committed (LAC) |
| `woodpecker_client_compaction_frontier_{segment,entry}` | how far compaction has reached |
| `woodpecker_client_truncation_frontier_{segment,entry}` | how far truncation has reached |

The reader side has volume and latency only — `read_requests_total`, `read_entries_total`, `read_latency`, `reader_bytes_read`, `reader_operation_latency` — and **no position**. There is no way to ask "where is this reader" from metrics.

That makes the most common reader question unanswerable on the dashboard: **is the reader behind, or is there simply nothing left to read?** Both look identical — a flat `read_entries_total` and a busy `read_requests_total`.

## What this costs today

Answering that question currently means going outside metrics entirely. A recent investigation had to:

1. read the segment list and each segment's `lastEntryId` out of etcd,
2. decode the embedding application's persisted consume-checkpoint protobuf from etcd,
3. cross-check both against `write_frontier` to conclude that the reader had in fact caught up and the log was simply idle.

Everything needed was already in the client's memory at the time. With `write_frontier − read_frontier` on a panel it would have been one glance.

## Proposed metric

```
woodpecker_client_read_frontier_segment{log_ns, log_id, reader_name}
woodpecker_client_read_frontier_entry  {log_ns, log_id, reader_name}
```

Value = the position of the last entry successfully returned to the caller. Set in `logBatchReaderImpl.ReadNext` (`woodpecker/log/log_reader.go:129`) after a successful read. `logBatchReaderImpl` is currently the only reader implementation, so there is one call site.

### Implementation notes

* **The existing `setMonotonicFrontier` helper cannot be reused as-is.** It keys its guard map by `logNs + "\x00" + logId` (`common/metrics/client_metrics.go:385`), so two readers on one log would fight over the same monotonic guard and clamp each other. The reader frontier needs the reader identity in the key. Keying by `(logNs, logId, readerName)` also gives the right semantics for free: a newly opened reader starting from an earlier position is a *different* series and legitimately starts lower.
[...]

PR #295, opened 2026-09-04
fix(metrics): drop the per-reader label from the reader bytes counter

### LINKED ISSUE #293: fix(metrics): reader_bytes_read carries an unbounded reader_name label that nothing queries
## Summary

`WpLogReaderBytesRead` is a counter labelled with the reader's identity (`common/metrics/client_metrics.go:149-154`):

```go
WpLogReaderBytesRead = prometheus.NewCounterVec(prometheus.CounterOpts{
    Name: "reader_bytes_read",
    Help: "Total bytes read by log readers",
}, []string{"log_ns", "log_id", "reader_name"})
```

`reader_name` carries a unique per-instance id — for example `in01-…-rootcoord-dml_8-r-recovery-1788429671580630447`. **Nothing ever deletes these series**: `grep` finds no `Delete*` or `Reset` call for this metric anywhere. A new series is created for every reader ever opened and is retained for the life of the process.

**And no query uses the label.** Both dashboards aggregate it away:

```promql
sum(rate(woodpecker_client_reader_bytes_read{...}))     # dashboard_client_k8s.json
sum(rate(woodpecker_client_reader_bytes_read{...}))     # dashboard_client.json
```

The only other reference in the repo is a documentation table (`tests/docker/monitor/README.md:114`).

So the label is pure cost: it grows the client's in-memory series map and Prometheus' active series count, with zero realised value.

Observed in the field: a log with `open_log_reader = 2` had exactly two `reader_bytes_read` series, one per reader ever opened — including the one that had already been closed. Readers are recreated on every scanner restart, and during a backoff loop that can be several per minute.

## Why removing the label beats adding cleanup

The obvious alternative is to delete the series in `Close`. Removing the label is better on three counts:

1. **No cleanup code to get wrong**, and no leak left to reason about.
2. **It closes a real blind spot.** A reader that lives *shorter than the scrape interval* (15–30s) currently produces a series that is **never scraped at all**, so the bytes it read are lost from the record entirely. Scanner backoff loops produce exactly this kind of short-lived reader. Aggregated per log, those bytes are counted like any others. Removing the label makes the data **more** complete, not less.
3. **Counter semantics become correct.** A counter without an instance label accumulates permanently, needs no deletion, and does not drop when a reader goes away — so even a bare `sum()` behaves.

## What we lose

The ability to ask "which reader read the most". In practice a log has one scanner reader plus, during recovery, a recovery reader — so the per-reader breakdown was answering a question nobody asks, and no query in the repo asks it.

If a reader-level dimension is wanted later, the right shape is a **low-cardinality** `reader_type` (`recovery` / `normal`), not a unique id.

Per-reader *position* is a separate matter and stays per-reader: `#290` adds `read_frontier_{segment,entry}{log_ns, log_id, reader_name}` as a **gauge**, deleted on close. That is the correct Prometheus shape for an instance dimension. After both changes the split is clean:

| purpose | type | reader identity | on close |
| --- | --- | --- | --- |
| volume / throughput | Counter | no | kept, accumulates forever |
| current position | Gauge | yes | deleted |

## Changes

* Drop `reader_name` from `WpLogReaderBytesRead`'s label set (`client_metrics.go:154`).
* Update the two increment sites: `woodpecker/log/log_reader.go:168` and `:271`.
* Update the label column in `tests/docker/monitor/README.md:114`.

### Migration note

Removing a label changes series identity. Across the deploy the old series go stale and new ones start from zero, so `sum(rate(...))` shows a one-scrape discontinuity. Both existing panels already use `rate`, which handles the transition, so the impact is a single blip — worth mentioning in the changelog so nobody reads it as an incident.

## Dashboard changes

**None required.** Both panels already aggregate over `reader_name` and continue to work unchanged. This is noted explicitly so the absence of a dashboard change is not mistaken for an omission.

## Acceptance criteria

* `woodpecker_client_reader_bytes_read` exposes exactly one series per `(log_ns, log_id)`, regardless of how many readers have been opened.
* After N open/close cycles on one log, the series count stays at one and its value is the sum of all readers' bytes — including readers that lived less than one scrape interval.
* Both dashboard queries return the same values as before for a long-lived reader.
* The README label table matches the new label set.

tinswzy:
Closes #293.

`reader_bytes_read` carried `reader_name`, which holds a unique per-reader id
(`…-r-recovery-1788429671580630447`). Nothing ever deleted those series, so one
accumulated for every reader ever opened and stayed for the life of the process —
readers are rebuilt on every scanner restart, and a backoff loop produces several
a minute. Nothing queried the label either: both dashboards already aggregate it
away with `sum(rate(...)) by (log_id)`.

Removing the label rather than deleting the series on close also closes a gap. A
reader that lives less than one scrape interval produced a series that was never
scraped at all, so the bytes it read were lost from the record entirely — scanner
backoff produces exactly that kind of reader. Aggregated per log they are counted
like any others.

The per-reader dimension belongs on a gauge rather than a counter; #290 adds the
reader's position there and deletes it on close.

## Changes

- `common/metrics/client_metrics.go` — drop `reader_name` from the label set
- `woodpecker/log/log_reader.go` — the two increment sites
- `tests/docker/monitor/README.md` — label column
- `common/metrics/reader_bytes_read_test.go` — new test

## Note for the reviewer

**No dashboard change is needed and none is included** — both panels already
aggregate over `reader_name` and keep working unchanged. Flagging this so the
absence is not read as an omission.

The test's real guard is the assertion on the **label set itself**: re-adding a
per-reader label compiles fine and would silently reintroduce the growth, so
counting series alone would not catch it.

## Release note

Removing a label changes series identity, so `sum(rate(woodpecker_client_reader_bytes_read))`
shows a one-scrape discontinuity across the deploy. Both existing panels use
`rate`, so it is a single blip rather than a reset — worth mentioning in the
changelog so nobody reads it as an incident.

## Verification

- `gofmt` clean, `go vet ./common/metrics/ ./woodpecker/log/` clean, `go build ./...` ok
- `go test ./common/metrics/` ok
- `go test ./woodpecker/log/` ok

### DIFF
--- a/common/metrics/client_metrics.go
+++ b/common/metrics/client_metrics.go
@@ -151,7 +151,7 @@ var (
 		Subsystem: clientRole,
 		Name:      "reader_bytes_read",
 		Help:      "Total bytes read by log readers",
-	}, []string{"log_ns", "log_id", "reader_name"})
+	}, []string{"log_ns", "log_id"})
 	WpLogReaderOperationLatency = prometheus.NewHistogramVec(prometheus.HistogramOpts{
 		Namespace: woodpeckerNamespace,
 		Subsystem: clientRole,
--- a/common/metrics/reader_bytes_read_test.go
+++ b/common/metrics/reader_bytes_read_test.go
@@ -1,0 +1,66 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package metrics
+
+import (
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Readers are short-lived and are recreated constantly — a recovery reader is
+// rebuilt on every scanner restart — so the bytes counter must not carry the
+// reader's identity. Doing so created one permanent series per reader ever
+// opened, and it also lost the bytes of any reader that finished within a
+// single scrape interval. Aggregating per log fixes both.
+func TestReaderBytesRead_AggregatesAcrossReaders(t *testing.T) {
+	WpLogReaderBytesRead.Reset()
+
+	reg := prometheus.NewRegistry()
+	reg.MustRegister(WpLogReaderBytesRead)
+
+	ns, logId := "bucket/root", "42"
+
+	// Three readers over the life of one log, as scanner restarts would produce.
+	for _, n := range []float64{100, 250, 30} {
+		WpLogReaderBytesRead.WithLabelValues(ns, logId).Add(n)
+	}
+
+	count, err := testutil.GatherAndCount(reg, "woodpecker_client_reader_bytes_read")
+	assert.NoError(t, err)
+	assert.Equal(t, 1, count, "one series per log regardless of how many readers ran")
+	assert.Equal(t, float64(380), testutil.ToFloat64(WpLogReaderBytesRead.WithLabelValues(ns, logId)))
+
+	// The label set is the actual regression guard: re-adding a per-reader label
+	// compiles fine and silently reintroduces the unbounded series growth.
+	families, err := reg.Gather()
+	require.NoError(t, err)
+	var labelNames []string
+	for _, f := range families {
+		if f.GetName() != "woodpecker_client_reader_bytes_read" {
+			continue
+		}
+		require.NotEmpty(t, f.GetMetric())
+		for _, l := range f.GetMetric()[0].GetLabel() {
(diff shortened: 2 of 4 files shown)


czs007:

This PR reduces the `WpLogReaderBytesRead` counter's label set from `{log_ns, log_id, reader_name}` to `{log_ns, log_id}`; both reviewers confirmed the definition, all call sites, downstream dashboards, docs, and the new regression test are consistent, and nothing else in the repo depends on the removed label.

**Verified:**

- `common/metrics/client_metrics.go:154` — the `CounterVec` label set is correctly reduced to `{log_ns, log_id}`, registration is untouched, and the labels now match sibling metrics such as `WpClientReadEntriesTotal`.
- `woodpecker/log/log_reader.go:168` and `woodpecker/log/log_reader.go:271` — the only two production increment sites are updated to two-argument `WithLabelValues` in the correct order; a repo-wide grep confirms no third call site that would panic on label-count mismatch.
- `woodpecker/log/log_reader.go:77` — the `readerName` field is still consumed by logging and `GetReaderName`, so dropping it from the metric leaves no dead field or dangling reference.
- `tests/docker/monitor/grafana/templates/dashboard_client.json:328`, `tests/k8s/monitor/grafana/templates/dashboard_client_k8s.json:328`, `tests/k8s/monitor/manifests/dashboard-configmaps.yaml:3797` — all panels use `sum(rate(...)) by (log_id)` with no `reader_name` selector or grouping, so they keep working after the label removal; the only remaining `reader_name` hit is the unrelated protobuf field in `proto/meta.proto:88`.
- `common/metrics/reader_bytes_read_test.go:33-66` — the test asserts a single aggregated series across three readers, the summed value, and the exact label set via `ElementsMatch`, which is a genuine regression guard against re-adding a per-reader label; it calls `Reset()` on the package-global metric and the package has no `t.Parallel`, so there is no cross-test contamination.
- `tests/docker/monitor/README.md:114` — the metric label column is updated, consistent with that table's existing convention of omitting `log_ns`.

The release note's callout about a one-scrape `rate()` discontinuity across deploy is accurate and expected given the change in series identity.

PR #296, opened 2026-09-04
fix(log): marshal the message before taking the writable segment handle

### LINKED ISSUE #288: [bug] Write marshals the payload while holding a segment handle that may already be rolling
## Summary

`GetOrCreateWritableSegmentHandle` returns a **perishable** reference: it takes the log handle lock (`woodpecker/log/log_handle.go:466`), decides whether to roll, and releases the lock before returning. Another writer goroutine can roll that segment immediately afterwards, at which point the handle you are holding is obsolete and `AppendAsync` will reject the append with `ErrSegmentHandleSegmentRolling` (`woodpecker/segment/segment_handle.go:282,302`).

The synchronous `Write` puts a full payload marshal **inside** that window:

```go
// woodpecker/log/internal_log_writer.go
165:  writableSegmentHandle, err := l.logHandle.GetOrCreateWritableSegmentHandle(ctx, ...)  // lock released on return
171:  bytes, err := MarshalMessage(msg)      // <-- holds a possibly-stale handle across this
181:  writableSegmentHandle.AppendAsync(ctx, bytes, callback)                                // rollingState checked here
```

`MarshalMessage` (`woodpecker/log/log_message.go:83`) is `proto.Marshal` over the payload — pure CPU, and **it does not need the handle at all**.

## Both `Write` implementations have it backwards; both `WriteAsync` have it right

| method | order | window contains marshal |
| --- | --- | --- |
| `internal_log_writer.Write` | GetOrCreate(165) → **Marshal(171)** → Append(181) | **yes** |
| `internal_log_writer.WriteAsync` | **Marshal(218)** → GetOrCreate(252) → Append(259) | no |
| `log_writer.Write` | GetOrCreate(242) → **Marshal(248)** → Append(258) | **yes** |
| `log_writer.WriteAsync` | **Marshal(295)** → GetOrCreate(329) → Append(336) | no |

The two `WriteAsync` paths are the existing, correct precedent in the same files. The asymmetry looks accidental rather than intentional.

## Why it matters

Embedding applications call the synchronous `Write`. In an observed Milvus deployment a single WAL message is ~2.7 MiB (`size=2729281` on the insert path), so the marshal is a multi-megabyte allocate-and-copy sitting in the window.

Field numbers from that deployment (rates, so **the window width below is an inference from them, not a direct measurement**):

| quantity | value |
| --- | --- |
| segment rolls | ≈ 0.69 /s (instance-wide, 32 logs) |
| append retries (`ErrSegmentHandleSegmentRolling` surfaced to the caller) | ≈ 0.023 /s |
| rejections per roll | 0.023 / 0.69 ≈ **0.033** |
| per-log append rate | ≈ 2 /s |
| **implied window** | 0.033 / 2 ≈ **~16 ms** |

Every one of those rejections is avoidable: the append had not been enqueued yet, and re-fetching the handle would have returned the new segment immediately. They cost the caller a retry with backoff, and each retried message holds its position in the embedding application's ack ordering until it lands.

With the marshal moved out, the remaining window is a single function call — nanoseconds instead of milliseconds.

## Why the fix is safe

Verified on the current code:

1. **The callback does not capture `bytes`.** In both `Write` implementations `bytes` is referenced only at the `AppendAsync` call and afterwards in the outer scope; the closure defined earlier does not use it. (This is exactly why `WriteAsync` *must* marshal first — its callback *does* reference `bytes` for `WpLogWriterBytesWritten`.)
2. **No fail-fast is lost.** Both `Write` implementations already check writer validity at the top of the function (`isWriterValid` / `sessionLock.IsValid()`), before either operation. Marshalling earlier does not waste work on a dead writer.
3. **`MarshalMessage` has no dependency on the handle** — its only parameter is the message.

The one behavioural difference: when a message is malformed *and* handle acquisition would also fail, the caller now sees the serialization error instead of the handle error. `ValidateMsg` failure is a deterministic caller bug that a retry cannot fix, so reporting it first is arguably better — and it avoids doing rolling work for a message that can never be written.

## Proposed fix

Move the marshal block above `GetOrCreateWritableSegmentHandle` in both synchronous `Write` implementations, matching the `WriteAsync` ordering already in the same files:

* `woodpecker/log/internal_log_writer.go` — `Write` (currently 165 / 171)
* `woodpecker/log/log_writer.go` — `Write` (currently 242 / 248)

No other change.

## Acceptance criteria

* In both `Write` implementations, `MarshalMessage` is called before `GetOrCreateWritableSegmentHandle`, and nothing but the `AppendAsync` call sits between acquiring the handle and using it.
* Regression test: N goroutines writing concurrently to one log while rolling is forced (small `maxSize`, or `SetRollingReady` driven directly), asserting **zero** `ErrSegmentHandleSegmentRolling` results. The test should fail on the current ordering — an artificial delay in place of the marshal makes the race reliable to reproduce.
* No change to `WriteAsync` on either implementation.

tinswzy:
Closes #288.

`GetOrCreateWritableSegmentHandle` returns a **perishable** reference: it takes
the log handle lock, decides whether to roll, and releases the lock before
returning. Another writer goroutine can roll that segment immediately
afterwards, and `AppendAsync` on the stale handle is then rejected with
`ErrSegmentHandleSegmentRolling`.

Both synchronous `Write` implementations did the payload marshal **inside** that
window. `MarshalMessage` is `proto.Marshal` over the payload — pure CPU that does
not need the handle at all — and for a multi-megabyte message it stretches a
window that should be one function call into milliseconds. Field rates from a
deployment writing ~2.7 MiB messages imply a window around 16 ms, and every
rejection it produced was avoidable: the append had not been enqueued, and
re-fetching the handle would have returned the new segment immediately.

Both `WriteAsync` implementations already marshal first — they have to, since
their callback captures the marshalled bytes — so this only aligns the
synchronous paths with the correct ordering already present in the same files.

## Changes

- `woodpecker/log/internal_log_writer.go` — `Write`: marshal moved above the handle acquisition
- `woodpecker/log/log_writer.go` — `Write`: same
- `woodpecker/log/log_writer_test.go` — new ordering test

`WriteAsync` is untouched on both implementations. After this all four paths
share the same order: marshal → acquire handle → `AppendAsync`.

## Why this is safe

Verified on the current code:

1. **The callback does not capture `bytes`** in either synchronous `Write` — it is
   referenced only at the `AppendAsync` call and afterwards in the outer scope.
   (This is precisely why `WriteAsync` *must* marshal first: its callback *does*
   use `bytes`, for `WpLogWriterBytesWritten`.)
2. **No fail-fast is lost** — both functions already check writer validity
   (`isWriterValid` / `sessionLock.IsValid()`) at the top, before either operation.
3. **`MarshalMessage` has no dependency on the handle.**

One behavioural difference: a message that is both malformed *and* racing a failed
handle acquisition now reports the serialization error first. `ValidateMsg` failure
is a deterministic caller bug that no retry can fix, so reporting it first is the
more useful of the two — and it avoids doing rolling work for a message that can
never be written.

## About the test

`TestWrite_MarshalsBeforeAcquiringWritableHandle` pins the ordering down **without
racing anything**: it writes a message that cannot be marshalled (`&WriteMessage{}`,
which `ValidateMsg` rejects) and asserts `GetOrCreateWritableSegmentHandle` was
never called. If the handle is acquired first, the mock records the call even
though the write can never proceed.

I verified it actually catches the regression — reverting only the two source files
to the previous ordering and keeping the test:

```
--- FAIL: TestWrite_MarshalsBeforeAcquiringWritableHandle/internalLogWriter
        Error: Should not have called with given arguments
--- FAIL: TestWrite_MarshalsBeforeAcquiringWritableHandle/logWriter
        Error: Should not have called with given arguments
```

Both subtests fail on the old ordering and pass on the new one. A concurrency-based
test was considered and rejected: it would be inherently flaky, whereas this is
deterministic and runs in milliseconds.

## Verification

- `gofmt` clean, `go vet ./woodpecker/log/` clean, `go build ./...` ok
- `go test ./woodpecker/log/` ok (full package, 56s)

### DIFF
--- a/woodpecker/log/internal_log_writer.go
+++ b/woodpecker/log/internal_log_writer.go
@@ -162,12 +162,6 @@ func (l *internalLogWriterImpl) Write(ctx context.Context, msg *WriteMessage) *W
 		}
 		close(ch)
 	}
-	writableSegmentHandle, err := l.logHandle.GetOrCreateWritableSegmentHandle(ctx, l.onWriterInvalidated)
-	if err != nil {
-		callback(-1, -1, err)
-		metrics.WpLogWriterOperationLatency.WithLabelValues(l.logNs, l.logIdStr, "write", "error").Observe(float64(time.Since(start).Milliseconds()))
-		return <-ch
-	}
 	bytes, err := MarshalMessage(msg)
 	if err != nil {
 		logger.Ctx(ctx).Warn("serialize message failed", zap.String("logName", l.logHandle.GetName()), zap.Int64("logId", l.logHandle.GetId()), zap.Error(err))
@@ -176,6 +170,13 @@ func (l *internalLogWriterImpl) Write(ctx context.Context, msg *WriteMessage) *W
 			LogMessageId: nil,
 			Err:          err,
 		}
+	}
+
+	writableSegmentHandle, err := l.logHandle.GetOrCreateWritableSegmentHandle(ctx, l.onWriterInvalidated)
+	if err != nil {
+		callback(-1, -1, err)
+		metrics.WpLogWriterOperationLatency.WithLabelValues(l.logNs, l.logIdStr, "write", "error").Observe(float64(time.Since(start).Milliseconds()))
+		return <-ch
 	}
 
 	writableSegmentHandle.AppendAsync(ctx, bytes, callback)
--- a/woodpecker/log/log_writer.go
+++ b/woodpecker/log/log_writer.go
@@ -239,12 +239,6 @@ func (l *logWriterImpl) Write(ctx context.Context, msg *WriteMessage) *WriteResu
 		}
 		close(ch)
 	}
-	writableSegmentHandle, err := l.logHandle.GetOrCreateWritableSegmentHandle(ctx, l.onWriterInvalidated)
-	if err != nil {
-		callback(-1, -1, err)
-		metrics.WpLogWriterOperationLatency.WithLabelValues(l.logNs, l.logIdStr, "write", "error").Observe(float64(time.Since(start).Milliseconds()))
-		return <-ch
-	}
 	bytes, err := MarshalMessage(msg)
 	if err != nil {
 		logger.Ctx(ctx).Warn("serialize message failed", zap.String("logName", l.logHandle.GetName()), zap.Int64("logId", l.logHandle.GetId()), zap.Error(err))
@@ -253,6 +247,13 @@ func (l *logWriterImpl) Write(ctx context.Context, msg *WriteMessage) *WriteResu
 			LogMessageId: nil,
 			Err:          err,
 		}
+	}
+
+	writableSegmentHandle, err := l.logHandle.GetOrCreateWritableSegmentHandle(ctx, l.onWriterInvalidated)
+	if err != nil {
+		callback(-1, -1, err)
+		metrics.WpLogWriterOperationLatency.WithLabelValues(l.logNs, l.logIdStr, "write", "error").Observe(float64(time.Since(start).Milliseconds()))
+		return <-ch
 	}
 
 	writableSegmentHandle.AppendAsync(ctx, bytes, callback)
--- a/woodpecker/log/log_writer_test.go
+++ b/woodpecker/log/log_writer_test.go
@@ -2620,3 +2620,52 @@ func TestMonitorSession_CheckAlive_Success(t *testing.T) {
 		t.Fatal("monitorSession did not exit after writerClose")
 	}
 }
+
+// The writable segment handle is perishable: another writer can roll the segment
+// as soon as GetOrCreateWritableSegmentHandle releases the log handle lock, and
+// AppendAsync on a rolled handle is rejected with ErrSegmentHandleSegmentRolling.
+// Marshalling the payload while holding one stretches that window from a function
+// call to a multi-megabyte proto.Marshal, so appends get rejected for no reason.
+//
+// A message that cannot be marshalled pins the ordering down without racing
+// anything: if the handle is acquired first, the mock records the call even
+// though the write can never proceed. Both WriteAsync paths already marshal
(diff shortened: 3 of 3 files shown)


czs007:

This PR moves `MarshalMessage` ahead of `GetOrCreateWritableSegmentHandle` in both synchronous `Write` paths so serialization no longer runs inside the perishable-handle window, and both reviewers confirmed the reorder is correct, behavior-preserving on every error branch, and covered by a deterministic regression test.

Verified:
- `woodpecker/log/internal_log_writer.go:147-164` / `woodpecker/log/log_writer.go:225-241` — the sync-path callback closure captures only `ctx`, `ch`, and the writer, never `bytes`, so marshalling earlier cannot change callback semantics.
- `woodpecker/log/internal_log_writer.go:165-173` / `woodpecker/log/log_writer.go:242-250` — the marshal-failure branch returns directly without invoking the callback; `ch` has no sender or receiver, so no goroutine leak (same branch body as before the change).
- `woodpecker/log/internal_log_writer.go:136` / `woodpecker/log/log_writer.go:214` — writer/session validity is checked before both the marshal and the handle acquire, so fail-fast on a dead writer is preserved.
- `woodpecker/log/log_message.go:83-97` — `MarshalMessage` is `ValidateMsg` + `proto.Marshal` and depends on no handle or lock, so hoisting it is safe.
- `woodpecker/log/log_handle.go:462-467` — the real handle implementation releases its lock on return, so the handle can be rolled by another writer afterward; the perishable window described in the PR is real and the reorder shortens it.
- `woodpecker/log/log_writer_test.go:82-119` — `TestWrite_MarshalsBeforeAcquiringWritableHandle` uses `&WriteMessage{}`, which `ValidateMsg` rejects at `log_message.go:117-119`, and asserts the handle was never acquired; under the old ordering the assertion fails deterministically, so it is a genuine regression test rather than a race-dependent one.

No surviving findings at any severity, and no pre-existing hazards were flagged.

PR #298, opened 2026-09-04
fix(metrics): surface an unset REGION/AVAILABILITY_ZONE instead of blanking it

### LINKED ISSUE #292: [bug] An unset REGION/AVAILABILITY_ZONE silently becomes an empty label instead of failing loudly

tinswzy:
Closes #292.

`topology.GetCurrentRegion` / `GetCurrentAvailabilityZone` fall back to the empty
string, and nothing checked the result. The server passed them straight into its
membership registration (`server/service.go`), from where they reached gossip and
were persisted into every segment's quorum info; clients read them back and
`Scope` degraded every classification to `unknown`. **No log line, no metric and
no health check said placement was missing.** On one cluster the condition
survived **49 days**: `az_scope` was uniformly `unknown`, and live quorum metadata
carried no `region`/`az` at all while `cluster_name` read `default` —
`envOrDefault`'s fallback, and the only visible hint.

The blank was also actively camouflaging: `count by (az) (...)` renders an empty
label as `{}`, which reads like an ordinary aggregation with no breakdown rather
than like missing data. **The dashboard panel that should have exposed this
looked normal.**

## What changed

**1. Startup warns.** Both the logstore node (`NewServerWithConfig`) and the
client (`NewClient` / `NewEmbedClient`) report when they cannot determine their
own placement, naming the environment variables involved. A warning rather than a
startup failure — placement is genuinely optional for a single-AZ deployment —
but silence is not acceptable. The client half matters independently: a client
missing these classifies **every** peer as unknown no matter how well the servers
are configured, so checking only the servers would miss half the failure mode.

**2. `Unknown` instead of blank at the label boundary.** Only one metric carries a
raw `az` label (`active_segment_node`); everything else goes through `Scope`,
which already reported `unknown`.

**3. An alert rule ships with the monitoring manifests**
(`prometheusrule-placement.yaml`, applied by `run_monitor_tests.sh`), held for
10m so a rolling restart does not trip it. The stack already has
`ruleSelectorNilUsesHelmValues: false`, so no wiring beyond the manifest was
needed.

## The one thing I deliberately did not do — please check this in review

**I did not change `GetCurrentRegion`/`GetCurrentAvailabilityZone` to return
`Unknown`.** That is the obvious simplification once `Unknown` exists, and it is
a trap:

```go
func Scope(localRegion, localAZ, peerRegion, peerAZ string) string {
	if localRegion == "" || localAZ == "" || peerRegion == "" || peerAZ == "" {
		return ScopeUnknown
	}
	...
	return ScopeLocal   // all four equal
}
```

With both sides reading `"unknown"`, region and AZ compare **equal** and two
unplaced processes are classified `ScopeLocal` — reporting every byte between
them as node-local traffic. That is strictly worse than reporting it as unknown,
and just as invisible.

So the empty string stays as `Scope`'s "never configured" sentinel, the
conversion happens only where a value becomes a label, `LabelOrUnknown` carries
that constraint in its doc comment, and `TestScope_EmptyIsTheUnsetSentinel`
asserts the trap directly so crossing the boundary fails loudly.

## Note on the issue text

The **Dashboard changes** section of #292 is wrong and I corrected it
[in a comment](https://github.com/zilliztech/woodpecker/issues/292#issuecomment-5539390425):
the client templates already carry a full Cross-AZ / Replica Traffic row,
including the `count by (az)` panel. I had inspected them from a tree that
predated #270. **No dashboard change is needed here** — with (2) in place the
existing panel shows an `unknown` bucket instead of a blank one, which is exactly
the intended signal.

## Verification

- `gofmt` clean, `go vet ./common/... ./woodpecker/... ./server/` clean, `go build ./...` ok
- `go test ./common/topology/ ./common/metrics/` ok
- `go test ./woodpecker/segment/` ok (full package)
- `bash -n run_monitor_tests.sh` ok; the PrometheusRule parses
- `TestActiveSegmentNodes` updated (it asserted the blank), plus a new case for a
  replica that registered with an empty `Az` — different from having no entry at
  all, and the case actually seen in the field

### DIFF
--- a/common/topology/topology_utils.go
+++ b/common/topology/topology_utils.go
@@ -19,12 +19,27 @@ const (
 	ScopeUnknown     = "unknown"
 )
 
+// IsUnknown reports whether a placement value carries no information.
+//
+// It accepts the rendered form as well as the empty string, and that is what
+// makes it total. Rewriting a placement to ScopeUnknown before classifying it is
+// the obvious simplification once LabelOrUnknown exists, and it is a trap: two
+// unplaced processes would compare equal on both region and AZ and be
+// classified ScopeLocal, so every byte between them would be reported as
+// node-local traffic. Recognising the rendered form here means that mistake
+// yields ScopeUnknown instead of a silent lie, and the boundary is held by this
+// function rather than by a comment on LabelOrUnknown telling callers where not
+// to use it.
+func IsUnknown(v string) bool {
+	return v == "" || v == ScopeUnknown
+}
+
 // Scope classifies a peer's placement relative to a local placement. Either
 // side may be unset — REGION/AVAILABILITY_ZONE are optional, and a process
 // without them cannot tell local from remote at all — which yields
 // ScopeUnknown rather than a guess.
 func Scope(localRegion, localAZ, peerRegion, peerAZ string) string {
-	if localRegion == "" || localAZ == "" || peerRegion == "" || peerAZ == "" {
+	if IsUnknown(localRegion) || IsUnknown(localAZ) || IsUnknown(peerRegion) || IsUnknown(peerAZ) {
 		return ScopeUnknown
 	}
 	if localRegion != peerRegion {
--- a/common/topology/topology_utils_test.go
+++ b/common/topology/topology_utils_test.go
@@ -41,6 +41,8 @@ func TestScope(t *testing.T) {
 		{"local az unset", "r1", "", "r1", "a1", ScopeUnknown},
 		{"peer placement unset", "r1", "a1", "", "", ScopeUnknown},
 		{"peer az unset", "r1", "a1", "r1", "", ScopeUnknown},
+		{"local placement already rendered", ScopeUnknown, ScopeUnknown, "r1", "a1", ScopeUnknown},
+		{"peer placement already rendered", "r1", "a1", ScopeUnknown, ScopeUnknown, ScopeUnknown},
 	}
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
@@ -96,4 +98,36 @@ func TestMissingPlacementEnv(t *testing.T) {
 		assert.Empty(t, MissingPlacementEnv())
 		assert.Equal(t, DefaultClusterNameValue, GetCurrentClusterName())
 	})
+}
+
+// IsUnknown decides "never configured", and it has to hold for both forms: the
+// empty string a placement is carried as, and the ScopeUnknown a metric label
+// spells it out as.
+func TestIsUnknown(t *testing.T) {
+	assert.True(t, IsUnknown(""), "the carried form of an unset placement")
+	assert.True(t, IsUnknown(ScopeUnknown), "the rendered form must not be mistaken for a real region")
+	assert.False(t, IsUnknown("us-west-2"))
+	assert.False(t, IsUnknown("us-west-2c"))
+	// Deliberately not tolerated: placement values are compared as given, so a
+	// value differing only in case or padding is a real misconfiguration between
+	// two deployment templates and must not be folded away into "unset".
+	assert.False(t, IsUnknown("Unknown"))
+	assert.False(t, IsUnknown(" "))
+}
+
+// Two unplaced processes must never be reported as co-located. Compared as
+// ordinary strings they match on both region and AZ, so every byte between them
+// would be counted as node-local traffic - a silent lie, and worse than
+// reporting the pair as unknown. Scope excludes unset placements before it
+// compares anything, through IsUnknown, so this holds for the carried form and
+// for a value that was rendered by LabelOrUnknown before reaching it.
+func TestScope_UnknownNeverBecomesLocal(t *testing.T) {
+	assert.Equal(t, ScopeUnknown, Scope("", "", "", ""),
+		"two unplaced processes are unknown to each other")
+	assert.Equal(t, ScopeUnknown, Scope(LabelOrUnknown(""), LabelOrUnknown(""), LabelOrUnknown(""), LabelOrUnknown("")),
+		"a placement rendered before classifying must not compare equal into ScopeLocal")
+	assert.Equal(t, ScopeUnknown, Scope(ScopeUnknown, ScopeUnknown, "r1", "a1"),
+		"a rendered local placement is still an unset one")
+	assert.Equal(t, ScopeUnknown, Scope("r1", "a1", ScopeUnknown, ScopeUnknown),
+		"a rendered peer placement is still an unset one")
 }
--- a/tests/k8s/monitor/manifests/prometheusrule-placement.yaml
+++ b/tests/k8s/monitor/manifests/prometheusrule-placement.yaml
@@ -1,0 +1,38 @@
+# A cluster that cannot determine its own placement should say so without anyone
+# opening a dashboard. An unset REGION/AVAILABILITY_ZONE is not an outage, but it
+# silently disables every cross-AZ traffic metric, and the startup warning that
+# reports it is only seen by whoever is reading that pod's log at the time.
+#
+# Lives in $NAMESPACE alongside the PodMonitors rather than with Prometheus:
+# kube-prometheus-values.yaml sets ruleSelectorNilUsesHelmValues: false and
+# leaves ruleNamespaceSelector at its default, so the operator adopts rules from
+# any namespace - the same cross-namespace discovery the PodMonitors here already
+# rely on. Alertmanager is disabled in this suite, which stops notification but
+# not evaluation: the rule still fires into ALERTS and shows in the Prometheus UI.
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: woodpecker-placement
+  namespace: woodpecker
+  labels:
+    app.kubernetes.io/name: woodpecker
+    app.kubernetes.io/component: client
+spec:
+  groups:
+    - name: woodpecker-placement
+      rules:
+        - alert: WoodpeckerPlacementUnknown
+          # Held for 10m so a rolling restart, during which a node briefly
+          # re-registers, does not trip it.
+          expr: count(woodpecker_client_active_segment_node{az="unknown"}) > 0
+          for: 10m
+          labels:
+            severity: warning
+          annotations:
+            summary: Woodpecker replicas report an unknown availability zone
+            description: >-
+              {{ $value }} active-segment replicas carry az="unknown", so cross-AZ
+              traffic accounting is unavailable for them. Either the logstore
+              nodes or the client are missing REGION / AVAILABILITY_ZONE; the
+              startup log of whichever is misconfigured says which variable, in a
+              "topology placement is not configured" warning.
--- a/tests/k8s/monitor/run_monitor_tests.sh
+++ b/tests/k8s/monitor/run_monitor_tests.sh
@@ -76,6 +76,7 @@ install_monitoring() {
   kubectl apply -f "$SCRIPT_DIR/manifests/podmonitor-server.yaml"
   kubectl apply -f "$SCRIPT_DIR/manifests/podmonitor-client.yaml"
   kubectl apply -f "$SCRIPT_DIR/manifests/dashboard-configmaps.yaml"
+  kubectl apply -f "$SCRIPT_DIR/manifests/prometheusrule-placement.yaml"
 }
 
 start_loadgen() {
--- a/woodpecker/woodpecker_client.go
+++ b/woodpecker/woodpecker_client.go
@@ -103,6 +103,29 @@ type woodpeckerClient struct {
 	closeState atomic.Bool
 }
 
+// reportClientPlacement says once, at startup, whether this client knows where
+// it is.
+//
+// The client reads its own REGION/AVAILABILITY_ZONE to work out how far each
+// replica sits from it. Unset, every peer classifies as ScopeUnknown, so the
+// per-replica traffic metrics stop distinguishing local from cross-region while
+// still looking well-formed. Reported separately from the server: the two are
+// configured independently and either can be the one that is wrong.
+//
+// It lives here rather than inline because a client is built by two independent
+// constructors - NewClient and NewEmbedClient - and an embedded deployment is no
+// better placed to notice the gap from a dashboard than a service one.
+func reportClientPlacement(ctx context.Context) {
+	if missing := topology.MissingPlacementEnv(); len(missing) > 0 {
+		logger.Ctx(ctx).Warn("topology placement is not configured; this client cannot classify replicas as local, cross-az or cross-region",
+			zap.Strings("unset", missing))
+		return
+	}
+	logger.Ctx(ctx).Info("topology placement",
+		zap.String("region", topology.GetCurrentRegion()),
+		zap.String("az", topology.GetCurrentAvailabilityZone()))
+}
+
 func NewClient(ctx context.Context, cfg *config.Configuration, etcdClient *clientv3.Client, managed bool) (Client, error) {
 	// Re-validate the object-storage section at the consumption point (the config may have
 	// been mutated after load); the client builds object-storage keys and RPC rootPath values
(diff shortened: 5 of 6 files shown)


czs007:

Adversarial review of this PR surfaced two medium correctness issues in the new placement alert path plus four low-severity follow-ups, all introduced by this PR.

## Medium

- **`woodpecker/segment/segment_handle.go:654`** — Any quorum without `Replicas` now yields `az="unknown"` for every node and trips the placement alert with no corresponding log. `activeSegmentNodes` defaults each member to `topology.Unknown` and only overrides it from `placements[node]`, which is built solely from `QuorumInfo.Replicas` (`segment_handle.go:610-619`). Two producers never fill `Replicas`: the embed client's `SelectQuorumNodes` (`woodpecker_embed_client.go:358-366`, `Nodes` only) and any segment metadata persisted before `QuorumInfo.replicas` existed (added in #168). Those quorums are stored verbatim via `storeNewSegmentMeta` and read back at handle construction, so `SetActiveSegmentNodes` emits `az="unknown"` regardless of `REGION`/`AVAILABILITY_ZONE` configuration. That matches `prometheusrule-placement.yaml:20`, while `warnIfPlacementUnknown` (`woodpecker_client.go:111-121`) only logs when the client's own env is missing, so the warnings the alert description points to never appear. Failure: embed-mode process, or a service-mode cluster with active segments created before #168, with placement env correctly set everywhere → every active-segment series carries `az="unknown"`, `WoodpeckerPlacementUnknown` fires after 10m on a correctly configured deployment, and its description sends oncall to startup warnings that were never logged. Suggestion: distinguish "no replica entry" from "replica entry with empty `Az`" (e.g. emit a separate sentinel or skip the series when `placements` is empty), and ensure the embed client and pre-#168 metadata paths cannot produce the alerting value. (raised by bigsheeper, czs007)

- **`tests/k8s/monitor/manifests/prometheusrule-placement.yaml:28`** — The alert only detects server-side missing placement, but the annotation claims it covers the client side too. The `az` label comes solely from `placements[node].Az` (`segment_handle.go:655-656`), i.e. the server-registered `QuorumNode.Az`. The client's own `REGION`/`AVAILABILITY_ZONE` only affect `az_scope` via `quorumNodeScopes` (`segment_handle.go:630-641`), which the expr never queries. The description "Either the logstore nodes or the client are missing REGION / AVAILABILITY_ZONE" therefore mismatches the expr. The repo's own monitor testbed is a live instance: `client-loadgen.yaml` has no `env:` section while servers get `zone-a`/`region-local` via operator-injected `topology.env`. Failure: servers have placement, client does not (exactly the shipped `tests/k8s/monitor` testbed) → every `az_scope` label is `unknown` and the client logs "client placement is unknown", yet `WoodpeckerPlacementUnknown` never fires; conversely, when it does fire, its description tells oncall to check client logs that cannot be the cause. Suggestion: either add a second expr (or OR clause) on the `az_scope="unknown"` metrics for the client side, or correct the description to state that this alert covers logstore-node placement only. (raised by bigsheeper, czs007)

## Low

- **`tests/k8s/monitor/manifests/prometheusrule-placement.yaml:19`** — The `for: 10m` rationale is wrong, and the alert will fire for hours or days after an upgrade until old segments roll. The comment says 10m absorbs rolling-restart re-registration, but the series' `az` value is a snapshot taken from persisted `segmentMeta.Metadata.Quorum` at handle construction (`segment_handle.go:115`, `141-146`) and only removed by `ClearActiveSegmentNodes` in `stopWritingUnsafe` (`segment_handle.go:1052`). Node restarts do not rewrite existing series, so 10m neither suppresses nor needs to suppress restart jitter. The real sustained source is metadata written before #168. Failure: cluster upgraded with still-active segments whose persisted `QuorumInfo` predates #168, on low-traffic logs that roll rarely → `WoodpeckerPlacementUnknown` stays firing well past 10m, until each old segment rolls, on a cluster whose placement env is correctly set. Suggestion: size the hold by the maximum natural segment roll period, or change the expr to only match "replica entry present but `Az` empty", and fix the comment to describe the actual transient being suppressed. (raised by bigsheeper, czs007)

- **`tests/k8s/monitor/manifests/prometheusrule-placement.yaml:20`** — The PrometheusRule expr has no offline validation against the registered metric name and label. It is purely string-coupled to `common/metrics/client_metrics.go:227-232`; `run_monitor_tests.sh:79` only `kubectl apply`s it, and neither `monitor_test.go` nor `dashboards_test.go` references `az`, `active_segment_node`, or Prometheus `/api/v1/rules`. If the label is renamed (e.g. `peer_az`) or the metric moves subsystem, all tests and the apply still succeed while `WoodpeckerPlacementUnknown` silently never matches any series. The README (`:97-104`) records that `dashboards_test.go` exists precisely to guard against stale manifests in the same script; this new PrometheusRule is the only artifact applied there without such a gate. Suggestion: add a cluster-free test alongside `dashboards_test.go` that parses the yaml expr and asserts the metric name is among the registered collectors and `az` is in its label list. (raised by czs007)

- **`common/metrics/client_metrics.go:399`** — `LabelOrUnknown` is not pushed down into `SetActiveSegmentNodes`, and the `ActiveSegmentNode` doc comment contradicts the new semantics. The unknown-sentinel substitution happens at the call site in `segment_handle.go:654-656` rather than inside `SetActiveSegmentNodes` (`client_metrics.go:407-411`), so any other caller can still emit an empty `az` label. The comment at `client_metrics.go:399-401` still says the field is "empty when the quorum metadata carries no placement for it", which this PR makes false. Suggestion: apply `LabelOrUnknown` inside `SetActiveSegmentNodes` so the invariant is enforced at the metric boundary, and update the doc comment. (raised by czs007, bigsheeper)

- **`tests/k8s/monitor/manifests/prometheusrule-placement.yaml:20`** — `count()` has no `by()` grouping or namespace restriction. The expr aggregates across every scraped namespace/job into a single alert with no identifying labels. The rule is loaded cluster-wide (`kube-prometheus-values.yaml:16` sets `ruleSelectorNilUsesHelmValues: false`) while the PodMonitor scopes scraping per namespace/instance (`podmonitor-client.yaml:10-16`) and relabels `woodpecker_id`, which the alert discards. Failure: Prometheus scrapes clients from more than one namespace/instance and any one has `az="unknown"` series → a single `WoodpeckerPlacementUnknown` alert fires with no `namespace`/`woodpecker_id`/`log_ns` labels, and oncall cannot identify which tenant or log to look at. Suggestion: group with `by (namespace, woodpecker_id, log_ns)` (or similar) so each affected instance produces its own labelled alert. (raised by czs007, bigsheeper)

tinswzy:
Reworked in d22a429. The complaint was right and the second half of it — keep the carried form as the empty string and spell out `unknown` only for metrics — is the better call; here is what I checked before landing it.

## Why the carried form stays empty

I first went the other way (normalize at the env read so `unknown` is the only form). That is wrong for a reason that only shows up at the edges: the values are not read from the environment on the consuming side. A peer's region and AZ come out of `proto.QuorumInfo` **already persisted in etcd**, written by whatever build was running at the time. The empty string is therefore the compatibility surface, and changing what we publish buys nothing:

* an unconfigured server would start advertising `region=unknown` on gossip and into every segment's quorum info;
* an **old** client, itself configured, would then compare `us-west-2` against `unknown`, and classify that replica `cross_region` instead of `unknown`. I checked the blast radius — `orderedQuorumReadCandidates` only branches on `== ScopeLocal`, so routing is unaffected and it is a metric label only — but it is a regression bought for nothing.

Keeping `""` also means no bytes added to every `QuorumNode` record, and "is this set?" stays a length check rather than a trim plus a case fold.

## What actually changes

The defect was never the empty string; it was that the boundary between the two forms was held up by a comment:

```go
// Do not apply it before Scope. ...
func LabelOrUnknown(v string) string
```

`IsUnknown` now recognises **both** forms, which makes it total:

```go
func IsUnknown(v string) bool { return v == "" || v == Unknown }
```

`Scope` excludes unset placements through it, so labelling a value too early yields `ScopeUnknown` instead of the silent lie the comment was warning about — two unplaced processes comparing equal on region and AZ and being reported as node-local traffic. The test that used to *assert the trap* is inverted:

```go
// before
assert.Equal(t, ScopeLocal, Scope(Unknown, Unknown, Unknown, Unknown),
    "pre-rewriting placement to Unknown makes unplaced processes look co-located — do not do it")

// after
assert.Equal(t, ScopeUnknown, Scope(Unknown, Unknown, Unknown, Unknown),
    "already-labelled placements must not compare equal into ScopeLocal")
```

Verified it bites: with `Scope` reverted to the empty-only guard, `TestScope_UnknownNeverBecomesLocal` fails with `expected: "unknown", actual: "local"`.

Also gone:

* `LabelOrUnknown` → `Label`, named for the boundary it serves (a metric label or a log field) rather than for what it does to a string, and with no warning attached.
* `CurrentPlacement`'s third return value. `configured` was `IsUnknown` spelled out for two callers; both now say `topology.IsUnknown(region) || topology.IsUnknown(az)` directly.

## On lower-casing

Deliberately not done, and it follows from the same decision: placement values are compared as given, so a value differing only in case is a real misconfiguration between two deployment templates. Folding it away would hide exactly the class of problem this PR exists to surface — and it would cost a trim plus an allocation on a comparison that no longer needs one.

For the record on cost: `Scope` is not on the append hot path. `quorumNodeScopes` runs at segment-handle construction and on quorum refresh; `recordReplicaResult` indexes a precomputed slice.

## Verification

`go build ./...`, `gofmt`, `go vet` clean; `go test -race ./common/topology/` and `./woodpecker/segment/` clean; no `LabelOrUnknown` references left in the tree.

tinswzy:
Rebased onto #294 and cut down to what is actually left. The branch is now a single commit on top of `4843ad9`: **114 lines added across 6 files**, down from 212 across 9.

## Dropped, because #294 does it — and in two places does it better

* **`LabelOrUnknown` and the two startup warnings** — the same idea, arrived at independently, down to the function name.
* **The label conversion in `activeSegmentNodes`.** #294 applies it inside `SetActiveSegmentNodes`, at the metric boundary, and that is the better place for exactly the reason its comment gives: no future caller can reintroduce the blank by forgetting. There is one caller today, so the two are behaviourally identical right now — the advantage is structural, and I would rather have it structural.
* **The `quorum_scope_test.go` assertions**, which checked for `unknown` at the wrong layer. `TestActiveSegmentNodes_UnsetAZIsUnknownNeverBlank` asserts the *gathered label value*, which is what a dashboard groups by and what was actually wrong.
* **`CurrentPlacement`** — its only purpose was telling a caller "not configured", which `MissingPlacementEnv` now does better, by naming which variable.

## What is left

**1. The embedded client never reported its placement.** `NewEmbedClient` does not go through `NewClient` — they are independent constructors — so the line added to `NewClient` covers the service path only. The report is extracted into `reportClientPlacement` and called from both; message and fields unchanged from #294.

**2. `IsUnknown` holds the boundary in code rather than in a comment.** `LabelOrUnknown` is documented as a rendering decision, which is right and which nothing enforces. A caller that renders a placement before classifying it makes two unplaced processes compare equal on region *and* AZ, so `Scope` returns `ScopeLocal` and every byte between them is reported as node-local traffic.

```go
func IsUnknown(v string) bool { return v == "" || v == ScopeUnknown }
```

`Scope` asks it instead of testing for `""` in four places, so the mistake yields `ScopeUnknown` instead of a silent lie. Verified it bites — with `Scope` reverted to the empty-only guard:

```
--- FAIL: TestScope_UnknownNeverBecomesLocal
        expected: "unknown"
        actual  : "local"
```

Case and padding are deliberately *not* tolerated (`IsUnknown("Unknown")` is false): a placement differing only in case is a real mismatch between two deployment templates, and folding it away would hide the class of problem this issue is about.

**3. The alert rule.** #294 held it back for want of alerting infrastructure. That is right for the **docker** suite — `prometheus.yml` has no `rule_files:`, so a rule there is inert, and none is added. It does not hold for the **k8s** suite:

* `kube-prometheus-values.yaml` sets `ruleSelectorNilUsesHelmValues: false`, so the operator's rule selector is nil and matches every `PrometheusRule`;
* `ruleNamespaceSelector` is left at its default, so rules are adopted from any namespace — the same cross-namespace discovery the `PodMonitor`s in this suite already depend on, since they live in `$NAMESPACE` while Prometheus lives in `$MON_NS`;
* `alertmanager: { enabled: false }` stops **notification**, not **evaluation** — the rule still fires into `ALERTS` and shows in the Prometheus UI.

The rule's description points at the startup warning #294 added, so the alert and the log line lead to the same place. That reasoning is recorded in the manifest itself rather than only here.

## Verification

`go build ./...`, `gofmt`, `go vet` clean; `go test -race ./common/topology/ ./common/metrics/` clean; the alert manifest parses.

PR #299, opened 2026-09-04
enhance(metrics): expose the reader frontier of each log

### LINKED ISSUE #290: enhance(metrics): no reader position visibility — all three frontiers are write-side

tinswzy:
Closes #290.

The client published three progress frontiers and **all three describe the write
side** — write, compaction, truncation. The reader side had volume and latency but
no position, which left the most common reader question unanswerable from the
dashboard: **is the reader behind, or is there simply nothing left to read?** Both
look identical — a flat `read_entries_total` and a busy `read_requests_total`.

Answering it currently means leaving metrics entirely. A recent investigation had
to read the segment list and each segment's `lastEntryId` out of etcd, decode the
embedding application's persisted consume-checkpoint protobuf, and cross-check
both against `write_frontier` — only to conclude that the reader had caught up and
the log was idle. Everything needed was already in the client's memory.

## Design notes for the reviewer

**Keyed by reader, unlike the write-side frontiers.** A log can carry several
readers at different positions — a recovery reader alongside a normal one. Sharing
one monotonic guard would let the reader further ahead **silently clamp every
update from the one behind**, so the lagging reader would look caught up. Keying by
reader also handles reopening at an earlier position for free: a different reader
is a different series and may legitimately start lower.

`setMonotonicFrontier` keeps its behaviour and gains a keyed variant rather than
being duplicated.

**`Close` drops the guard entry as well as the series.** Dropping only the series
would be invisible until a reader name recurred, at which point the new reader's
genuinely lower position would be clamped to the dead one's and its frontier would
look stuck. `TestClearReadFrontier_DropsSeriesAndGuard` covers exactly this.

Samples already scraped stay in Prometheus, so deleting on close loses no history
— it only stops a dead reader from appearing in instant queries with a frozen
value.

## One deliberate deviation from the issue

The issue suggested a **read lag in entries**:

```promql
write_frontier_entry - on(log_ns, log_id) group_right() read_frontier_entry
```

**That is misleading and I did not ship it.** Entry ids restart at every segment,
so a difference taken across a segment boundary is meaningless — a writer at
segment 10 entry 5 and a reader at segment 8 entry 900 computes as `-895`.

The `Reader Segment Lag` panel uses the **segment** frontiers instead, which stay
comparable across boundaries. The entry-level view is still available on the
`Entry Frontiers` panel, where both curves are plotted together and the reader can
be compared against the writer within a segment.

## Dashboard

* `Segment Frontiers` / `Entry Frontiers` — a fourth series per panel, in both
  client templates
* `Reader Segment Lag` — new panel in the free slot of the Progress Frontiers row
* `manifests/dashboard-configmaps.yaml` regenerated via
  `go test ./tests/k8s/monitor -run TestK8sDashboardConfigMaps_InSync -update`

Panel id **115**, deliberately skipping 112–114: the open PR for #291 claims those,
and duplicate ids across two merges would collide.

Documentation: the frontier family was entirely absent from the monitor README.
The section added covers all four rather than only the new one — documenting the
reader frontier while leaving its three siblings undocumented would have been
worse.

## Verification

- `gofmt` clean, `go vet ./common/metrics/ ./woodpecker/log/` clean, `go build ./...` ok
- `go test ./common/metrics/` ok — four new tests: readers independent, monotonic
  guard, clear drops series *and* guard, and 50 open/close cycles leaving neither
  a series nor a guard entry behind
- `go test ./woodpecker/log/` ok (full package)
- `go test ./tests/k8s/monitor -run TestK8sDashboard` ok

## Related

#293 removes `reader_name` from the `reader_bytes_read` **counter**. After both
land the split is clean: counters aggregate per log and are never deleted; gauges
carry the reader identity and are deleted on close.

### DIFF
--- a/common/metrics/client_metrics.go
+++ b/common/metrics/client_metrics.go
@@ -381,27 +381,15 @@ func UpdateSegmentState(logNs, logId, oldState, newState string) {
 	WpClientSegmentState.WithLabelValues(logNs, logId, newState).Inc()
 }
 
-// The frontier setters below record a position and nothing else: no guard, no
-// shared state, no lock. A position is written where it is reached, by whoever
-// reached it.
+// The frontier setters record a position and nothing else, from wherever it is
+// reached. They rely on convergence rather than ordering: each publisher owns the
+// transition it reports, so the last write is the right one, and the write
+// frontier settles again on the next ack. Metrics are per-process, so two writers
+// export two series that never touch.
 //
-// A monotonic guard used to sit in front of them so a graph line could never
-// step back. It cost a process-wide mutex on the append-ack path and, once the
-// read frontier arrived, on every delivered entry - and it could not deliver
-// what it looked like it was delivering. The segment and entry halves are two
-// separately registered collectors, so Gather walks them one after the other and
-// a publisher can run in between: the pair is never atomic with respect to a
-// scrape, lock or no lock. Nor was the value the guard preferred more truthful.
-// During a roll it admitted (N+1, -1), the seed for a segment with nothing in it
-// yet, over (N, lac), the position actually acknowledged - and then rejected the
-// correction, because the guard had already moved on.
-//
-// What is left is convergence, which is enough here. Within one process the
-// races are narrow and self-healing: the write frontier settles on the next ack,
-// and truncation and compaction are published by the code that owns those
-// transitions, so the last write is the right one. Across processes there is no
-// race at all - metrics are per-process, and two of them export two series that
-// never touch.
+// The two halves are separately registered collectors, so a scrape can land
+// between them and see a position torn across a roll. Entry ids restart at zero
+// in every segment, which is also why lag is measured in segments.
 
 func SetWriteFrontier(logNs, logId string, segmentId, entryId int64) {
 	WpClientWriteFrontierSegment.WithLabelValues(logNs, logId).Set(float64(segmentId))
--- a/common/metrics/client_metrics.go
+++ b/common/metrics/client_metrics.go
@@ -327,6 +327,12 @@ type progressFrontier struct {
 	segmentID   int64
 	entryID     int64
 	initialized bool
+	// The gauge children for this key, resolved on first publish. Holding them
+	// keeps WithLabelValues - which hashes the label values and takes the
+	// vector's own lock - off a path the read frontier walks once per delivered
+	// entry, without having to set the gauges outside the guard.
+	segmentGauge prometheus.Gauge
+	entryGauge   prometheus.Gauge
 }
 
 // RegisterClientMetricsWithRegisterer registers all client-side metrics with the given registerer.
@@ -446,23 +452,27 @@ func setMonotonicFrontier(frontiers map[string]progressFrontier, segmentGauge, e
 
 func setMonotonicFrontierKeyed(frontiers map[string]progressFrontier, segmentGauge, entryGauge *prometheus.GaugeVec, key string, labels []string, segmentId, entryId int64) {
 	frontierMu.Lock()
+	defer frontierMu.Unlock()
+
 	current := frontiers[key]
 	if current.initialized && (segmentId < current.segmentID || (segmentId == current.segmentID && entryId < current.entryID)) {
-		frontierMu.Unlock()
 		return
 	}
-	frontiers[key] = progressFrontier{segmentID: segmentId, entryID: entryId, initialized: true}
-	frontierMu.Unlock()
+	if current.segmentGauge == nil {
+		current.segmentGauge = segmentGauge.WithLabelValues(labels...)
+		current.entryGauge = entryGauge.WithLabelValues(labels...)
+	}
+	current.segmentID, current.entryID, current.initialized = segmentId, entryId, true
+	frontiers[key] = current
 
-	// The gauge writes stay outside the critical section. frontierMu is shared by
-	// every log's frontier in the process, and the read frontier is published once
-	// per delivered entry, so doing a label lookup under it would put every reader
-	// and writer in the process behind one another. Each key has a single
-	// publisher - one reader goroutine, or the ordered append-ack path for a log -
-	// so the guard above still decides the order and these two writes cannot be
-	// overtaken by a competing update to the same key.
-	segmentGauge.WithLabelValues(labels...).Set(float64(segmentId))
-	entryGauge.WithLabelValues(labels...).Set(float64(entryId))
+	// Set under the same lock that decided the order. One log's write frontier
+	// has three publishers - a segment handle becoming writable, the append-ack
+	// path, and the fence path - so setting outside would let a guard that
+	// admitted the newer position see the older Set land last, leaving the gauge
+	// behind the guard and the guard rejecting every correction after that. The
+	// expensive part, resolving the children, already happened once per key.
+	current.segmentGauge.Set(float64(segmentId))
+	current.entryGauge.Set(float64(entryId))
 }
 
 // ActiveSegmentNode is one replica of a log's current writable segment: the
--- a/common/metrics/client_metrics.go
+++ b/common/metrics/client_metrics.go
@@ -95,13 +95,13 @@ var (
 		Namespace: woodpeckerNamespace,
 		Subsystem: clientRole,
 		Name:      "read_frontier_segment",
-		Help:      "Highest segment id delivered to the caller per reader",
+		Help:      "Read position of each reader: the segment it last delivered from, or the one it opened at",
 	}, []string{"log_ns", "log_id", "reader_name"})
 	WpClientReadFrontierEntry = prometheus.NewGaugeVec(prometheus.GaugeOpts{
 		Namespace: woodpeckerNamespace,
 		Subsystem: clientRole,
 		Name:      "read_frontier_entry",
-		Help:      "Highest entry id delivered to the caller in the read frontier segment per reader",
+		Help:      "Read position of each reader within read_frontier_segment: the entry it last delivered, or the one it opened at",
 	}, []string{"log_ns", "log_id", "reader_name"})
 
 	// client append data to log
@@ -425,9 +425,13 @@ func SetReadFrontier(logNs, logId, readerName string, segmentId, entryId int64) 
 // simply stops appearing in instant queries.
 func ClearReadFrontier(logNs, logId, readerName string) {
 	labels := prometheus.Labels{"log_ns": logNs, "log_id": logId, "reader_name": readerName}
+	// Series and guard entry go together, under the same lock: dropping one
+	// without the other leaves either a frozen series with no guard or a guard
+	// with no series, and reader_name is unique per open, so whatever is left
+	// behind is never revisited.
+	frontierMu.Lock()
 	WpClientReadFrontierSegment.Delete(labels)
 	WpClientReadFrontierEntry.Delete(labels)
-	frontierMu.Lock()
 	delete(readFrontiers, readFrontierKey(logNs, logId, readerName))
 	frontierMu.Unlock()
 }
@@ -448,9 +452,17 @@ func setMonotonicFrontierKeyed(frontiers map[string]progressFrontier, segmentGau
 		return
 	}
 	frontiers[key] = progressFrontier{segmentID: segmentId, entryID: entryId, initialized: true}
+	frontierMu.Unlock()
+
+	// The gauge writes stay outside the critical section. frontierMu is shared by
+	// every log's frontier in the process, and the read frontier is published once
+	// per delivered entry, so doing a label lookup under it would put every reader
+	// and writer in the process behind one another. Each key has a single
+	// publisher - one reader goroutine, or the ordered append-ack path for a log -
+	// so the guard above still decides the order and these two writes cannot be
+	// overtaken by a competing update to the same key.
 	segmentGauge.WithLabelValues(labels...).Set(float64(segmentId))
 	entryGauge.WithLabelValues(labels...).Set(float64(entryId))
-	frontierMu.Unlock()
 }
 
 // ActiveSegmentNode is one replica of a log's current writable segment: the
--- a/common/metrics/client_metrics.go
+++ b/common/metrics/client_metrics.go
@@ -29,13 +29,6 @@ import (
 // without registration they silently collect data that goes nowhere.
 var (
 	WpClientRegisterOnce sync.Once
-	frontierMu           sync.Mutex
-	writeFrontiers       = make(map[string]progressFrontier)
-	compactionFrontiers  = make(map[string]progressFrontier)
-	truncationFrontiers  = make(map[string]progressFrontier)
-	// Keyed by reader as well as by log: several readers may follow one log at
-	// different positions, and one must not clamp another.
-	readFrontiers = make(map[string]progressFrontier)
 
 	// Log name-id mapping
 	// WARNING: In large-scale deployments with many logs, the "log_name" label
@@ -323,18 +316,6 @@ var (
 	}, []string{"log_ns", "operation", "status"})
 )
 
-type progressFrontier struct {
-	segmentID   int64
-	entryID     int64
-	initialized bool
-	// The gauge children for this key, resolved on first publish. Holding them
-	// keeps WithLabelValues - which hashes the label values and takes the
-	// vector's own lock - off a path the read frontier walks once per delivered
-	// entry, without having to set the gauges outside the guard.
-	segmentGauge prometheus.Gauge
-	entryGauge   prometheus.Gauge
-}
-
 // RegisterClientMetricsWithRegisterer registers all client-side metrics with the given registerer.
 // Without calling this, metrics still work but are not scraped by any registry.
 func RegisterClientMetricsWithRegisterer(registerer prometheus.Registerer) {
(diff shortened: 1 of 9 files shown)


czs007:

Adversarial review of the reader-frontier metrics and the new `Reader Segment Lag` dashboard panel surfaced one dashboard-breaking PromQL issue plus three smaller observability/concurrency gaps.

## High

**`tests/docker/monitor/grafana/templates/dashboard_client.json:901`** — The new `Reader Segment Lag` expression uses `write_frontier_segment{...} - on(log_ns, log_id) group_right() read_frontier_segment{...}`. `group_right()` is many-to-one, so the left-hand side must be unique per `(log_ns, log_id)`, and it cannot be: `NewLogHandle` calls `seedSegmentFrontierMetrics` unconditionally (`woodpecker/log/log_handle.go:152`) and emits `SetWriteFrontier` whenever the log has any segment meta (`:186-188`), regardless of whether the process ever writes — and `OpenLog` routes readers through the same constructor (`woodpecker/woodpecker_client.go:257`). The series is never deleted (`common/metrics/client_metrics.go:390-392` has only a `Set` path, no `WpClientWriteFrontier*.Delete`), and `woodpecker_id` is a cluster-level static label that does not distinguish processes or pods (`tests/docker/monitor/prometheus.yml:10-30`), while `on(log_ns, log_id)` discards the `instance`/`job`/`pod` labels that would. Failure: with ≥2 processes in one `woodpecker_id` cluster opening the same log (e.g. the writer service plus a standalone consumer — the exact topology this PR motivates) and `woodpecker_id` left at its default `All`, the query returns `found duplicate series for the match group {log_ns=..., log_id=...} on the left hand-side` → the panel errors out with no data in precisely the multi-node scenario it was added for. Suggestion: wrap the left-hand side in `max by (log_ns, log_id)` (add `namespace` to both the aggregation and `on()` in the k8s variants). Write frontier is monotonic, so an active writer's value is ≥ any stale seeded value, and `max` also removes the negative-lag case where only a reader instance matches. The same expression appears in three places and all must change: `tests/docker/monitor/grafana/templates/dashboard_client.json:901`, `tests/k8s/monitor/grafana/templates/dashboard_client_k8s.json:901`, and `tests/k8s/monitor/manifests/dashboard-configmaps.yaml:4370` (regenerate the last with `go test ./tests/k8s/monitor -run TestK8sDashboardConfigMaps_InSync -update`). *(raised by czs007, bigsheeper)*

## Low

**`common/metrics/client_metrics.go:436`** — `setMonotonicFrontierKeyed` performs two `WithLabelValues` calls and two `Set` calls inside the package-global `frontierMu` critical section (`client_metrics.go:30`). This PR puts that lock on the read hot path: `woodpecker/log/log_reader.go:169` (per-entry delivery from a cached batch) and `:273` (first entry of a new batch) call `SetReadFrontier` once per delivered entry. The lock is shared by every log's write/compaction/truncation frontier in the process, notably with the append-ack path at `woodpecker/segment/segment_handle.go:387`, so concurrent readers and writers at high entry rates will see measurable contention. This is a performance observation, not a correctness problem. Suggestion: move the gauge `Set` calls outside the critical section, or shard the lock per log/reader. *(raised by czs007, bigsheeper)*

**`woodpecker/log/log_reader.go:169`** — `SetReadFrontier` has exactly two call sites (`:169` and `:273`), both after an entry has already been delivered to the caller; the starting position obtained by `OpenLogReader` (`woodpecker/log/log_handle.go:729-741`) is never published. Failure: a reader that never delivers its first entry — e.g. it takes the infinite `ErrSegmentNotFound` wait/retry branch at `log_reader.go:177-183`, or a new consumer opens at the tail of an idle log — has no `read_frontier_segment`/`_entry` series at all, so the `Reader Segment Lag` panel (whose `group_right()` right-hand side drops the whole row when the series is missing) omits it entirely → operators cannot distinguish "reader stuck at its start point" from "no reader running", which is half of the question this PR states it wants to answer. Both `ErrSegmentNotFound` (`:177-183`) and `ErrEntryNotFound` (`:223-229`) wait and retry indefinitely without emitting anything, so "stuck since open" is a stable long-lived state. Suggestion: emit one `SetReadFrontier` from `startPoint` in `NewLogBatchReader` so a reader has a position as soon as it opens. *(surfaced during verification)*

**`common/metrics/client_metrics.go:419`** — `ClearReadFrontier` deletes the two gauge children first and only then takes `frontierMu` to remove the guard map entry, and `logBatchReaderImpl` has no mutual exclusion between `Close` (`log_reader.go:288`) and `ReadNext` (`:169`/`:273`). Failure: an application shutting down calls `reader.Close(ctx)` from another goroutine while the read goroutine is delivering an entry, landing between the `Delete` and the `SetReadFrontier` → both gauge children and the `readFrontiers` guard entry are recreated after the reader is closed and nothing will ever clean them up (there is no second `Close`), leaving a frozen-value series on `/metrics` and a permanently leaked map entry. Since `readerName` carries a nanosecond timestamp (`log_handle.go:725-727`), every occurrence under reader churn adds another high-cardinality zombie series to the dashboard — exactly what `TestReadFrontier_NoSeriesLeakAcrossReaderChurn` is meant to prevent. Suggestion: move the `Delete` calls inside the `frontierMu` critical section and keep a closed tombstone in the map, or have the reader guard emission after `Close` with a `sync.Once`/state flag. *(surfaced during verification)*

## Nitpick

**`common/metrics/metrics_registration_test.go:107`** — The client sub-test of `TestRegisterCompactedCleanupAndFrontierMetrics` asserts each frontier metric name individually (`:107-112` covers the six write/compaction/truncation names) but the new `woodpecker_client_read_frontier_segment` / `_entry` were not added. The new `common/metrics/read_frontier_test.go` builds private registries and calls `MustRegister` on the two `GaugeVec`s directly, bypassing `RegisterClientMetricsWithRegisterer`. To be precise, `client_metrics.go:339-340` *is* executed by the existing `TestRegisterClientMetricsWithRegisterer` (duplicate registration would panic), so it is not untested — the actual gap is that if either registration line were dropped or a metric name misspelled, no assertion would fail and the metric would be silently absent from `/metrics`. Suggestion: add `assert.True(t, names["woodpecker_client_read_frontier_segment"])` plus the entry equivalent, and call `SetReadFrontier` once in that sub-test so the series appears in the `Gather` result. *(raised by czs007, bigsheeper)*

tinswzy:
All four are real. I checked each against the code before fixing rather than taking the report at face value; every claim held, including the reasoning chains. Fixed in cd79ef7.

## High — confirmed, and it breaks exactly where the panel is supposed to earn its keep

Verified each link:

* `NewLogHandle` calls `seedSegmentFrontierMetrics` unconditionally, and that emits `SetWriteFrontier` whenever the log has any segment metadata — nothing about it is conditional on this process ever writing.
* `OpenLog` → `openLogUnsafe` → `NewLogHandle` (`woodpecker_client.go:257`), so a reader-only process publishes a write frontier too.
* `WpClientWriteFrontier*` has only a `Set` path — no `Delete` anywhere, unlike the read frontier.

So two processes in one cluster opening the same log produce two `write_frontier_segment` series differing only in `instance`/`job`/`pod`, and `on(log_ns, log_id)` discards all three. With `group_right()` the left side is the "one" side, so that is `found duplicate series for the match group` and an empty panel — in the multi-process topology the panel exists for.

Fixed as suggested, with the label set made explicit:

```promql
max by (woodpecker_id, log_ns, log_id) (woodpecker_client_write_frontier_segment{...})
  - on(woodpecker_id, log_ns, log_id) group_right()
  woodpecker_client_read_frontier_segment{...}
```

k8s adds `namespace` to both. I included `woodpecker_id` in the grouping as well: with `$woodpecker_id` at `All` the selector spans clusters, and `max by (log_ns, log_id)` alone would collapse two clusters into one match group. It is a static target label on every job in the docker suite and derived from the pod label in the k8s one, so it is present on both sides.

All three files changed and `dashboard-configmaps.yaml` regenerated with `-update`; the panel description now records why the aggregation is there.

## Low (no series until the first delivery) — confirmed

Both call sites did run after delivery, and there was no emission at open. A reader is now given its opening position in `NewLogBatchReader`. The gauge Help text changed with it: the value is the reader's position — last delivered, or the one it opened at — rather than strictly "delivered".

`TestLogReader_PublishesItsOpeningPosition` covers it and fails against the previous behaviour.

## Low (Close racing a delivery) — confirmed, including that nothing else orders them

`logBatchReaderImpl` has no mutex at all, so `Close` and `ReadNext` are only ordered by the application. The reader now keeps a `frontierClosed` flag under its own mutex, so a delivery that lands after the clear is refused rather than re-creating the series and its guard entry. `ClearReadFrontier` also drops both under one lock now instead of deleting the children first and taking the lock afterwards.

`TestLogReader_RetiredFrontierIsNotRepublished` covers it and fails against the previous behaviour.

## Low (lock on the read hot path) — confirmed

Both `WithLabelValues` calls and both `Set`s were inside `frontierMu`, and this PR is what put that lock on a per-delivered-entry path. The gauge writes now happen outside the critical section, which is left holding only the map lookup and store. Safe because each key has a single publisher — one reader goroutine, or the ordered append-ack path for a log — so the guard still decides the order; the reasoning is written at the code.

## Nitpick — confirmed, and your correction to it is right

The two names were genuinely absent from the assertions, and your note that they were nonetheless *executed* by `TestRegisterClientMetricsWithRegisterer` (a duplicate registration would panic) is accurate — so this closes a gap against a future dropped `MustRegister` or misspelling rather than fixing something untested today. Added, with `SetReadFrontier` called in that sub-test so the series is actually gathered.

## Verification

`go test -race ./common/metrics/ ./woodpecker/log/` clean, `go test ./tests/k8s/monitor -run TestK8sDashboard` clean, `gofmt`/`go vet`/`go build ./...` clean.

czs007:

Adversarial review of this round surfaced one new high-severity issue introduced by the incremental changes: the opening-position publish in `NewLogBatchReader` writes the `Latest` sentinel into the monotonic frontier guard, so tail readers never report a real read frontier.

## High

- **`woodpecker/log/log_reader.go:161`** — `NewLogBatchReader` unconditionally calls `publishReadFrontier(from.SegmentId, from.EntryId)`. When a reader is opened with `LatestLogMessageID()` (`log_message.go:63-66` returns `SegmentId = EntryId = math.MaxInt64`), `shouldMoveToTruncationPoint` at `log_handle.go:1017` only rewrites `Earliest`, so the `Latest` sentinel reaches the reader unchanged via `log_handle.go:741` (and `log_reader.go:361` relies on exactly that sentinel to detect a tail read). The guard in `client_metrics.go:447` is then seeded with `segmentID = MaxInt64, initialized = true`, and every subsequent `publishReadFrontier` from `:210` / `:314` with the real position is silently rejected by the `segmentId < current.segmentID` check at `client_metrics.go:443`. Failure: open a tail reader with `OpenLogReader(ctx, &LatestLogMessageID(), name)` (the path used by `tests/integration/e2e_client_read_write_test.go`, `tests/benchmark/client_test.go`, `tests/docker/chaos/basic_test.go`) and let `ReadNext` deliver entries → `woodpecker_client_read_frontier_segment` / `_entry` stay pinned at ~9.223e18 for the reader's whole lifetime, and the Reader Segment Lag panel (`max(write_frontier_segment) - read_frontier_segment`) shows ~-9.22e18 until `Close` deletes the series, making it impossible to tell whether the reader is behind. This is the opposite of the PR's stated goal of publishing the opened-at position to distinguish "stuck" from "no reader". The new `TestLogReader_PublishesItsOpeningPosition` (`log_reader_frontier_test.go:44-75`) only covers a concrete position `(4, 12)` and does not exercise the sentinel. Suggestion: skip the opening-position publish when `from.SegmentId == LatestLogMessageID().SegmentId`, and instead publish once from `handleTailRead` after it resolves `pendingReadSegmentId` / `pendingReadEntryId` to a real position; add a test that opens with `Latest` and asserts the gauges never contain `MaxInt64`. (raised by bigsheeper, czs007)

czs007:

Adversarial review found one high-severity correctness regression, one gap in test coverage, and one minor metrics ordering issue, all introduced by this PR.

## High

- **`woodpecker/log/log_reader.go:165`** — A reader opened with `LatestLogMessageID()` seeds the monotonic frontier guard with the `math.MaxInt64` sentinel, so every subsequent real read position is rejected and the read frontier is stuck for the reader's lifetime. Opening a tail reader via `OpenLogReader(ctx, &LatestLogMessageID(), name)` and then delivering entries through `ReadNext` → `woodpecker_client_read_frontier_segment/entry` stays pinned at ~9.22e18, all later `publishReadFrontierMetric` calls are rejected at `client_metrics.go:443`, and the Reader Segment Lag panel shows ~-9.22e18 until `Close` removes the series. `NewLogBatchReader` publishes `from` unconditionally, and `adjustPendingReadPointIfTruncated` / `shouldMoveToTruncationPoint` (`log_handle.go:1016-1021`) only rewrite the Earliest sentinel, so Latest passes through unchanged (`log_handle.go:741`). `handleTailRead` later resolves the real `pendingReadSegmentId` but does not publish, and the publishes at `log_reader.go:214` / `:318` all carry a segment id smaller than `MaxInt64`. This is the exact "is the reader lagging or is the log idle" scenario the metric was added to answer. **Suggestion:** skip the opening-position publish when `from.SegmentId == LatestLogMessageID().SegmentId` and publish once `handleTailRead` has resolved the real position, or normalize the `MaxInt64` sentinel in `NewLogBatchReader` before seeding the guard.
  *This was raised in the previous round as a must-fix. The author has not responded to it in the PR comments. The two commits since then (`0be813b31a`, a pure rename of `publishReadFrontier` to `publishReadFrontierMetric`, and `1abfc98b73`, dashboard unit/description changes only) do not touch this logic, so the issue remains unaddressed.* (raised by xiaocai2333, bigsheeper)

## Medium

- **`woodpecker/log/log_reader_frontier_test.go:40`** — No test opens a reader through `NewLogBatchReader` with the `LatestLogMessageID()` sentinel and asserts on the frontier gauges, so the regression above is invisible to CI. Opening with Latest and delivering a real entry via `ReadNext` → the gauge pinned at ~9.22e18 is not caught by any case in `make unit-test`. `TestLogReader_PublishesItsOpeningPosition` only uses the concrete position `{SegmentId: 4, EntryId: 12}` and asserts `4` / `12`. The six cases in `log_reader_test.go` that set `pendingReadSegmentId` to the Latest sentinel construct `logBatchReaderImpl` literals directly, bypass `NewLogBatchReader`, and contain no frontier assertions. `common/metrics/read_frontier_test.go` uses only small integers. **Suggestion:** add a case that opens via `NewLogBatchReader` with `LatestLogMessageID()`, asserts `read_frontier_segment/entry` is not `float64(math.MaxInt64)`, and after the tail read resolves and one entry is delivered, asserts the gauges equal that entry's `(SegId, EntryId)`. (raised by bigsheeper)

## Low

- **`common/metrics/client_metrics.go:457`** — Moving the `segmentGauge` / `entryGauge` `Set` calls outside `frontierMu` relies on the comment's "single publisher per key" assumption, which holds for the read frontier (serialized by the per-reader `frontierMetricMu`) but not for the write frontier. During a segment roll, goroutine A constructs the new segment handle N+1 and calls `SetWriteFrontier` from `segment_handle.go:135` while the old handle N publishes `(N, lac)` from the ack path at `:387`; both pass the guard but A's `Set` runs first → `woodpecker_client_write_frontier_segment` stays at N while the guard holds N+1, and if the log then goes idle the Reader Segment Lag panel under-reports by one segment until the next append ack. There is a third write-side publisher on the fence path at `segment_handle.go:1895`. Before this change both `Set` calls were inside the lock, so guard order matched gauge order. **Suggestion:** keep the write-frontier `Set` calls inside the lock (they fire once per ack, not once per entry like the read side), or update the comment to state that the write side has multiple publishers and briefly reordered gauge writes are accepted. (surfaced during verification)

tinswzy:
All three are real, and the High is a defect I introduced in the previous round. Fixed in `76bda54`.

First, the process point, which is fair: the High was raised twice and I did not answer it. The two commits in between (`0be813b`, `1abfc98`) came from a separate conversation about naming and dashboard units and never touched this path — I had not looked back at the review. Owning that rather than explaining it away.

## High — confirmed, and it hits most readers

Verified each link:

* `LatestLogMessageID()` is `{SegmentId: math.MaxInt64, EntryId: math.MaxInt64}`.
* `shouldMoveToTruncationPoint` rewrites only the **Earliest** sentinel, so Latest reaches `NewLogBatchReader` unchanged.
* My opening-position publish sent it into the guard, and every real position afterwards failed `segmentId < current.segmentID`.

Reproduced exactly, with the guard restored to its previous behaviour:

```
--- FAIL: TestLogReader_OpeningAtLatestDoesNotPoisonTheFrontier
        expected: 0    actual: 1     the Latest sentinel is not a position and must not create a series
        expected: 9    actual: 9.223372036854776e+18
        expected: 41   actual: 9.223372036854776e+18
```

The fix takes both halves of your suggestion:

1. `publishReadFrontierMetric` refuses the sentinel — at the **one choke point** rather than at each caller, so a future call site cannot reintroduce it.
2. The tail read publishes the position it resolves the sentinel *into*. `handleTailRead` sets `pendingReadSegmentId` / `pendingReadEntryId` on every path it can — the tail of the last segment, the start of a segment that does not exist yet, or `0/0` when the log has none — so a tail reader still gets a frontier as soon as it has one, which is what the opening-position publish was for. On the two error paths that leave the sentinel in place, the publish is a no-op by (1).

## Medium — confirmed, test added

`TestLogReader_OpeningAtLatestDoesNotPoisonTheFrontier` opens through `NewLogBatchReader` with `LatestLogMessageID()`, asserts no series is created for the sentinel, then publishes the resolved position and asserts the gauges hold **9 / 41** rather than `MaxInt64` — and that a further delivery still advances rather than being rejected as a step back. It fails against the previous commit, as above.

## Low — confirmed, and the "single publisher" claim was mine and wrong

Checked the write side: three paths publish a write frontier for one log — `segment_handle.go:135` (a segment handle becoming writable), `:387` (the append-ack path), `:1895` (the fence path). "Single publisher per key" holds for a reader and not for a log, so the comment I wrote to justify setting outside the lock was false.

I took the first of your two suggestions rather than the second: documenting reordered gauge writes as acceptable would not have been accurate. Because the guard is monotonic, the divergence is not brief — once the guard holds the newer position it **rejects every correction**, so the gauge stays behind until something publishes past it, and an idle log stays wrong indefinitely.

The `Set` calls are back under the guard's lock. The reason they were moved out — a `WithLabelValues` on a path the read frontier walks once per delivered entry — is handled instead by resolving the two gauge children **once per key** and holding them in the guard entry. The critical section is now a map lookup, a comparison, and two atomic stores; the label hashing and the vector's own lock are gone from it entirely. That answers the previous round's hot-path finding without giving up atomicity.

## Verification

`go test -race ./common/metrics/ ./woodpecker/log/` clean; `gofmt`, `go vet`, `go build ./...` clean.

czs007:

Verified:
- `woodpecker/log/log_reader.go:711` — `publishReadFrontierMetric` returns early when `segmentId` equals the `LatestLogMessageID` sentinel (`MaxInt64`), so the sentinel can never seed the monotonic frontier guard; centralizing the check here also prevents any future call site from reintroducing the pin.
- `woodpecker/log/log_reader.go:177` — opening a reader at Latest publishes `from.SegmentId == MaxInt64`, which is now a no-op, so no `read_frontier` series is created with a poisoned value; `from.SegmentId` was already dereferenced before this change, so the new publish adds no nil risk.
- `woodpecker/log/log_reader.go:384` — tail resolution in `handleTailRead` runs first and every non-error branch rewrites `pendingReadSegmentId` to a concrete value (0, `lastSegmentId`, or the resolved segment) before the frontier is published; the only branch that leaves the sentinel in place is the hard-error return, where the publish is correctly skipped.
- `woodpecker/log/log_reader.go:225` and `:330` — the `ReadNext` delivery paths publish `SegId` from real entries, so a tail reader's first published frontier is always a genuine position that subsequent deliveries can advance past.
- `woodpecker/log/log_reader_frontier_test.go:116` — `TestLogReader_OpeningAtLatestDoesNotPoisonTheFrontier` asserts no series is created at open-at-Latest, the resolved tail position is published, and later deliveries advance rather than being rejected as a step back, which is exactly the failure mode reported in the previous round.
- `tests/docker/monitor/grafana/templates/dashboard_client.json` — the remaining increment (`unit: short → none`, `decimals: 0`, panel descriptions) is Grafana JSON only, with no correctness surface.

The previously raised high-severity item (Latest-opened reader writing the `MaxInt64` sentinel into the monotonic guard and permanently pinning the read frontier) was addressed in the increment since `0be813b` and is confirmed resolved at the source with regression coverage. No new findings survived verification in this round.

Note: verification was read-only; the reviewers did not execute the test suite.

PR #300, opened 2026-09-04
fix(meta): stop one provider mutex from serializing every log's metadata writes

### LINKED ISSUE #289: enhance(meta): a provider-wide mutex serializes segment metadata writes across every log on the client

tinswzy:
Closes #289.

`metadataProviderEtcd` embedded a `sync.Mutex`, and **one provider instance is
handed to every log handle a client hosts** (`woodpecker_client.go:120` creates
it, `:257` passes the same one to each). Nine acquisitions across eight methods
took it — all writes, all holding it **across the etcd round-trip**. Metadata
writes for thirty-two logs queued behind one another even though `clientv3` is
concurrency-safe and multiplexes over a single HTTP/2 connection.

etcd was healthy on the deployment this came from, so the lock had headroom. The
problem is blast radius, not a live incident: one slow etcd write stalls
metadata writes for *every* log on that client, which turns a per-request
slowdown into a client-wide stall in exactly the regime where a WAL can least
absorb it.

## What the mutex protected

**`closed`, and nothing else.** It is the only field the struct assigns after
construction — `e.<field> =` across both files returns exactly one hit,
`e.closed = true` in `Close()`. Everything else is set once in the constructor or
is a `sync.Map`.

Every write under it is already guarded by the compare it needs:

| method | its own guard |
| --- | --- |
| `StoreSegmentMetadata` | txn: parent log exists AND segment absent |
| `UpdateSegmentMetadata` | txn: `ModRevision(segment) == segmentMeta.Revision` |
| `DeleteSegmentMetadata` | txn: `CreateRevision(segment) > 0` |
| `DeleteLogMetadata` | txn: `ModRevision(logKey) == rev` read before parking |
| `CreateLog` | txn: `Value(idgen) == currentID` |
| `ClearMeta` | delete-then-reseed; idgen reseed is create-if-absent |

Those compares are not optional and not incidental. **A second client process is
excluded by nothing in this process**, so etcd — not a mutex here — is what makes
concurrent metadata writes correct. `ClearMeta`'s own comment says so outright:
it explains that it must not read-then-write because that "would race a
concurrent CreateLog". The code was already written as though the lock were not
there.

## The change

`StoreSegmentMetadata`, `UpdateSegmentMetadata`, `DeleteSegmentMetadata` and
`DeleteLogMetadata` now take **no lock at all**.

The embedded mutex becomes two named fields, so the provider stops being a
`sync.Locker` and each remaining acquisition has to be written out and justified:

* **`stateMu`** — `closed`. Held by `Close` and `CreateReaderTempInfo`, never
  across a round-trip. Unchanged behaviour.
* **`instanceMu`** — `CreateLog` and `ClearMeta`, which write the instance-wide
  keys every log shares. Unchanged behaviour; they simply no longer share a lock
  with the per-log write path, so `Close` and reader open no longer queue behind
  a log creation.

For `CreateLog` this is **liveness, not correctness** — its transaction is
CAS-guarded and `createLogUnsafe` retries. But that budget is five attempts, and
concurrent creation of more than five logs in one process would burn it for no
reason.

## One property given up deliberately

`DeleteLogMetadata` parks a log's segments before deleting them. A segment stored
between the park read and the delete is now deleted **without a parked copy**.

The provider-wide lock closed that window within one process — but only within
one process, since another client parks and deletes without consulting this one,
so it was never a system-level guarantee. What is lost is one archived copy of a
segment created while its own log was being deleted; the delete itself stays
atomic and CAS-guarded. Not worth a lock, and not worth a striping scheme either
— which is where I started before concluding there was nothing to stripe.

Both comments that claimed the lock (`DeleteLogMetadata`'s and
`parkLogInBatches`'s "we already hold `e.Lock()`") are corrected rather than left
to mislead.

## Tests

**`testSegmentMetadataWritesTakeNoProviderLock`** holds **both** remaining
mutexes and requires every per-log write to complete anyway. The struct has
exactly two, so this is an exhaustive statement rather than a spot check:
re-adding a shared lock to any of the four methods, *whichever* mutex it reached
for, deadlocks the test.

Verified it actually bites — temporarily putting `stateMu` back on
`StoreSegmentMetadata`:

```
--- FAIL: TestAll/test_segment_metadata_writes_take_no_provider_lock (20.57s)
    StoreSegmentMetadata blocked while a provider mutex was held
```

**`testConcurrentSegmentMetadataWritesAcrossLogs`** drives one provider with
eight logs writing at once, asserting outcomes rather than timings: every segment
lands in the state its own goroutine wrote, none applied to a neighbour's log.
Its real job is giving `-race` the concurrency it needs to see this path.

**`TestSegmentMetadataWriteScaling`** (`tests/benchmark`, manual, skipped like
its neighbours) sweeps 1→32 logs and reports writes/sec per row — the
acceptance criterion's "throughput scaling with N rather than flattening". Flat
means something is serializing again.

## Observability

`WpEtcdMetaOperationLatency` measures this already: `startTime` is taken **before**
the lock in every one of these methods, so queueing has always been inside the
histogram. (The issue's "nothing about it is visible today" was wrong; I have
posted a correction there along with two other errors in its call-site table.)
`startTime` stays where it is, so both sides of this change are comparable on a
metric that already ships. No new metric.

## Verification

- `gofmt` clean, `go vet ./meta/ ./tests/benchmark/` clean, `go build ./...` ok
- `go test ./meta/ -race` ok — full package, including the two new subtests
- full CI unit set (`go list ./... | grep -v tests/`) with `-race -short` ok


czs007:

This PR removes the embedded provider mutex from the etcd metadata provider, replacing it with two narrow locks (`stateMu` for `closed` / reader-session state, `instanceMu` for `CreateLog` / `ClearMeta`) and leaving the segment write paths lock-free, with correctness carried by the existing etcd transaction compares.

**Verified:**

- `meta/metadata_etcd.go:130-152` — every struct field other than `closed` is assigned once at construction or is a `sync.Map`, and the only write to `closed` is in `Close()` at `:1761`, so the removed lock was protecting nothing else.
- `meta/metadata_etcd.go:897-905`, `:958-964`, `:1106-1112` — `StoreSegmentMetadata`, `UpdateSegmentMetadata`, and `DeleteSegmentMetadata` each run as a single etcd transaction guarded by a CreateRevision / ModRevision compare, with no read-modify-write window and no shared in-process mutable state.
- `meta/metadata_etcd.go:1429-1443` and `:1756-1779` — `CreateReaderTempInfo` and `Close` still gate on `closed` and register / drain `readerTempSessions` under the same `stateMu`, so a session registered after drain self-reclaims and no live keepalive goes untracked.
- `meta/metadata_etcd.go:394-395` and `:466-483` — `CreateLog` reads the idgen value and executes the consuming CAS transaction under `instanceMu`, so in-process callers no longer CAS-conflict with each other, while cross-process conflicts are still caught by the idgen CAS and returned as retryable.
- `meta/metadata_etcd.go:2472-2474` and `:2488-2517` — `DeleteLogMetadata` still deletes via a CAS on the log key ModRevision; a segment stored between the park snapshot and the delete is removed atomically with no orphan, and the missing parked copy has no reader since `ListParkedLogIds` (`:670-672`) skips the segments prefix.
- `meta/` grep for `.Lock()` — no remaining bare `e.Lock()` calls, and `stateMu` and `instanceMu` are never nested in production code, so removing the embedded `sync.Mutex` breaks no caller and introduces no lock-order hazard.

The new tests in `metadata_etcd_locking_test.go` hold both locks while requiring the four unlocked methods to complete, which is an exhaustive check that none of them reacquires a shared lock, and the cross-log concurrent write test covers the lock-free paths under the race detector with no data races observed.

One note, non-blocking: the two new tests do not call `Close()` on the provider. This matches the style of the other subtests in `TestAll` and is not something this PR introduced.

PR #301, opened 2026-09-04
fix(client): stop FastSuccess from cancelling a replica that is still answering

### LINKED ISSUE #287: [bug] FastSuccess cancels the stream of a replica that has not answered, turning its normal completion into an error path
## Summary

**Scope: normal quorum write path. No data is lost — see [What does not break](#what-does-not-break).**

The design is that once an append reaches its ack quorum (`Aq`), the caller is told it succeeded and the remaining replica is **left alone to finish quietly**. `receivedAckCallback` already implements that ending: when the slow replica's result finally arrives, `op.fastCalled` is set, so it logs at Debug and returns (`woodpecker/segment/append_op.go:240-244`).

That ending no longer happens. `FastSuccess` closes **every** entry in `op.resultChannels` (`append_op.go:385`), and in service mode `RemoteResultChannel.Close()` cancels the stream context (`common/channel/remote_result_channel.go:163`). So the still-pending replica's `Recv()` is aborted; `ReadResult` gets `rpc error: code = Canceled`, logs a WARN at `remote_result_channel.go:134`, and only then returns into the Debug-level guard.

**A replica that was completing perfectly normally is forced down an error path and produces a warning, once per entry.**

With `Es=3, Aq=2` there is normally exactly one such replica per entry — measured at 0.77 warnings per entry on the observed cluster.

## When this broke

Two commits, seven months apart:

| commit | date | what it did |
| --- | --- | --- |
| `d6f4a5e` *fix chan leak risk by enforcing proper async chan lifecycle in client* | 2025-06-04 | added the `ch.Close(ctx)` loop to `FastSuccess`. Result channels were plain Go channels then — `Close` was local cleanup and cancelled nothing. |
| `61c5efa` *feature: add support service mode* | **2026-01-08** | rewrote `RemoteResultChannel` from a callback holder (`sendFunc`/`readFunc`/`closeFunc`) into a direct `grpc.ServerStreamingClient` holder with `ctx`/`cancel`, giving `Close()` the new meaning **"cancel the peer RPC"**. |

```go
// 61c5efa^ — Close() only invoked a caller-supplied callback
if !r.closed {
    r.closed = true
    if r.closeFunc != nil { r.closeFunc(ctx, r.identifier) }
    logger.Ctx(ctx).Debug("closed remote result channel", ...)
}

// 61c5efa — Close() now aborts the stream
if !r.closed {
    r.closed = true
    if r.cancel != nil { r.cancel() }        // <-- new
    if r.ch != nil { r.ch.CloseSend() }      // <-- new
}
```

`61c5efa` did not revisit the `FastSuccess` call site. The cancellation is semantic drift, not a decision.

## Symptom

`ReadResult` cannot tell a self-inflicted cancel from a real stream failure, so it warns on both. Over 24h on one instance:

| log line | count (24h) |
| --- | --- |
| `failed to read result from remote channel` (`error="rpc error: code = Canceled …"`) | **4,879,009** |
| genuine failure paths (`read chan timeout` + `HandleAppendRequestFailure` + `append fail after retry`) | **295** |

99.994% of these warnings describe a replica that was finishing normally. At a measured 579 bytes per line (WARN attaches a zap stacktrace) that is **≈2.8 GB/day for a single instance**.

## What does not break

**Data is not lost.** The server buffers the entry before the client can cancel: `Server.AddEntry` → `logStore.AddEntry` → `segmentProcessor.AddEntry` → `LocalFileWriter.WriteDataAsync` → `currentBuffer.WriteEntryWithNotify(entryId, data, resultCh)`, and **the request context is not retained** — the flush runs on the writer's own goroutine under the segment sync policy. Cancelling the result stream only ends the server goroutine that was waiting to report the ack.

Consistent with that, the logstore side is quiet: 0 WARN/ERROR on the pool nodes over a 10-minute window while the client was cancelling continuously.

**Pre-quorum retry is unaffected.** A replica that fails *before* the op reaches `Aq` still goes through `HandleAppendRequestFailure` and its per-node retry (`AppendRequestRetryOp`, bounded by `SegmentAppend.MaxRetries`). This issue only concerns what happens to a replica that is still in flight *after* quorum is already satisfied.

## The batch path already gets this right

`batch_append_op.go` gives each entry a `LocalResultChannel` as an in-process sink (`:122`) and keeps the gRPC stream's lifetime with the drain goroutine (`:130-146`):

```go
// The drain goroutine below owns the stream's lifetime: streamCtx is cancelled
// when the drain exits (normally or on give-up), so a stalled peer can't leave
// the client demux goroutine + gRPC stream + server handler hanging.
```

`FastSuccess` closing a `LocalResultChannel` is harmless — it closes a Go channel and touches no RPC. **The defect exists only on the single-entry path**, where `RemoteResultChannel` directly owns the stream and is handed to the op.

## Proposed fix

Give the stream's lifetime to the per-replica goroutine, mirroring the batch path:

1. `receivedAckCallback`: add `defer resultChan.Close(ctx)` at the top. It is already bounded by the 30s budget at `:234`, so a hung peer cannot leak the stream.
2. `FastSuccess`: drop the `ch.Close(ctx)` loop (`:385`). Keep `SendResult`, which still matters for the local-channel case.
3. `FastFail`: **unchanged.** Abandoning the entry and cancelling every replica is correct there.

`Close()` is idempotent (`if !r.closed`, `remote_result_channel.go:155`), so a concurrent `FastFail` and the deferred close cannot conflict.

After this, a replica that answers late reads its **real** result, returns normally, and is retired by the existing Debug-level guard — which is what the code was always meant to do.

### Out of scope

`recordReplicaResult` sits *after* the `fastCalled` guard (`:273`), so a post-quorum replica's outcome is not counted in `woodpecker_client_replica_append_total` — measured at 2.01 outcomes per entry where `Es=3` would predict ~3.0. **That is the guard's own behaviour, not a consequence of the cancellation**, and removing the cancel will not change it. Whether the per-replica metric should count outcomes that arrive after quorum is a separate question; filed separately if we want it.

## Acceptance criteria

* `failed to read result from remote channel` no longer fires on ordinary quorum success; the late replica is retired through the Debug path at `append_op.go:241`.
* With a peer that never answers, the replica goroutine still exits within the existing 30s budget and closes its own channel — no stream or goroutine leak.
* `FastFail` behaviour is unchanged: all replicas are still cancelled when the entry is abandoned.
* Regression test: N concurrent appends against a quorum where one replica is deliberately delayed past quorum; assert that replica's `ReadResult` returns its real result (not `Canceled`) and that no WARN is emitted on that path.

tinswzy:
Closes #287.

`FastSuccess` closed **every** entry in `op.resultChannels`, and in service mode
`RemoteResultChannel.Close()` cancels the stream context. So the replica outside
the ack quorum — with `Es=3, Aq=2` there is normally exactly one — had its `Recv()`
aborted while it was answering perfectly normally, and its completion came back as
`rpc error: code = Canceled`.

**4,879,009 WARN lines in 24h on one instance (~2.8 GB), against 295 from the
genuine failure paths.** 99.994% of them describe a replica that was finishing
fine.

No data was ever at risk: the server buffers the entry before the client can
cancel and does not retain the request context, so cancelling the result stream
only ends the goroutine that was waiting to report the ack. The logstore side is
silent throughout.

**This is drift, not a decision.** `d6f4a5e` (2025-06-04) added the `Close` loop
when result channels were plain Go channels and `Close` was local cleanup.
`61c5efa` (2026-01-08) rewrote `RemoteResultChannel` around a gRPC stream and gave
`Close` the new meaning *"cancel the peer RPC"* — without revisiting this call site.

## The change

The channel belongs to the goroutine reading it. `receivedAckCallback` closes it
on every exit path, **after** the replica has actually answered.

`FastFail` is unchanged and still closes early, deliberately: there the entry is
abandoned, so every waiter should stop at once rather than sit out its budget.
`Close` is idempotent, so the early close and the goroutine's own close cannot
conflict.

`FastSuccess` keeps its `SendResult` — only the `Close` goes.

The channel is retired **as soon as the replica's answer is in**, not only on the
way out: the stream is dead weight the moment the ack arrives, and the bookkeeping
that follows takes the segment handle's lock. The deferred call stays as the
catch-all for every other exit path.

### 1. The read budget was decorative

`receivedAckCallback` wraps the read in a 30s context. **`ReadResult` ignores it**
once a stream is attached:

```go
	resultResponse, readErr := ch.Recv()   // ctx unused on this path
```

`Recv` observes only the stream's own context, which is `context.WithCancel` with
no deadline, and `newConnection` sets no keepalive. So `FastSuccess`'s cancel was
**the only thing that ever unblocked a replica that accepted an entry and then went
silent**. Removing it without a real bound would have leaked the goroutine and the
stream for as long as the connection survived.

`ReadResult` now enforces the caller's deadline by cancelling the stream when it
expires — the same device `AppendEntry` already uses to bound the first response,
not a new mechanism.

This is a standalone bug, but it cannot be separated from the fix: the fix is
unsafe without it. It has its own tests and this section is the record.

### 2. Retries reused the channel object

`sendWriteRequest` reused the slot when the type matched, asserted in four tests as
*"retry idempotency"*. It carries none — the entry id is what makes an append
idempotent, server-side. What reuse actually did:

* every attempt opens its **own** stream, context and cancel;
* `InitResponseStream` **overwrites all three without cancelling the previous ones**,
  so the earlier attempt's cancel is dropped and can never be called;
* it also **does not clear `r.result`**, so a recorded result can be returned to the
  next attempt's read instead of its new stream's.

Splitting by where the failure came from:

| retry trigger | attempt 1's channel | reuse vs fresh |
| --- | --- | --- |
| `GetLogStoreClient` fails | never touched | identical |
| `AddEntry` fails / first `Recv` errors / first-response timeout / first response is a failure | `AppendEntry` cancelled the stream itself and never called `InitResponseStream`, so `ch`/`cancel` are still nil | identical |
| server's second frame reports failure; connection drops | stream handed over, `cancel` never called | reuse orphans that cancel permanently |
| **server accepts, then never sends the ack** | stream handed over and live | **reuse breaks the retry** |

That last row is the decisive one, and only after this PR's other changes:

```
G1(attempt 1): ReadResult exhausts its budget
G1:            HandleAppendRequestFailure  → submits the retry   (synchronous, in-body)
                       G2(attempt 2): sendWriteRequest reuses the same channel
                       G2:            AppendEntry → InitResponseStream(stream2, cancel2)
G1:            returns → deferred Close() → r.cancel() is now cancel2
                       ← attempt 2's brand-new stream is killed by attempt 1's cleanup
```

`maxRetries: 3`, and the interleaving can repeat every round. Budget gone ⇒ the
entry `FastFail`s ⇒ `minRemoveId` fast-fails every queued entry behind it in the
segment.

**Why this is new**: today the only `Close` is in `FastSuccess`/`FastFail`, which hold
`op.mu`, as does `AppendRequestRetryOp.Execute` — they are mutually exclusive. The
deferred close added here runs **outside** `op.mu`, so reuse becomes unsafe the
moment the goroutine owns the channel.

I considered keeping reuse with a compare-and-close (`op.mu` + "close only if the
slot still points at me") and rejected it: `AppendOp.Execute` holds `op.mu` across
the whole parallel fan-out including `wg.Wait()`, so every ack's cleanup would
queue behind the send path — and it would still leave the orphaned cancel and the
stale `r.result`. A fresh channel per attempt costs one small object and loses
nothing, because nothing worth carrying survives an attempt.

## Tests

Every changed and added test was confirmed to **fail against the previous code**
before being kept:

```
--- FAIL: TestAppendOp_FastSuccess_DoesNotCancelInFlightReplicaStreams
        replica 0: FastSuccess must not close the channel
        replica 0: FastSuccess must not cancel the stream
--- FAIL: TestAppendOp_receivedAckCallback_ClosesItsOwnChannel/after_a_normal_ack
--- FAIL: TestAppendOp_receivedAckCallback_ClosesItsOwnChannel/after_a_send_error
--- FAIL: TestAppendOp_Retry_UsesAFreshResultChannelPerAttempt
--- FAIL: TestRemoteResultChannel_ReadResult_ViaStream_HonoursCallerDeadline
        ReadResult ignored the caller's deadline: a silent peer blocks it indefinitely
```

New:

* `TestAppendOp_FastSuccess_DoesNotCancelInFlightReplicaStreams` — three replicas
  holding streams that have not answered; after `FastSuccess` none may be closed
  **or cancelled**. This is the defect itself.
* `TestAppendOp_FastFail_CancelsInFlightReplicaStreams` — the deliberate contrast.
  It passes on the old code too: it guards behaviour this PR must *not* change.
* `TestAppendOp_receivedAckCallback_ClosesItsOwnChannel` — the other half of the
  exchange, on both the normal-ack and the early-return path.
* `TestRemoteResultChannel_ReadResult_ViaStream_HonoursCallerDeadline` — a silent
  peer must not block the read forever. Fails cleanly rather than hanging.
* `TestRemoteResultChannel_ReadResult_ViaStream_DeadlineNotReached` — the budget
  fires only when exhausted; an in-budget ack is returned and its stream left alone.

Rewritten, with the reason recorded in each: the four `RetryIdempotency` /
`ChannelReuse` tests now assert a fresh channel per attempt, and one of them adds
the property that matters — closing an earlier attempt's channel must leave the
current one alone. `TestAppendOp_Retry_ChannelKeepsTheOpIdentifier` now actually
checks the identifier it was named for (`GetIdentifier()` is on the interface; the
old comment said it was unreachable).

`TestAppendOp_FastSuccess`, `_Idempotent` and `_ConcurrentFastCalls` had their
close assertions inverted; the last now asserts the close state **agrees with
whichever of FastFail/FastSuccess won the CAS**, which is stronger than what it
checked before.

### A test-hygiene problem this surfaced

Several append tests observed "the ack was processed" through a mock call made
from **inside** `receivedAckCallback` (`SendAppendSuccessCallbacks`,
`HandleAppendRequestFailure`) and returned there — leaving the goroutine still
writing to the result channel while testify's `AssertExpectations` cleanup
reflects over that same object **without holding its lock**. `-race` reports it.

Retiring the channel before those calls fixes most of it, because the callback
then really is a join point. The remainder needed `waitAckGoroutinesDone`: in the
quorum tests a *failing* replica's `HandleAppendRequestFailure` is mocked but
never awaited, so nothing ever joined that goroutine. Two consecutive
`go test -race ./woodpecker/segment/` runs are clean.

## Out of scope

`recordReplicaResult` sits after the `fastCalled` guard, so a post-quorum replica's
outcome is still not counted in `woodpecker_client_replica_append_total` (measured
at 2.01 outcomes per entry where `Es=3` predicts ~3.0). That is the guard's own
behaviour and removing the cancel does not change it, as the issue notes.

The `ReadResult` WARN stays at WARN. After this change it fires on genuine stream
failures and on `FastFail` — rare, and downgrading it would hide the real ones.

## Verification

- `gofmt` clean, `go vet ./common/channel/ ./woodpecker/segment/ ./woodpecker/client/` clean, `go build ./...` ok
- `go test -race ./woodpecker/segment/` ok, twice in a row
- `go test -race ./common/channel/ ./woodpecker/client/ ./woodpecker/log/` ok

One unrelated local failure: `TestNewEmbedClientFromConfig_MinioObjectStorageError`
fails the same way on `master` in a clean worktree (427s, `https://.blob./` → EOF).
It is environment-dependent and green on CI.

### DIFF
--- a/common/channel/local_result_channel.go
+++ b/common/channel/local_result_channel.go
@@ -126,6 +126,15 @@ func (l *LocalResultChannel) Close(ctx context.Context) error {
 	return nil
 }
 
+// String renders the channel without exposing its fields to reflection; see the
+// comment on RemoteResultChannel.String for why a mutex-guarded type has to
+// control how it is printed.
+func (l *LocalResultChannel) String() string {
+	l.mu.RLock()
+	defer l.mu.RUnlock()
+	return fmt.Sprintf("LocalResultChannel{identifier:%s closed:%v}", l.identifier, l.closed)
+}
+
 func (l *LocalResultChannel) IsClosed() bool {
 	l.mu.RLock()
 	defer l.mu.RUnlock()
--- a/common/channel/local_result_channel.go
+++ b/common/channel/local_result_channel.go
@@ -92,7 +92,7 @@ func (l *LocalResultChannel) ReadResult(ctx context.Context) (*AppendResult, err
 		)
 		return r, nil
 	case <-ctx.Done():
-		return nil, ctx.Err()
+		return nil, readBudgetExhausted(ctx.Err())
 	}
 }
 
--- a/common/channel/local_result_channel_test.go
+++ b/common/channel/local_result_channel_test.go
@@ -680,3 +680,19 @@ func TestLocalResultChannel_TryReadResult(t *testing.T) {
 	assert.False(t, ok)
 	assert.Nil(t, r)
 }
+
+// TestLocalResultChannel_StringIsWhatFmtUses mirrors the remote case: the batch
+// path hands these to a mocked AppendEntries, so they are formatted by code this
+// package does not control while the drain goroutine is closing them.
+func TestLocalResultChannel_StringIsWhatFmtUses(t *testing.T) {
+	resultChannel := NewLocalResultChannel("test-local-stringer")
+
+	rendered := fmt.Sprintf("%v", resultChannel)
+	assert.Equal(t, resultChannel.String(), rendered, "fmt must render via Stringer, not by reflecting over the fields")
+	assert.Contains(t, rendered, "test-local-stringer")
+	assert.Contains(t, rendered, "closed:false")
+	assert.NotContains(t, rendered, "mu:", "reflected field output would mean fmt is reading the fields directly")
+
+	assert.NoError(t, resultChannel.Close(context.Background()))
+	assert.Contains(t, fmt.Sprintf("%v", resultChannel), "closed:true")
+}
--- a/common/channel/local_result_channel_test.go
+++ b/common/channel/local_result_channel_test.go
@@ -97,7 +97,7 @@ func TestLocalResultChannel_ReadTimeout(t *testing.T) {
 	// Try to read from empty channel with timeout
 	_, err := rc.ReadResult(ctx)
 	assert.Error(t, err)
-	assert.Equal(t, context.DeadlineExceeded, err)
+	assert.ErrorIs(t, err, context.DeadlineExceeded)
 }
 
 func TestLocalResultChannel_SendTimeout(t *testing.T) {
@@ -304,7 +304,7 @@ func TestLocalResultChannel_ReadResultTimeout(t *testing.T) {
 	duration := time.Since(start)
 
 	assert.Error(t, err)
-	assert.Equal(t, context.DeadlineExceeded, err)
+	assert.ErrorIs(t, err, context.DeadlineExceeded)
 	// Should timeout around 100ms, not much longer
 	assert.True(t, duration < 200*time.Millisecond, "ReadResult took too long: %v", duration)
 }
@@ -415,7 +415,7 @@ func TestLocalResultChannel_ConcurrentReads(t *testing.T) {
 
 	for _, err := range errors {
 		// The other reader should get a context timeout (channel empty, no more data)
-		assert.Equal(t, context.DeadlineExceeded, err, "non-winning reader should timeout")
+		assert.ErrorIs(t, err, context.DeadlineExceeded, "non-winning reader should timeout")
 	}
 }
 
@@ -567,7 +567,7 @@ func TestLocalResultChannel_MultipleReadersOneMessage(t *testing.T) {
 
 	for _, err := range errors {
 		// The other readers should get a context timeout (channel empty, no more data)
-		assert.Equal(t, context.DeadlineExceeded, err, "non-winning reader should timeout")
+		assert.ErrorIs(t, err, context.DeadlineExceeded, "non-winning reader should timeout")
 	}
 }
 
--- a/common/channel/local_result_channel_test.go
+++ b/common/channel/local_result_channel_test.go
@@ -97,7 +97,14 @@ func TestLocalResultChannel_ReadTimeout(t *testing.T) {
 	// Try to read from empty channel with timeout
 	_, err := rc.ReadResult(ctx)
 	assert.Error(t, err)
-	assert.ErrorIs(t, err, context.DeadlineExceeded)
+	// Both classifications the append paths perform must match. This is the channel
+	// the batch drain reads, and a bare DeadlineExceeded is not classifiable
+	// downstream: the entry would take a terminal failure and spend none of its
+	// retries on what is only a timeout.
+	assert.True(t, werr.IsRetryableErr(err),
+		"a timed-out read must be retryable, not terminal")
+	assert.ErrorIs(t, err, context.DeadlineExceeded,
+		"the append path's read-timeout branch tests for this")
 }
 
 func TestLocalResultChannel_SendTimeout(t *testing.T) {
--- a/common/channel/remote_result_channel.go
+++ b/common/channel/remote_result_channel.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"go.uber.org/zap"
@@ -136,13 +137,30 @@ func (r *RemoteResultChannel) ReadResult(ctx context.Context) (*AppendResult, er
 	// response. Without this ctx is decorative on this path: a peer that accepts
 	// the entry and then goes silent blocks the read for as long as the
 	// connection survives.
+	budgetExpired := &atomic.Bool{}
 	if deadline, ok := ctx.Deadline(); ok && cancel != nil {
-		budgetTimer := time.AfterFunc(time.Until(deadline), cancel)
+		budgetTimer := time.AfterFunc(time.Until(deadline), func() {
+			budgetExpired.Store(true)
+			cancel()
+		})
 		defer budgetTimer.Stop()
 	}
 
 	resultResponse, readErr := ch.Recv()
 	if readErr != nil {
+		if budgetExpired.Load() {
+			// The read ran out of its own budget. Recv reports that as a bare gRPC
+			// status, which nothing downstream can classify: errors.Is against
+			// context.DeadlineExceeded does not match it, and werr.IsRetryableErr
+			// extracts a woodpeckerError and so returns false - so the caller would
+			// treat a timeout as a terminal failure and spend none of its retries
+			// on it. Report the timeout as what it is, and keep the transport
+			// detail in the log.
+			logger.Ctx(ctx).Warn("append result read exceeded its budget",
+				zap.String("identifier", r.identifier),
+				zap.Error(readErr))
+			return nil, werr.ErrAppendOpTimeout.WithCauseErr(context.DeadlineExceeded)
+		}
 		logger.Ctx(ctx).Warn("failed to read result from remote channel",
 			zap.String("identifier", r.identifier),
 			zap.Error(readErr))
--- a/common/channel/remote_result_channel.go
+++ b/common/channel/remote_result_channel.go
@@ -127,7 +127,19 @@ func (r *RemoteResultChannel) ReadResult(ctx context.Context) (*AppendResult, er
 	// If gRPC stream is available, read from it
 	r.mu.RLock()
 	ch := r.ch
+	cancel := r.cancel
 	r.mu.RUnlock()
+
+	// Recv observes only the stream's own context, which carries no deadline, so
+	// the caller's budget has to be enforced by cancelling that stream when it
+	// runs out - the same device AppendEntry already uses to bound the first
+	// response. Without this ctx is decorative on this path: a peer that accepts
+	// the entry and then goes silent blocks the read for as long as the
+	// connection survives.
+	if deadline, ok := ctx.Deadline(); ok && cancel != nil {
+		budgetTimer := time.AfterFunc(time.Until(deadline), cancel)
+		defer budgetTimer.Stop()
+	}
 
 	resultResponse, readErr := ch.Recv()
 	if readErr != nil {
--- a/common/channel/remote_result_channel.go
+++ b/common/channel/remote_result_channel.go
@@ -207,6 +207,21 @@ func (r *RemoteResultChannel) Close(ctx context.Context) error {
 	return nil
 }
 
+// String renders the channel without exposing its fields to reflection.
+//
+// This type is mutable and mutex-guarded, so anything that renders it by walking
+// its fields - fmt's default struct formatting, and through it testify's
+// argument diffing - reads `closed` and `ch` with no lock and races a concurrent
+// Close. A result channel is handed to LogStoreClient.AppendEntry and is
+// therefore held, and printed, by code this package does not control, so
+// controlling the rendering here is what makes that safe.
+func (r *RemoteResultChannel) String() string {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	return fmt.Sprintf("RemoteResultChannel{identifier:%s closed:%v streamed:%v}",
+		r.identifier, r.closed, r.ch != nil)
+}
+
 func (r *RemoteResultChannel) IsClosed() bool {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
--- a/common/channel/remote_result_channel.go
+++ b/common/channel/remote_result_channel.go
@@ -108,7 +108,7 @@ func (r *RemoteResultChannel) ReadResult(ctx context.Context) (*AppendResult, er
 		for {
 			select {
 			case <-ctx.Done():
-				return nil, ctx.Err()
+				return nil, readBudgetExhausted(ctx.Err())
 			case <-ticker.C:
 				r.mu.RLock()
 				if r.closed {
@@ -159,7 +159,7 @@ func (r *RemoteResultChannel) ReadResult(ctx context.Context) (*AppendResult, er
 			logger.Ctx(ctx).Warn("append result read exceeded its budget",
 				zap.String("identifier", r.identifier),
 				zap.Error(readErr))
-			return nil, werr.ErrAppendOpTimeout.WithCauseErr(context.DeadlineExceeded)
+			return nil, readBudgetExhausted(context.DeadlineExceeded)
 		}
 		logger.Ctx(ctx).Warn("failed to read result from remote channel",
 			zap.String("identifier", r.identifier),
--- a/common/channel/remote_result_channel_test.go
+++ b/common/channel/remote_result_channel_test.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc/metadata"
 
 	"github.com/zilliztech/woodpecker/proto"
@@ -424,6 +425,75 @@ func TestRemoteResultChannel_ReadResult_ViaStream_Synced(t *testing.T) {
 	assert.NotNil(t, result)
 	assert.Equal(t, int64(100), result.SyncedId)
 	assert.NoError(t, result.Err)
+}
+
+// TestRemoteResultChannel_ReadResult_ViaStream_HonoursCallerDeadline pins the
+// bound on the async ack read.
+//
+// Recv observes only the stream's own context, which carries no deadline, so
+// before this the ctx passed here was decorative on the stream path: a peer that
+// accepted the entry and then went silent blocked the read until the connection
+// died. That went unnoticed because FastSuccess used to cancel the stream at
+// quorum — which is exactly the cancellation being removed.
+func TestRemoteResultChannel_ReadResult_ViaStream_HonoursCallerDeadline(t *testing.T) {
+	resultChannel := NewRemoteResultChannel("test-read-stream-budget")
+	streamCtx, streamCancel := context.WithCancel(context.Background())
+	defer streamCancel()
+
+	// A replica that accepted the entry and never sends its durability ack.
+	stream := &mockStreamClient{
+		recvFunc: func() (*proto.AddEntryResponse, error) {
+			<-streamCtx.Done()
+			return nil, streamCtx.Err()
+		},
+	}
+	resultChannel.InitResponseStream(stream, streamCtx, streamCancel)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
+	defer cancel()
+
+	type readOutcome struct {
+		result *AppendResult
+		err    error
+	}
+	done := make(chan readOutcome, 1)
+	go func() {
+		result, err := resultChannel.ReadResult(ctx)
+		done <- readOutcome{result: result, err: err}
+	}()
+
+	select {
+	case outcome := <-done:
+		assert.Error(t, outcome.err, "a read that ran out of budget must report it")
+		assert.Nil(t, outcome.result)
+	case <-time.After(10 * time.Second):
+		t.Fatal("ReadResult ignored the caller's deadline: a silent peer blocks it indefinitely")
+	}
+}
+
+// TestRemoteResultChannel_ReadResult_ViaStream_DeadlineNotReached checks the
+// budget only fires when it is actually exhausted: an ack that arrives in time
+// is returned, and the stream it came on is left alone.
+func TestRemoteResultChannel_ReadResult_ViaStream_DeadlineNotReached(t *testing.T) {
+	resultChannel := NewRemoteResultChannel("test-read-stream-in-budget")
+	streamCtx, streamCancel := context.WithCancel(context.Background())
+	defer streamCancel()
+
+	stream := &mockStreamClient{
+		recvFunc: func() (*proto.AddEntryResponse, error) {
+			return &proto.AddEntryResponse{EntryId: 42, State: proto.AddEntryState_Synced}, nil
+		},
+	}
+	resultChannel.InitResponseStream(stream, streamCtx, streamCancel)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	result, err := resultChannel.ReadResult(ctx)
+	assert.NoError(t, err)
+	require.NotNil(t, result)
+	assert.Equal(t, int64(42), result.SyncedId)
+	assert.NoError(t, streamCtx.Err(), "an in-budget read must not cancel the stream")
 }
 
 func TestRemoteResultChannel_ReadResult_ViaStream_Failed(t *testing.T) {
--- a/common/channel/remote_result_channel_test.go
+++ b/common/channel/remote_result_channel_test.go
@@ -26,6 +26,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"google.golang.org/grpc/metadata"
 
+	"github.com/zilliztech/woodpecker/common/werr"
 	"github.com/zilliztech/woodpecker/proto"
 )
 
@@ -464,8 +465,17 @@ func TestRemoteResultChannel_ReadResult_ViaStream_HonoursCallerDeadline(t *testi
 
 	select {
 	case outcome := <-done:
-		assert.Error(t, outcome.err, "a read that ran out of budget must report it")
+		require.Error(t, outcome.err, "a read that ran out of budget must report it")
 		assert.Nil(t, outcome.result)
+
+		// Recv reports the cancellation as a bare gRPC status, which nothing
+		// downstream can classify: the caller would record a terminal failure and
+		// spend none of its retries on what is only a timeout. Both of the
+		// classifications the append path actually performs must match.
+		assert.True(t, werr.IsRetryableErr(outcome.err),
+			"a timed-out read must be retryable, not terminal")
+		assert.ErrorIs(t, outcome.err, context.DeadlineExceeded,
+			"the append path's read-timeout branch tests for this")
 	case <-time.After(10 * time.Second):
 		t.Fatal("ReadResult ignored the caller's deadline: a silent peer blocks it indefinitely")
 	}
--- a/common/channel/remote_result_channel_test.go
+++ b/common/channel/remote_result_channel_test.go
@@ -156,7 +156,7 @@ func TestRemoteResultChannel_ReadResult_NotInitialized(t *testing.T) {
 	result, err := channel.ReadResult(ctx)
 	assert.Error(t, err)
 	assert.Nil(t, result)
-	assert.Equal(t, context.DeadlineExceeded, err)
+	assert.ErrorIs(t, err, context.DeadlineExceeded)
 }
 
 func TestRemoteResultChannel_ReadResult_Closed(t *testing.T) {
@@ -272,7 +272,7 @@ func TestRemoteResultChannel_ReadResult_ContextCancellation(t *testing.T) {
 	result, err := channel.ReadResult(ctx)
 	assert.Error(t, err)
 	assert.Nil(t, result)
-	assert.Equal(t, context.DeadlineExceeded, err)
+	assert.ErrorIs(t, err, context.DeadlineExceeded)
 }
 
 // Test the workflow that simulates real usage in append operations
--- a/common/channel/remote_result_channel_test.go
+++ b/common/channel/remote_result_channel_test.go
@@ -272,7 +272,12 @@ func TestRemoteResultChannel_ReadResult_ContextCancellation(t *testing.T) {
 	result, err := channel.ReadResult(ctx)
 	assert.Error(t, err)
 	assert.Nil(t, result)
-	assert.ErrorIs(t, err, context.DeadlineExceeded)
+	// No stream is installed, so this is the polling exit rather than the stream
+	// one; it must classify the same way.
+	assert.True(t, werr.IsRetryableErr(err),
+		"a timed-out read must be retryable, not terminal")
+	assert.ErrorIs(t, err, context.DeadlineExceeded,
+		"the append path's read-timeout branch tests for this")
 }
 
 // Test the workflow that simulates real usage in append operations
--- a/common/channel/remote_result_channel_test.go
+++ b/common/channel/remote_result_channel_test.go
@@ -19,6 +19,7 @@ package channel
 import (
 	"context"
 	"errors"
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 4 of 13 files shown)


czs007:

## Review summary

Adversarial review found four correctness/logging defects introduced by this PR's change of result-channel ownership: the new early `retireResultChannel()` makes `FastSuccess`/`FastFail` write into already-closed channels on every successful append, leaves batch-path local channels with no closer, and turns read-budget expiry into a non-retryable bare gRPC error.

## Critical

**`woodpecker/segment/append_op.go:422` — `FastSuccess` unconditionally `SendResult`s into channels already closed by `receivedAckCallback`, producing a WARN on every successful append.**
This PR moves channel ownership to the reading goroutine: `receivedAckCallback` calls `retireResultChannel()` at `append_op.go:282`, before `SetAndCount` at `:319`. The ordering is deterministic, not racy — for replica B's `SetAndCount` to return `>= Aq`, replica A's `:319` must already have run, and A's close at `:282` precedes it. The quorum replica then continues synchronously in the same goroutine through `:323 SendAppendSuccessCallbacks` → `segment_handle.go:371 op.FastSuccess(ctx)`. `FastSuccess` (`append_op.go:416-435`) iterates `op.resultChannels` with no `IsClosed()` guard, and both `RemoteResultChannel.SendResult` (`remote_result_channel.go:71-73`) and `LocalResultChannel.SendResult` (`local_result_channel.go:55-58`) return an error for a closed channel, hitting the WARN at `:426-429`.
Failure: any normal successful append on the service-mode single-entry remote path (Es=3/Aq=2 or Es=1/Aq=1) → 1–2 lines of `send FastSuccess result to channel failed` per entry (2 lines at Es=3/Aq=2). At the load cited in the PR description (4.88M entries/24h), log volume rises rather than falls, replacing the ~2.8 GB/day of `code = Canceled` WARNs this PR aims to eliminate with an equal or larger volume of new WARNs.
Suggestion: guard the send with `if ch.IsClosed() { continue }`, or drop this `SendResult` entirely — ownership now belongs to the reading goroutine, and on the remote path `ReadResult` already checks `r.result` before `Recv()` (`remote_result_channel.go:93`), so there is no remaining recipient. Existing tests cannot catch this: `append_op_test.go:598` only asserts `NotPanics` and ignores `sendErr`, `append_op_test.go:525` uses fresh unclosed channels, and `append_op_replica_lifetime_test.go:124` mocks `SendAppendSuccessCallbacks` (`:127`) so real `FastSuccess` never runs. Please add an assertion that a quorum-triggered `FastSuccess` produces no `sendErr`, using a real `segmentHandle` or a mock that invokes the real `FastSuccess`. *(raised by czs007, xiaocai2333)*

## Medium

**`woodpecker/segment/batch_append_op.go:124` — batch-path `LocalResultChannel`s now have no closer at all, so a lagging replica's late ack hits the buffer-full branch, logs a WARN, and is silently discarded.**
On the batch path `op.resultChannels[nodeIdx]` is a `LocalResultChannel` (`batch_append_op.go:122-124`); the drain goroutine only reads (`:151`), `applyNodeAck` (`append_op.go:343-370`) does not close, and with `FastSuccess`'s `Close` removed by this PR nothing closes them.
Failure: on the group-commit/batch path at Es=3/Aq=2, when replica C's durability ack for an entry arrives after A and B reach quorum and `FastSuccess` writes its synthetic result into `ch[C][i]` (cap=1, `local_result_channel.go:44`) but before the node's drain goroutine consumes it → `local channel is full or closed` WARN (`local_result_channel.go:70-77`), C's real ack is silently dropped (the error is discarded at `logstore_client_remote.go:268`), and the drain consumes the synthetic result instead of C's actual one. Before this change, `FastSuccess`'s `Close` set `l.closed` first, so the late `SendResult` returned an error silently at `local_result_channel.go:55` with no log.
There is no data-correctness impact: `applyNodeAck` short-circuits on `if op.fastCalled.Load() { return }` (`append_op.go:344`), so the synthetic result never reaches `ackSet`. The impact is log noise plus a silently dropped replica result, triggered by a race rather than on every entry.
Suggestion: give the batch path an explicit owner for these channels — close them in the drain goroutine once it has read its result (or after the batch completes) — so late sends take the closed path rather than the full path. *(raised by czs007, xiaocai2333)*

**`common/channel/remote_result_channel.go:139` — the new read-budget timer surfaces a bare gRPC `Canceled`, which `IsRetryableErr` rejects, so the entry goes straight to terminal failure without spending a single retry and forces segment rolling.**
The budget timer added at `remote_result_channel.go:139-142` (`time.AfterFunc(time.Until(deadline), cancel)`) tears down the gRPC stream on expiry; `ch.Recv()` (`:144`) returns a `*status.Error(codes.Canceled)` which `ReadResult` passes through unwrapped at `:149`. It is stored into `op.channelErrors[serverIndex]` (`append_op.go:303`) and reported via `HandleAppendRequestFailure` (`:305`), but the retry decision at `segment_handle.go:433-435` depends on `werr.IsRetryableErr`, which uses `errors.As` to extract a `woodpeckerError` (`werr/errors.go:320-325`) and therefore returns false for a bare status error. Related: the `errors.IsAny(readChanErr, context.Canceled, context.DeadlineExceeded)` check at `append_op.go:292` also fails to match a bare gRPC status error, so the retry intent expressed by the author's own `// read chan timeout, retry` comment at `:293` never takes effect.
Failure: in service mode at Es=3/Aq=2 with MaxRetries=3, when both replicas have accepted the entry (Buffered returned) but durability acks do not arrive within the 30s budget at `append_op.go:277` (e.g. a stalled flush), so quorum is unmet and `fastCalled` is still false → both ack goroutines are cancelled by the new budget timer, each records a terminal failure with zero retries consumed, `finalFailureSet` reaches `Wq-Aq+1=2`, and the entry plus every queued entry after it is `FastFail`ed back to the caller (`segment_handle.go:472-486`) while `rollingState.Store(true)` (`:491-494`) forces a segment roll.
Scope note: the error-classification gap itself predates this PR (a dropped connection yields the same bare gRPC error and is likewise not retried), and before this change the same scenario left `ReadResult` blocked forever, which was worse — this PR adds a new, reliably triggerable source feeding that classification.
Suggestion: in `ReadResult`, when `ctx.Err() != nil` (or via an explicit flag set by the budget timer), wrap the error as a retryable `woodpeckerError` (`werr.ErrXxx.WithCauseErr(readErr)`) so the retry described at `append_op.go:293` actually happens. *(raised by czs007)*

## Low

**`woodpecker/segment/append_op.go:388` — `FastFail` likewise sends into an already-retired slot, adding a `send FastFail result to channel failed` WARN on terminal-failure paths.**
Same root cause as the critical item: `receivedAckCallback` closes the replica's channel at `:282` (read failure) or `:266` (send failure) before calling `HandleAppendRequestFailure` at `:305`, which synchronously invokes `op.FastFail(ctx, err)` at `segment_handle.go:485`; the `SendResult` at `append_op.go:388` then fails and logs at `:392-395`.
Failure: whenever a replica's `ReadResult` errors before quorum (stream break or budget exhaustion) and the op is driven to terminal failure → one extra `send FastFail result to channel failed` WARN per terminal failure, interleaved with the genuine failure-cause logs.
Suggestion: no functional impact and low volume — cover the `FastFail` loop with the same `IsClosed()` guard used to fix the critical item. *(raised by czs007, xiaocai2333)*

tinswzy:
All four are real. I verified each against the code before fixing, and each fix is backed by a test that was confirmed to **fail against the previous commit** (5d488d7 fixes them).

## Critical — confirmed, and it would have made this PR a net loss

The ordering is exactly as described and it is deterministic, not racy: `retireResultChannel()` at `append_op.go:282` runs before `SetAndCount` at `:319`, and `SendAppendSuccessCallbacks` → `segment_handle.go:371` → `op.FastSuccess(ctx)` all run synchronously in that same goroutine. At Es=3/Aq=2 both acking replicas have closed their own channels by then, so `FastSuccess` produced 2 failed sends and 2 WARNs on **every successful append** — replacing ~2.8 GB/day of `code = Canceled` with a comparable volume of `send FastSuccess result to channel failed`.

I took the second of your two suggestions rather than the `IsClosed()` guard: **`FastSuccess` now touches no result channel at all.** The guard would have silenced the retired ones but left the in-flight replica still receiving a synthetic success — which is read *instead of* that replica's real answer, and is the same class of defect this PR exists to remove. With ownership in the reading goroutine there is no recipient left that benefits.

Test, built the way you asked (real `FastSuccess`, invoked from the mock handle exactly as the quorum path invokes it): `TestAppendOp_QuorumSuccess_WritesNothingIntoRetiredChannels`. It uses a recording channel so "must not write here" is an assertion rather than an absent log line. Against the previous commit it fails on all three replicas.

## Medium (batch) — confirmed, and the fix is upstream of the symptom

Removing `FastSuccess`'s `SendResult` removes the failure you describe at its source: nothing writes into `ch[C][i]`, so the buffer is empty when C's real ack arrives and the drain consumes the real one.

I also did what you suggested and gave those channels an owner: the drain goroutine retires them on its way out (`batch_append_op.go`), which is the same rule the single-entry path now follows. Leaving "nobody owns these" in a PR whose thesis is "the reader owns the channel" would have been incoherent. Asserted in `TestBatchAppendOp_InstallsLocalResultChannelInOpSlot`.

## Medium (retryability) — confirmed, including the second-order effect

`IsRetryableErr` uses `errors.As` for a `woodpeckerError` and returns false for a bare status error, and `errors.IsAny(readChanErr, context.Canceled, context.DeadlineExceeded)` at `:292` does not match one either, since gRPC's `Canceled` does not wrap `context.Canceled`. So the author-intended retry at `:293` never fired and the entry went terminal with zero retries spent.

`ReadResult` now returns `werr.ErrAppendOpTimeout.WithCauseErr(context.DeadlineExceeded)` when its own budget expired, distinguished by a flag the timer sets rather than by `ctx.Err()` (which has a window where it is still nil). That single error satisfies **both** classifications — `IsRetryableErr` matches the `woodpeckerError`, and `errors.Is` reaches `context.DeadlineExceeded` through `Unwrap` — so the retry happens and the read-timeout WARN is the one that fires. The transport error is logged where it occurs, so nothing diagnostic is lost.

Your scope note is right that the classification gap predates this PR; the budget timer is what made it reliably reachable, so it belongs here.

## Low — confirmed, fixed with the guard you suggested

`FastFail` skips channels whose reader already retired them. It keeps its `SendResult` for live waiters, which is what stops them and tells them why. `TestAppendOp_FastFail_SkipsChannelsTheirReaderRetired`.

## Verification

`go test -race ./woodpecker/segment/` clean; `common/channel`, `woodpecker/client`, `woodpecker/log` clean; `gofmt`/`go vet`/`go build ./...` clean.

Each new assertion was run against the previous commit first:

```
--- FAIL: TestAppendOp_QuorumSuccess_WritesNothingIntoRetiredChannels
        FastSuccess wrote into replica 0's channel: ...  (and 1, and 2)
--- FAIL: TestAppendOp_FastFail_SkipsChannelsTheirReaderRetired
        a retired channel has no waiter left to wake
--- FAIL: TestAppendOp_FastSuccess
        FastSuccess must not write into a replica's channel
--- FAIL: TestBatchAppendOp_InstallsLocalResultChannelInOpSlot
        the batch drain must retire the channels it read
--- FAIL: TestRemoteResultChannel_ReadResult_ViaStream_HonoursCallerDeadline
        a timed-out read must be retryable, not terminal
        the append path's read-timeout branch tests for this
```

czs007:

Verified:
- `woodpecker/segment/append_op.go:437` — `FastSuccess` now only flips `fastCalled` and invokes `op.callback`; the `SendResult`/`Close` loop is gone, so the per-append WARN from writing into a channel `receivedAckCallback` already retired at `:282` is structurally impossible.
- `woodpecker/segment/append_op_replica_lifetime_test.go:511` — regression test asserts `sendCount == 0` into retired channels, and `TestAppendOp_FastSuccess` asserts `!IsClosed()` and `!TryReadResult()` on both replicas.
- `woodpecker/segment/batch_append_op.go:153-160` — the drain goroutine now owns its result channels via a deferred close; defers run LIFO, so channels are retired before `streamCancel()`, and a late demux send takes `LocalResultChannel`'s silent closed-return instead of the "full or closed" WARN path.
- `common/channel/remote_result_channel.go:151-162` — read-budget expiry returns `ErrAppendOpTimeout` wrapping `context.DeadlineExceeded`; `IsRetryableErr` and the `errors.IsAny(...DeadlineExceeded)` check at `append_op.go:292` both recognise it, so a timeout spends a retry instead of counting as a final failure.
- `woodpecker/segment/append_op.go:194-215` — the fresh-channel-per-attempt slot rewrite does not widen the race against `FastFail`'s iteration: every write and read of the slot is under `op.mu` (retry entry at `append_request_retry_op.go:50-52`, initial fan-out inside `Execute`, batch path at `batch_append_op.go:125-127`).
- Lock ordering — `receivedAckCallback` closes channels outside `op.mu` taking only the channel's own mutex, while `FastFail` holds `op.mu` then the channel mutex; there is no inverse dependency and no deadlock.

## Low — non-blocking

- `woodpecker/segment/append_op.go:388` — `FastFail`'s new `IsClosed()` skip has a narrow TOCTOU window: the reader can retire the channel in `receivedAckCallback` (`:254`, `:282`, executed outside `op.mu`) between the check at `:388` and `SendResult` at `:394`. Both `LocalResultChannel.SendResult` (`local_result_channel.go:54-57`) and `RemoteResultChannel.SendResult` (`remote_result_channel.go:68-73`) re-check `closed` under their own mutex and return an error, so there is no send-on-closed panic and no correctness impact; the worst case is a rare "send FastFail result to channel failed" WARN at `:399`. This is the residue of the item raised in the previous round at the same location; the author responded in commit `5d488d7` by adding the `IsClosed()` skip, which resolves the deterministic WARN and is accepted. The remaining window is rare and benign, and no further change is requested. (raised by xiaocai2333, czs007)

## Nitpick — non-blocking, recorded for tuning/audit only

- `woodpecker/segment/batch_append_op.go:164` — with `FastSuccess` no longer writing a synthetic result, a batch entry that already reached quorum on other replicas but whose ack from this node is stalled keeps its drain goroutine and stream alive until `ReadResult` runs out the full `batchAckReadTimeout` (30s), and only then triggers the deferred `streamCancel`/`Close`. This extends resource lifetime only: `applyNodeAck` returns early on `fastCalled` (`append_op.go:344-346`), so no failure is reported for such entries in either old or new code, and entries that genuinely depend on this node already waited out the same timeout before this PR. Behaviour matches the single-entry path's 30s budget in `remote_result_channel.go:141-147`. No code change needed. (raised by czs007)

tinswzy:
Added `ed9cfb1`: the per-replica counters now count the replica that answers after the quorum has completed. This was listed as **Out of scope** in #287 and is now in scope, because the rest of this PR is what makes it correct.

## The gap

`replica_append_total` read **2.01 outcomes per entry** in the field where `Es=3` predicts 3.0. `recordReplicaResult` sits after the fast-completed guard in both ack paths:

```go
if op.fastCalled.Load() {
    logger...Debug("received ack but already fast completed", ...)
    return                                    // ← guard
}
...
op.recordReplicaResult(serverIndex, "success")   // ← never reached for the 3rd replica
```

With `Aq=2` the third replica *always* answers once the op is already done, so its outcome was dropped and `replica_append_bytes_total` understated cross-AZ write traffic by the same third.

## Why it belongs in this PR and not before it

Until `FastSuccess` stopped cancelling a replica that was still answering, that replica's read returned `rpc error: code = Canceled`. Counting it then would have recorded a **fabricated error for every entry** — trading an undercount for a worse overcount, and putting our own cancellations into the error series, which is exactly what this PR removes from the logs.

Now the replica reads its real result, so there is something true to record. The fix is unblocked by the change, not independent of it.

## The rule

**After the op completes, only an *answer* is counted.** A read that *failed* after completion may be our own doing — `FastFail` cancels every replica still in flight — so it is not counted. An answer that *reports* a failure is a real failure and still is.

`replicaOutcome` is now the single place that decides what "success" means for these counters, so the single-entry ack, the batch drain, and the post-quorum recording in both cannot drift; `TestReplicaOutcome_ClassifiesTheSameWayTheAckPathsDo` pins that it matches the existing branches.

## Verified against the previous behaviour

With the recording removed again:

```
--- FAIL: TestAppendOp_ReplicaAnsweringAfterQuorumIsStillCounted
        expected: 3   actual: 2
        every replica that answered must be counted, not just the ack quorum
--- FAIL: TestApplyNodeAck_ReplicaAnsweringAfterQuorumIsStillCounted
        expected: 3   actual: 2
--- FAIL: TestAppendOp_ReplicaAnsweringWithAnErrorAfterQuorumIsCounted
        expected: 1   actual: 0
```

`TestAppendOp_ReplicaFailingAfterCompletionIsNotCounted` covers the other half and passes both before and after — it guards behaviour this must *not* change.

## Still not exact

The bytes counted are `len(op.value)`, the entry payload, so the metric still understates what actually crosses an AZ boundary by the protobuf and gRPC framing around it. It is no longer off by a factor, though, which is what made it unusable for traffic accounting.

`go test -race ./woodpecker/segment/ ./common/channel/` clean.

tinswzy:
Correction to `ed9cfb1`, fixed in `5bd6926`. Reviewing that commit rather than assuming it was right turned up a hole in it.

It gated the post-quorum recording on `readChanErr == nil`. The two channel implementations do not agree on what an error means:

```go
// RemoteResultChannel.ReadResult — the server answered "sync failed"
syncedErr := werr.ErrFileWriterSyncFailed.WithCauseErrMsg(...)
return &AppendResult{SyncedId: ..., Err: syncedErr}, syncedErr   // result AND error, both non-nil

// LocalResultChannel.ReadResult — the same answer
return r, nil                                                    // no error at all
```

So a replica that answered *with a write failure* after quorum was counted in embedded mode and **silently dropped in service mode** — the mode this whole change is about, and the case the counter is most useful for (one replica failing consistently while the quorum hides it).

The discriminator is the **result**, not the error: a remote read returns `(nil, err)` for a transport failure and `(result, err)` for an answer that reports one. `5bd6926` keys on `result != nil` in both ack paths.

Worth noting why `TestAppendOp_ReplicaAnsweringWithAnErrorAfterQuorumIsCounted` did not catch it: it calls `applyNodeAck` directly with `readErr = nil`, so it never exercised the remote convention. The new `TestAppendOp_RemoteReplicaAnsweringWithAFailureAfterQuorumIsCounted` drives a real `RemoteResultChannel` whose stream answers `AddEntryState_Failed`, which is the shape the failure actually arrives in. Against `ed9cfb1` it fails:

```
--- FAIL: TestAppendOp_RemoteReplicaAnsweringWithAFailureAfterQuorumIsCounted
        expected: 1   actual: 0
        a remote replica that answered with a write failure is a real failure,
        even though ReadResult reports it as an error as well as a result
```

The rest of the reasoning is unchanged: only an answer counts, a read that produced none stays uncounted because it may be our own cancellation.

`go test -race ./common/channel/ ./woodpecker/segment/ ./woodpecker/client/` clean.

czs007:

**Overview:** After verification, the remaining findings on this PR concern post-completion replica accounting and channel-ownership side effects introduced by the new per-attempt channel lifecycle; none affect the data path.

### Medium

- **`woodpecker/segment/append_op.go:338`** (also `:400-404`) — The post-completion replica accounting added in this PR treats `result != nil` as "the replica actually answered", but on a `LocalResultChannel` a `FastFail` produces a non-nil synthetic result: `FastFail` CAS-sets `fastCalled`, then `SendResult({SyncedId:-1, Err})` on the still-open channel, then `Close`. Since `LocalResultChannel` is a cap-1 buffered channel and `Close` only does `close(l.ch)`, the blocked reader receives the buffered value with `ok == true`, hits the `fastCalled` branch, calls `recordReplicaAnswer`, and `replicaOutcome` classifies `SyncedId == -1` as `"error"`. This is the default path for both embedded single-entry appends (`append_op.go:267`) and the batch drain (`batch_append_op.go:124,165`). Failure: a cascaded `FastFail` from `segment_handle.go:472-485` on ops with in-flight replica acks → every in-flight, healthy replica of every cascaded op is counted once in `WpClientReplicaAppendTotal{status="error"}`, attributing the client's own cancellation to the replicas, and later successful flushes never correct it (amplification = cascaded ops × in-flight replicas). The baseline returned immediately in the `fastCalled` branch without counting, so this is introduced here. The guarding test `TestAppendOp_ReplicaFailingAfterCompletionIsNotCounted` (`append_op_replica_lifetime_test.go:388-391`) only calls `Close()` without a preceding `SendResult`, so it models the remote-style empty-buffer close rather than the real send-then-close sequence; seeding `{SyncedId:-1, Err}` before `Close` reproduces the miscount. Suggestion: mark the synthetic `FastFail` result (a dedicated sentinel error works) and have the reader skip `recordReplicaAnswer` when `errors.Is(result.Err, sentinel)` in the `fastCalled` branch, and extend the test accordingly. Impact is limited to metric accuracy. *(raised by xiaocai2333, bigsheeper)*

### Low

- **`woodpecker/segment/batch_append_op.go:153`** (same shape at `append_op.go:335`) — The new deferred owner-close of all `resultChs` assumes the closed path is silent, but in embedded mode (`IsRemoteClient() == false`) these channel objects are handed to the server unchanged via `logstore_client_local.go:40/49` and stored as `BufferEntry.NotifyChan` (`sequential_buffer.go:126-131`). `LocalResultChannel.SendResult` itself is silent on a closed channel, but the caller `NotifyPendingEntryDirectly` logs `WARN "Send result to channel failed"` on every failed send (`sequential_buffer.go:434-443`). Failure: a replica's flush exceeds `batchAckReadTimeout` (or `appendAckReadTimeout` on the single path), the drain fast-forward-fails `i+1..N` and returns, the defer closes all N channels → when that flush later completes, the server logs one WARN per closed channel (N−i+1 lines for batch, 1 for single). The baseline never closed these channels in the drain, so late notifications landed in the empty buffer silently. Net WARN volume is still far lower than before this PR since this only fires on flush timeouts, but the comment at `:149-152` does not hold for embedded mode. Remote mode is unaffected because the server uses its own channels and the demux discards `SendResult` errors. Suggestion: either skip the owner-close for local channels (let the server-side notify land in the buffer as before), or make the server-side notify treat `ErrAppendOpResultChannelClosed` as expected and log at debug level; at minimum, fix the comment. *(raised by bigsheeper)*

- **`woodpecker/segment/append_op.go:247`** — `sendWriteRequest` now unconditionally creates a fresh channel per attempt, and neither `AppendRequestRetryOp.Execute` (`append_request_retry_op.go:43-53`) nor `sendWriteRequestRetry` checks `fastCalled`. In the baseline the reused channel had already been closed by `FastFail`, so a queued retry short-circuited passively: remote `InitResponseStream` cancelled the stream on `closed` (`remote_result_channel.go:235-240`) and `ReadResult` returned immediately on both channel types. With a fresh channel that short-circuit is gone. Failure: a retry for entry N is already queued on the executor (`segment_handle.go:456`) when N is cascade-`FastFail`ed, then the retry runs and replica k accepts the entry but is slow to ack → the retry opens a real `AddEntry` stream and holds the stream plus the `receivedAckCallback` goroutine until the replica acks or `appendAckReadTimeout` (30s) expires, where the baseline released immediately. Bounded resource holding only; resending the entry itself is idempotent server-side and already happened in the baseline. Suggestion: check `fastCalled` at the top of `AppendRequestRetryOp.Execute` or `sendWriteRequest` and return early. *(raised by bigsheeper, xiaocai2333)*

**Previously raised items:** the earlier findings on `FastSuccess`/`FastFail` writing to already-closed channels, the unclosed batch channels, and the bare `Canceled` from the read budget were re-checked in this round and are resolved by the current commits (`FastSuccess` no longer touches channels, `FastFail` skips closed slots, the batch drain owns and closes its channels, and `readBudgetExhausted` wraps `DeadlineExceeded` as a retryable timeout). They are not re-raised here. The medium item above was first aggregated in this round and has not yet received an author response.

czs007:

A quick preface: the rendered comment is below.

---

Adversarial review confirmed one open finding, a metrics-correctness issue introduced by this PR's new post-quorum accounting in the FastFail path; no data-loss or availability defects were found.

## Medium

- **`woodpecker/segment/append_op.go:466`** — `FastFail` sends a synthetic `{SyncedId: -1, Err}` result into every `LocalResultChannel` that has not yet been retired, and the reader then records it as a real replica answer. `IsClosed()` only skips channels the reader already retired; a reader still blocked in `ReadResult` receives the buffered synthetic value (`local_result_channel.go:44`, `:82-93`) as a non-nil result, and the new `fastCalled` branches in `receivedAckCallback` (`append_op.go:336-339`) and `applyNodeAck` (`append_op.go:413-416`) pass it to `recordReplicaAnswer`, where `replicaOutcome` maps `SyncedId == -1` to `"error"`. This contradicts the rule stated in the `recordReplicaAnswer` comment (`append_op.go:148-152`) that self-inflicted FastFail cancellations must not enter the error series. That rule holds only for `RemoteResultChannel`, where `Close` cancels the stream and the synthetic value is never observable (`remote_result_channel.go:113-117`, `:149-167`). Failure scenario: a retry-exhausted entry causes `segment_handle` to `FastFail` the ops queued behind it while they still have replicas in flight → each in-flight replica's reader consumes the synthetic result and increments `replica_append_total{status="error"}` for a replica that never actually answered (up to 3 spurious errors per FastFailed entry on a 3-replica cluster). This is not limited to embedded mode: the batch path installs a `LocalResultChannel` for every node regardless of client type (`batch_append_op.go:120-127`), and the remote client delivers acks into those channels via `SendResult` (`logstore_client_remote.go:268-270`), so service-mode batch writes hit the same path. The existing test `TestAppendOp_ReplicaFailingAfterCompletionIsNotCounted` (`append_op_replica_lifetime_test.go:388-391`) models FastFail as a bare `Close()` with no preceding `SendResult`, so it exercises the remote-cancel shape rather than the real send-then-close behavior and does not catch this. Suggestion: have `FastFail` only `Close` local channels without sending a synthetic result (closing already wakes `ReadResult`, which returns nil plus `ErrAppendOpResultChannelClosed` and is correctly ignored), or make the `fastCalled` branches distinguish FastFail-injected results from genuine replica answers. Add a test that calls the real `op.FastFail` against a live `LocalResultChannel` and then drives `receivedAckCallback` / `applyNodeAck`, asserting the error counter is unchanged. Note: this was raised in an earlier round of this review (at `append_op.go:338`). We have no record of an author response addressing it, and the current diff still contains the same send-then-close path feeding `recordReplicaAnswer`, so it is re-listed here rather than presented as a new discovery. (raised by xiaocai2333)

PR #307, opened 2026-09-08
fix(disk): accept a finalized empty segment on recovery; share footer index parsing across local backends

### LINKED ISSUE #306: disk storage: recovery rejects a finalized empty segment (IndexLength=0), so fencing it fails and the log can never be reopened
## Summary

With `woodpecker.storage.type = local`, a segment that was **finalized while empty** (zero data blocks) cannot be reopened in recovery mode: `LocalFileWriter.recoverBlocksFromFooterUnsafe` rejects its footer as corrupt. Because `segmentProcessor.Fence` always opens the writer in recovery mode, such a segment can never be fenced again. If its metadata is still `Active` when the next writer opens the log, `OpenLogWriter` fails on every attempt and the log is unopenable until the data is wiped.

Seen from Milvus (standalone, embedded woodpecker, local storage, master `d6179da4f7`, woodpecker v0.1.40): after a restart the streaming node could not open the WAL for two pchannels, the streamingcoord balancer looped in backoff, the channels stayed in `PCHANNEL_META_STATE_ASSIGNING`, DDL blocked indefinitely, and `healthz` kept returning OK.

```
[ERROR] [wp/opener.go:40] ["failed to open log writer"] [log_name=...-dml_0]
        [error="failed to fence log: failed to fence segment 6: recover from existing file: invalid footer record: IndexOffset=25, IndexLength=0 ..."]
[WARN]  [walmanager/wal_lifetime.go:162] ["open new wal fail"] [channel=...-dml_0] [expected=(130,true)] [current=(129,false)]
[WARN]  [balancer/balancer_impl.go:420] ["fail to apply balance, start a backoff..."]
```

## Root cause

`Finalize` on a segment with zero blocks writes a complete, valid file: the 25-byte header record, no data blocks, no index records, then the footer. The footer therefore says `TotalBlocks=0, TotalSize=25, IndexOffset=25, IndexLength=0` (CRC and magic correct; the file is 78 bytes with a V6 footer).

`recoverFromExistingFileUnsafe` parses that footer successfully and calls `recoverBlocksFromFooterUnsafe`, which starts with (`server/storage/disk/writer_impl.go:1190`, v0.1.41):

```go
if footerRecord.IndexOffset == 0 || footerRecord.IndexLength == 0 {
    return fmt.Errorf("invalid footer record: IndexOffset=%d, IndexLength=%d", ...)
}
```

`IndexLength == 0` is the legitimate value for an empty finalized segment, not a sign of corruption. A truly torn footer would fail CRC, fail to parse, and fall into the tolerant full-scan recovery path; the well-formed empty footer is the only case that hits this branch.

The fence path always reopens in recovery mode: `segmentProcessor.Fence` → `getOrCreateSegmentWriter(ctx, true)` → `NewLocalFileWriterWithMode(..., recoveryMode=true)` → `recoverFromExistingFileUnsafe`. So any empty finalized segment whose metadata is still `Active` (completion phase 1 done on the log store, phase 2 metadata update never landed, e.g. the process died in between) makes `OpenLogWriter` → `fenceAllActiveSegments` fail forever.

## Only the disk backend

The three backends implement footer recovery separately:

| backend | `IndexLength == 0` on a finalized segment |
|---|---|
| `disk` (`server/storage/disk/writer_impl.go:1190`) | rejected as `invalid footer record` |
| `stagedstorage` (`server/storage/stagedstorage/writer_impl.go:2210`) | accepted: "empty index length, no data blocks at all, fast return" |
| `objectstorage` (`recoverFromFooter`) | no such check; an empty index section parses to zero records |

The disk and staged functions are otherwise line-for-line identical apart from this branch (and a `recoveredFooter` assignment), so this is duplicated recovery logic that diverged.

## Reproduction

Writer level (deterministic, no external services), against v0.1.40 / v0.1.41:

```go
w, _ := disk.NewLocalFileWriter(ctx, dir, 1, 7, cfg)
w.Finalize(ctx, -1)              // 78-byte file: header + footer, zero blocks
w.Close(ctx)
_, err := disk.NewLocalFileWriterWithMode(ctx, dir, 1, 7, cfg, true)
// err: recover from existing file: invalid footer record: IndexOffset=25, IndexLength=0
```

Client level (embedded client, `storage.type=local`, needs etcd): create a log, create a segment without writing to it, finalize the segment file the way the log store does, keep the metadata `Active`, then open a second writer. `OpenLogWriter` fails with

```
failed to fence log: failed to fence segment 0: recover from existing file: invalid footer record: IndexOffset=25, IndexLength=0
```

through `disk/writer_impl.go:201 → processor/segment_processor.go:165 → segment_handle.go (Fence) → log_handle.go (fenceAllActiveSegments)`, the same chain as the Milvus logs.

## Proposed fix

1. Treat `IndexLength == 0 && TotalBlocks == 0` as a valid empty finalized segment in the disk backend (empty block index, first/last entry id `-1`, `finalized` set), keeping `IndexOffset == 0` and `IndexLength == 0 && TotalBlocks != 0` as errors.
2. Move footer validation and index-section parsing into `codec` and have both local-file backends call it, so "an empty finalized segment is valid" is defined once.

A PR with the fix, the shared codec function, a disk-writer unit test and an integration test for the reopen path follows.

czs007:
Fixes #306

## Problem

`Finalize` on a segment with zero blocks writes a complete, valid file: header record, no blocks, no index records, footer. Its footer therefore says `IndexOffset = 25` (the header size) and `IndexLength = 0`. The disk backend's `recoverBlocksFromFooterUnsafe` rejected exactly that footer:

```go
if footerRecord.IndexOffset == 0 || footerRecord.IndexLength == 0 {
    return fmt.Errorf("invalid footer record: ...")
}
```

`segmentProcessor.Fence` always reopens the writer in recovery mode, so an empty segment that had been finalized on the log store but was still `Active` in metadata (completion phase 2 never landed) could never be fenced again, and `OpenLogWriter` → `fenceAllActiveSegments` failed on every attempt. From Milvus this is a WAL that never reopens after a restart (`failed to fence log: failed to fence segment N: recover from existing file: invalid footer record: IndexOffset=25, IndexLength=0`), with the balancer in backoff, the pchannel stuck in `ASSIGNING`, DDL blocked and `healthz` still OK.

A torn footer does not reach this branch: it fails CRC, does not parse, and goes to the full-scan recovery. Only the well-formed empty footer did.

## Change

- `codec`: new `ValidateIndexSection`, `ParseIndexRecords`, `ReadIndexSection`. One definition of which footers are acceptable: `IndexOffset == 0` is invalid; `IndexLength == 0 && TotalBlocks == 0` is an empty finalized segment (empty index, no entries); `IndexLength == 0 && TotalBlocks != 0` is invalid; otherwise the section is read and parsed strictly, as before.
- `disk` and `stagedstorage` writers: `recoverBlocksFromFooterUnsafe` now calls `codec.ReadIndexSection` and only applies the result to writer state. The two functions were line-for-line identical except that staged already accepted the empty case; the divergence is what let the disk backend keep the bug. Behavior for populated segments is unchanged (same strict parsing, same state updates including `currentBlockNumber`).
- `objectstorage` is untouched: it recovers from object listings and already parses an empty index section to zero records.

## Tests

- `codec/index_section_test.go`: validation rules, round trip, length mismatch, wrong record type, corrupt record (CRC), reading an empty finalized file and a populated one.
- `disk/writer_recover_empty_test.go`: finalize an empty segment → reopen in recovery mode → `Fence` → `Finalize` again (idempotent); and the same flow for a segment with data.
- `tests/integration/local_empty_segment_fence_test.go` (storage type `local`, needs etcd): a segment is created but never written, its file is finalized while its metadata stays `Active`, then a second client opens the log and writes.

Verified locally: the integration test **fails on `master` without the fix** with the exact production error above, and passes with it. `go test ./server/storage/codec/ ./server/storage/disk/ ./server/storage/stagedstorage/` pass; `gofumpt`/`gci` clean; `golangci-lint` (repo config) reports no issues on the touched packages.

### DIFF
--- a/server/storage/codec/index_section.go
+++ b/server/storage/codec/index_section.go
@@ -1,0 +1,101 @@
+// Licensed to the LF AI & Data foundation under one
+// or more contributor license agreements. See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership. The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License. You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package codec
+
+import (
+	"fmt"
+	"io"
+)
+
+// ValidateIndexSection checks that a parsed footer describes a readable index
+// section. It is the single definition of which footers are acceptable, shared
+// by every storage backend that recovers a finalized segment from its footer.
+//
+// A segment that was finalized with zero blocks is valid: Finalize writes the
+// header and the footer with nothing in between, so IndexOffset equals the
+// header size and IndexLength is 0. Such a file is complete, it just carries no
+// entries; it must not be reported as corrupt, or the segment can never be
+// fenced or read again.
+func ValidateIndexSection(footer *FooterRecord) error {
+	if footer == nil {
+		return fmt.Errorf("invalid footer record: nil")
+	}
+	if footer.IndexOffset == 0 {
+		// Every finalized file starts with a header record, so the index
+		// section can never begin at offset 0.
+		return fmt.Errorf("invalid footer record: IndexOffset=%d, IndexLength=%d",
+			footer.IndexOffset, footer.IndexLength)
+	}
+	if footer.IndexLength == 0 && footer.TotalBlocks != 0 {
+		return fmt.Errorf("invalid footer record: IndexLength=0 but TotalBlocks=%d", footer.TotalBlocks)
+	}
+	return nil
+}
+
+// ParseIndexRecords parses the index section of a finalized segment. indexData
+// is exactly the IndexLength bytes starting at IndexOffset. Every record must be
+// a well-formed IndexRecord; a trailing fragment shorter than a record header is
+// ignored. For an empty index section (a segment finalized with zero blocks)
+// the result is an empty, non-nil slice.
+func ParseIndexRecords(indexData []byte, footer *FooterRecord) ([]*IndexRecord, error) {
+	if err := ValidateIndexSection(footer); err != nil {
+		return nil, err
+	}
+	records := make([]*IndexRecord, 0, footer.TotalBlocks)
+	if footer.IndexLength == 0 {
+		return records, nil
+	}
+	if len(indexData) != int(footer.IndexLength) {
+		return nil, fmt.Errorf("index section is %d bytes, footer says %d", len(indexData), footer.IndexLength)
+	}
+
+	recordSize := RecordHeaderSize + IndexRecordSize
+	offset := 0
+	for offset < len(indexData) {
+		if offset+RecordHeaderSize > len(indexData) {
+			break
+		}
+		record, err := DecodeRecord(indexData[offset:])
+		if err != nil {
+			return nil, fmt.Errorf("failed to decode index record at offset %d: %w", offset, err)
+		}
+		if record.Type() != IndexRecordType {
+			return nil, fmt.Errorf("expected index record type %d, got %d at offset %d",
+				IndexRecordType, record.Type(), offset)
+		}
+		records = append(records, record.(*IndexRecord))
+		offset += recordSize
+	}
+	return records, nil
+}
+
+// ReadIndexSection reads the index section described by footer from r and
+// parses it with ParseIndexRecords. It is what a local-file backend calls when
+// it reopens a finalized segment for recovery.
+func ReadIndexSection(r io.ReaderAt, footer *FooterRecord) ([]*IndexRecord, error) {
+	if err := ValidateIndexSection(footer); err != nil {
+		return nil, err
+	}
+	if footer.IndexLength == 0 {
+		return make([]*IndexRecord, 0), nil
+	}
+	indexData := make([]byte, footer.IndexLength)
+	if _, err := r.ReadAt(indexData, int64(footer.IndexOffset)); err != nil {
+		return nil, fmt.Errorf("failed to read index section: %w", err)
+	}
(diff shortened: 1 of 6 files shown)


tinswzy:

Verified:
- `server/storage/codec/index_section.go:37` — both local `Finalize` paths write the header record before capturing `indexStartOffset`, so a local file's `IndexOffset` is always ≥ 25 and the new rejection rule matches the write side.
- `server/storage/codec/index_section.go:43` — a zero `IndexLength` is accepted only when `TotalBlocks == 0`, matching the header-plus-footer layout both local writers actually produce.
- `server/storage/stagedstorage/writer_impl.go:2184` — the only staged caller passes footers read from the tail of a local file; remote compacted footers (which carry `IndexOffset: 0`) travel the separate `readRemoteFooter` path, so the tightened validation causes no regression.
- `server/storage/disk/writer_impl.go:200` — recovery completes before the flush goroutine starts, so the extracted `...Unsafe` helpers introduce no concurrent mutation of writer state.
- `tests/integration/local_empty_segment_fence_test.go:104` — the test explicitly asserts metadata is still `Active` before reopening the writer, so a self-invalidating "the close already completed the segment" scenario would be caught.

Tests were inspected, not executed.

## Low — non-blocking

- `server/storage/codec/index_section.go:26` — The doc comment declares `ValidateIndexSection` to be "the single definition of which footers are acceptable, shared by every storage backend that recovers a finalized segment from its footer", and the `IndexOffset == 0` branch asserts that "the index section can never begin at offset 0". That invariant holds only for the local-file backends. Three legitimate footers in this repo always carry `IndexOffset: 0`, because their index section starts at the beginning of a standalone footer object: `objectstorage/writer_impl.go:932`, `objectstorage/writer_impl.go:2538` (`serializeFooterAndIndexes`), and `stagedstorage/writer_impl.go:2046` (`uploadCompactedFooter`). No caller is affected today — objectstorage uses its own inline parsing loop (`objectstorage/writer_impl.go:261-325`) and staged storage only routes local-file footers here — so this is an over-broad scope claim in the contract, not a runtime defect. Suggestion: narrow the comment to the local-file backend, or parameterize the offset rule so the caller declares it, so that a later change converging objectstorage onto `ParseIndexRecords` (which this comment actively encourages) does not start rejecting healthy finalized segments. (raised by bigsheeper)

PR #311, opened 2026-09-09
enhance(test): cover empty-segment recovery on every backend and scope the shared footer validator

tinswzy:
Follow-up to #307.

## Why

#307 fixed the disk backend rejecting a finalized empty segment. The coverage that should have caught it existed — but in the wrong package.

`stagedstorage` has had `TestRecoverBlocksFromFooter_EmptyIndexLength` for a while: finalize an empty segment, reopen in recovery mode, assert the recovered state. That is exactly the #306 case. It lives on the one local backend that already handled it, while `disk` — byte-for-byte the same function apart from that one branch — never got a mirror.

Running that same test against `disk` on the pre-#307 code reproduces the production error verbatim:

```
recover from existing file: invalid footer record: IndexOffset=25, IndexLength=0
```

So the failure mode was not "nobody thought of empty segments". It was that coverage is organised per package rather than per path, and a shared behaviour drifted between two copies of it. This PR closes the same gap on the other two backends and on the fence path.

## Tests added

**`objectstorage`** — had no test that finalized an empty segment and recovered it.

- `TestMinioFileWriter_RecoverFromFooter_EmptyFinalizedSegment` — a footer.blk holding only the footer record recovers to zero blocks, finalized, not writable.
- `TestObjectStorageFooterIsNotValidatedAsInlineIndex` — this backend keeps its index at offset 0 of a separate `footer.blk`, so `IndexOffset=0` is its **normal** value, the opposite of the local-file backends where it marks corruption. The test pins that `codec.ValidateIndexSection` rejects these footers *by design*, so a future attempt to "unify" `recoverFromFooter` onto the shared helper fails a test instead of declaring every object-storage segment corrupt.

**`stagedstorage`** — `TestRecoverBlocksFromFooter_EmptyIndexLength` stops at inspecting recovered state; the #306 sequence continues into `Fence`, which reopens the writer in recovery mode. That combination was never exercised here — the same shape of gap that let the defect survive in `disk`. Its existing `Fence` tests all fence a fresh or written-to writer, never a recovered finalized empty one.

- `TestStagedFileWriter_RecoverFinalizedEmptySegment_ThenFence` — the mirror of disk's `TestLocalFileWriter_RecoverFinalizedEmptySegment`: finalize empty → reopen in recovery mode → `Fence` → idempotent `Finalize`.
- `TestStagedFileWriter_RecoverFromFooter_RejectsZeroIndexOffset` and `..._RejectsIndexLengthZeroWithBlocks` — #307 routed this backend through `codec.ValidateIndexSection`, replacing a fast path that accepted **any** footer with `IndexLength == 0` regardless of `IndexOffset`. That is a real tightening of staged's contract, and the codec package tests the rule, not that this backend applies it.

**`tests/integration`** — `local_empty_segment_fence_test.go` hard-codes the on-disk layout it writes the finalized empty file to. If that layout ever drifts, the file lands where the log store never looks, `recoverFromExistingFileUnsafe` takes its `os.IsNotExist` fast path, `OpenLogWriter` trivially succeeds, and the test passes **with the bug fully present**. It now asserts the `write.fence` flag that `disk.Fence` drops beside `data.log`, which can only exist if the log store resolved the same path.

## Comment fix

`ValidateIndexSection` claimed to be "the single definition of which footers are acceptable, shared by every storage backend that recovers a finalized segment from its footer". Two footer flavours in this repository legitimately carry `IndexOffset = 0`:

| footer | IndexOffset | reader |
| --- | --- | --- |
| disk / staged `data.log` (index inline) | `>= 25` | `codec.ReadIndexSection` |
| objectstorage `footer.blk` | `0` | `recoverFromFooter` slices the object |
| staged compacted `footer.blk` | `0` | `parseIndexDataUnsafe` starts at 0 |

The comment now states which backends the rule covers and which it must not be used for, so the invitation to unify the remaining two is no longer sitting in the code.

## License headers

The three files added by #307 carry the LF AI & Data / Apache-2.0 header, while every other file in `server/storage/codec` and `server/storage/disk` uses the Zilliz AGPLv3/SSPLv1 header — 45 files to 2 across `server/`. In a dual-licensed repository that is a licensing statement, not formatting, so they are restored to the surrounding convention. (`tests/integration` genuinely uses the LF AI header, 22 files to 2, so that file is left alone.)

## Verification

- `server/storage/{codec,disk,stagedstorage,objectstorage}` all pass.
- `gofmt`, `gofumpt` and `gci` clean on every changed file.
- The integration assertion was exercised locally: the `write.fence` check passes, and the test then fails at the following write on `logstore local disk above hard watermark` — a local disk-space condition on my machine, not the test. CI runs it on a clean runner.

## Still open

The error path in `recoverBlocksFromFooterUnsafe` still emits no metric — `recover_footer` is only recorded with `status="success"`, so during an outage of exactly this kind the dashboards show the success counter going flat with no error series to alert on. Left out here to keep this to tests and comments; happy to send it separately.

### DIFF
--- a/server/storage/codec/index_section.go
+++ b/server/storage/codec/index_section.go
@@ -1,18 +1,19 @@
-// Licensed to the LF AI & Data foundation under one
-// or more contributor license agreements. See the NOTICE file
-// distributed with this work for additional information
-// regarding copyright ownership. The ASF licenses this file
-// to you under the Apache License, Version 2.0 (the
-// "License"); you may not use this file except in compliance
-// with the License. You may obtain a copy of the License at
+// Copyright (C) 2025 Zilliz. All rights reserved.
 //
-//     http://www.apache.org/licenses/LICENSE-2.0
+// This file is part of the Woodpecker project.
+//
+// Woodpecker is dual-licensed under the GNU Affero General Public License v3.0
+// (AGPLv3) and the Server Side Public License v1 (SSPLv1). You may use this
+// file under either license, at your option.
+//
+// AGPLv3 License: https://www.gnu.org/licenses/agpl-3.0.html
+// SSPLv1 License: https://www.mongodb.com/licensing/server-side-public-license
 //
 // Unless required by applicable law or agreed to in writing, software
-// distributed under the License is distributed on an "AS IS" BASIS,
+// distributed under these licenses is distributed on an "AS IS" BASIS,
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-// See the License for the specific language governing permissions and
-// limitations under the License.
+// See the license texts for specific language governing permissions and
+// limitations under the licenses.
 
 package codec
 
@@ -22,8 +23,17 @@ import (
 )
 
 // ValidateIndexSection checks that a parsed footer describes a readable index
-// section. It is the single definition of which footers are acceptable, shared
-// by every storage backend that recovers a finalized segment from its footer.
+// section of a segment file that stores its index inline -- the local-file
+// backends, disk and stagedstorage, whose data.log is
+// header | blocks | index records | footer.
+//
+// It is NOT valid for footers whose index lives in a separate object. In
+// objectstorage, and for stagedstorage's compacted footer, the index records
+// start at offset 0 of footer.blk, so those footers carry IndexOffset = 0 as
+// their normal value and the readers never consult IndexOffset at all
+// (objectstorage.recoverFromFooter and stagedstorage's parseIndexDataUnsafe
+// slice the object instead). Routing them through this function would reject
+// every object-storage segment as corrupt.
 //
 // A segment that was finalized with zero blocks is valid: Finalize writes the
 // header and the footer with nothing in between, so IndexOffset equals the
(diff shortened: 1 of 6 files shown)


czs007:

Verified:
- `server/storage/codec/index_section.go:43` — the change is comment-only; the validation body (`IndexOffset == 0` rejection, `IndexLength == 0 && TotalBlocks != 0` rejection) is unchanged, so there is no behavioral risk.
- `server/storage/codec/index_section.go:25` — the narrowed contract matches reality: the only callers are the two inline-index paths (`disk/writer_impl.go:1192`, `stagedstorage/writer_impl.go:2213`), and both compacted-footer writers set `IndexOffset: 0` while neither reader consumes it.
- `server/storage/objectstorage/writer_recover_empty_test.go:43` — the fixture is built with the production serializer (`serializeFooterAndIndexes(ctx, nil)`), the same bytes a zero-block `Finalize` uploads, so the test exercises a genuinely reachable state rather than a synthetic input.
- `server/storage/stagedstorage/writer_recover_empty_test.go:67` — the hand-built file is 78 bytes, above `GetMaxFooterReadSize()` (53), so recovery takes the footer branch instead of silently falling back to a full scan; without this the two rejection tests would pass vacuously.
- `server/storage/stagedstorage/writer_recover_empty_test.go:51` — the recovery constructor cancels and closes on failure (`writer_impl.go:259-265`, `:2166`), and the recovered writer is closed after the idempotent-finalize check, so no goroutine or descriptor leaks into sibling tests in the package.
- `tests/integration/local_empty_segment_fence_test.go:116` — the asserted `write.fence` path matches `getFenceFlagPath`, and fence creation is synchronous inside `OpenLogWriter` (`os.Create` in `disk/writer_impl.go:1074`), so the assertion is neither racy nor vacuous.

PR #321, opened 2026-09-10
enhance(config): expose the two settings missing from woodpecker.yaml

### LINKED ISSUE #316: config: woodpecker.yaml omits two enforced settings and one comment states the wrong value
`config/woodpecker.yaml` is what an operator reads to learn what is tunable. Two enforced settings
are missing from it entirely, and one comment states a value that is off by ~8x. Three edits, all
to that file.

### 1. Add `logstore.segmentCompactionPolicy.timeout`

Defined at `common/config/configuration.go:196`, validated at `:812`, and enforced at
`server/processor/segment_processor.go:641`, where it wraps the entire `writer.Compact()` call:

```go
timeout := time.Duration(cfg.Woodpecker.Logstore.SegmentCompactionPolicy.Timeout.Seconds()) * time.Second
ctx, cancel := context.WithTimeout(ctx, timeout)
defer cancel()
```

Default is 300s. It appears nowhere in the YAML, so there is no way to discover from the shipped
config that a 5-minute compaction deadline exists, let alone tune it.

```yaml
    segmentCompactionPolicy:
      maxBytes: 2000000
      maxParallelUploads: 4
      maxParallelReads: 8
      timeout: 300 # Deadline for one segment compaction in seconds, default is 300s. Covers the whole read/merge/upload pass; exceeding it aborts the compaction and the auditor retries on its next cycle.
```

### 2. Add the `woodpecker.runtime` section

`Runtime.OpRegistry.Capacity` and `Runtime.OpRegistry.WarnAge` are read at `cmd/main.go:203-204`
to build the in-flight operation registry that backs `wp ops stats` and the op metrics:

```go
opReg := opregistry.New(
    cfg.Woodpecker.Runtime.OpRegistry.Capacity,
    time.Duration(cfg.Woodpecker.Runtime.OpRegistry.WarnAge.Seconds())*time.Second,
)
```

Defaults are 1024 and 30s (`configuration.go:991-994`). The whole `runtime` section is absent from
the YAML.

```yaml
  # Runtime subsystem configuration.
  runtime:
    opRegistry:
      capacity: 1024 # Size of the in-flight operation registry, default is 1024
      warnAge: 30 # Age in seconds past which an in-flight operation is reported as stalled, default is 30s
```

### 3. Fix the `maxSize` comment

```yaml
# config/woodpecker.yaml:32
maxSize: 256000000 # Maximum segment size in bytes, default is 2GB. ...
```

256,000,000 bytes is 256 MB. The file writes sizes in decimal units throughout (`2000000` is
commented as "2MB"), so by its own convention this should read 256MB.

Three edits to one file, no code change.

tinswzy:
## Problem

`config/woodpecker.yaml` is what an operator reads to learn what is tunable. Two enforced settings
never appeared in it, and one comment stated a value roughly 8x off.

Fixes #316

## Changes

**`logstore.segmentCompactionPolicy.timeout`** — defined at `common/config/configuration.go:196`,
validated at `:812`, and enforced at `server/processor/segment_processor.go:641`, where it wraps the
entire `writer.Compact()` call:

```go
timeout := time.Duration(cfg.Woodpecker.Logstore.SegmentCompactionPolicy.Timeout.Seconds()) * time.Second
ctx, cancel := context.WithTimeout(ctx, timeout)
defer cancel()
```

Nothing in the shipped config revealed that a 5-minute compaction deadline exists, let alone how to
change it.

**`woodpecker.runtime`** — the whole section was absent. `opRegistry.capacity` and
`opRegistry.warnAge` are read at `cmd/main.go:203-204` to build the in-flight operation registry
that backs `wp ops stats` and the operation metrics, so in practice they were untunable. Placed
after `storage` to match the field order in `WoodpeckerConfig`.

**`segmentRollingPolicy.maxSize` comment** — the value is 256,000,000 bytes and the comment said
2GB. The file writes sizes in decimal units throughout (`2000000` is commented as "2MB"), so by its
own convention this reads 256MB.

Both new keys carry their existing in-code defaults, so no behaviour changes.

## Verification

Because the added keys carry the same values as their in-code defaults, a wrong indentation or
nesting level would be invisible: the key would be ignored, the default would apply, and every
assertion would still pass. Checked explicitly with a throwaway config setting them to 777 / 4242 /
99 — all three came through, confirming the keys are really parsed rather than falling back.

`TestNewConfiguration` now pins all three on both paths (YAML-loaded and defaults-only), matching
how it already covers every other field; without that the new YAML values would be the only
unpinned ones in the file.

- `go build ./...`: ok
- `go test ./common/config/...`: ok
- `golangci-lint run ./common/config/...` with the CI-pinned v2.11.4: 0 issues
- `gofmt -l .`: clean
- `go mod tidy`: no diff

### DIFF
--- a/common/config/configuration_test.go
+++ b/common/config/configuration_test.go
@@ -75,6 +75,7 @@ func TestNewConfiguration(t *testing.T) {
 	assert.Equal(t, int64(2000000), config.Woodpecker.Logstore.SegmentCompactionPolicy.MaxBytes.Int64())
 	assert.Equal(t, 4, config.Woodpecker.Logstore.SegmentCompactionPolicy.MaxParallelUploads)
 	assert.Equal(t, 8, config.Woodpecker.Logstore.SegmentCompactionPolicy.MaxParallelReads)
+	assert.Equal(t, 300, config.Woodpecker.Logstore.SegmentCompactionPolicy.Timeout.Seconds())
 	assert.Equal(t, int64(2000000), config.Woodpecker.Logstore.SegmentReadPolicy.MaxBatchSize.Int64())
 	assert.Equal(t, 32, config.Woodpecker.Logstore.SegmentReadPolicy.MaxFetchThreads)
 	assert.Equal(t, 259200, config.Woodpecker.Logstore.RetentionPolicy.TTL) // 72h = 259200s
@@ -92,6 +93,8 @@ func TestNewConfiguration(t *testing.T) {
 	assert.Equal(t, 1800, config.Woodpecker.Logstore.MaintenanceStrategy.ReconcileMinDataLogAge.Seconds())    // pull reconcile footer-HEAD age gate (30m)
 	assert.Equal(t, "minio", config.Woodpecker.Storage.Type)
 	assert.Equal(t, "/var/lib/woodpecker", config.Woodpecker.Storage.RootPath)
+	assert.Equal(t, 1024, config.Woodpecker.Runtime.OpRegistry.Capacity)
+	assert.Equal(t, 30, config.Woodpecker.Runtime.OpRegistry.WarnAge.Seconds())
 	assert.Equal(t, "info", config.Log.Level)
 	assert.Equal(t, "text", config.Log.Format)
 	assert.True(t, config.Log.Stdout)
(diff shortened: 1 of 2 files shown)


czs007:

This review found one operator-facing documentation issue in the newly added `woodpecker.runtime` configuration block.

## Medium

**`config/woodpecker.yaml:154` — `warnAge` comment states the wrong semantics**

The comment reads `Seconds after which a still-running operation is reported as stalled`, but `warnAge` is never used to flag running operations. It is read in exactly two places: `common/runtime/opregistry/registry.go:187` (`isOld := age >= r.warnAge`), inside `evictOldestLocked()`, whose only caller is `Register()` when `len(pool) >= capacity` (`registry.go:72-75`); and `Stats()` at `registry.go:159`, which merely echoes it back as `warn_age_ms`. Listing in-flight operations filters on `Filter.LongerThan` instead (`common/runtime/opregistry/types.go:63-77`). The in-code field comment is accurate — `common/config/configuration.go:434` says `// Young/old eviction boundary; default 30s` — so the new YAML text contradicts it.

Failure: an operator lowers `warnAge` from 30 to 5 to "surface stalls earlier" and the node then hits capacity 1024 under high throughput → every normal drop-oldest eviction is counted as `evictedOld`, so `wp ops stats` prints ` (STALL SIGNAL)` (`cmd/wpcli/cmd/ops_stats.go:55-58`) and the `old_evictions` red rule fires (`cmd/wpcli/cmd/metrics_report.go:51-54`) with no stall present; conversely, an operation hung for 5 minutes while in-flight count stays well below capacity triggers no eviction at all, `evicted_old` stays 0, and the operator concludes from the comment that nothing is stalled.

Suggestion: describe it as the eviction classification boundary, e.g. `Age boundary used when the registry is full and evicts its oldest operation: evictions of operations at least this old are counted as evicted_old (a stall signal), younger ones as evicted_young; default is 30s.`

(raised by bigsheeper, sijie-ni-0214)

PR #322, opened 2026-09-10
enhance(compaction): drop one payload copy and the pool around a contiguous read

### LINKED ISSUE #319: enhance(compaction): staged writer holds three copies of every merged block and pools a contiguous read
### Summary

`processMergeTask` in the staged writer holds three full copies of every merged block at peak and
allocates one goroutine pool per merge task to fragment what is a contiguous sequential read. Both
are avoidable without changing behaviour: the third copy exists only to hand `PutObject` a
contiguous buffer, and the read parallelism is spent on adjacent bytes of one local file.

### Three copies alive at once

```go
// ① one buffer per source block — server/storage/stagedstorage/writer_impl.go:2010
blockData := make([]byte, blockIndex.BlockSize)
file.ReadAt(blockData, blockIndex.StartOffset)

// the intermediate steps are all subslices, so ① is never released:
dataRecords := blockData[blockHeaderRecordSize:]     // parseLocalBlockForCompaction:1955
return dataRecords[:offset], ...                     // strictDataRecordPrefix:1991
allBlocks = append(allBlocks, extractedBlockData{dataRecords: cropped})

// ② the merged payload
mergedDataRecords = append(mergedDataRecords, blk.dataRecords...)

// ③ header + merged, purely to get one contiguous slice
completeBlockData = append(completeBlockData, codec.EncodeRecord(headerRecord)...)
completeBlockData = append(completeBlockData, codec.EncodeRecord(blockHeaderRecord)...)
completeBlockData = append(completeBlockData, mergedDataRecords...)
w.storageCli.PutObject(ctx, ..., bytes.NewReader(completeBlockData), int64(len(completeBlockData)), ...)
```

All three are live when `PutObject` runs: ① because `readFutures` and `allBlocks` still reference
it and Go keeps a backing array alive for any subslice into it, ② because it is in scope, ③ because
it is being uploaded.

③ carries no computation — it is a concatenation done solely so `bytes.NewReader` has something
contiguous. `ObjectStorage.PutObject` takes an `io.Reader`
(`common/objectstorage/object_storage.go:48`), so `io.MultiReader` over the header bytes, the block
header bytes and `mergedDataRecords` removes it outright. The size argument is still exactly
computable.

Peak per merge task goes from 3 x `segmentCompactionPolicy.maxBytes` to 2 x.

### A goroutine pool per merge task, over a contiguous read

```go
// server/storage/stagedstorage/writer_impl.go:1752
maxConcurrentReads := w.compactPolicyConfig.MaxParallelReads
readPool := conc.NewPool[*blockReadResult](maxConcurrentReads, conc.WithPreAlloc(true))
defer readPool.Release()
```

`planMergeBlockTasks` (`writer_impl.go:1682`) walks `w.blockIndexes` in order and groups
**consecutive** blocks until `targetBlockSize`, so one merge task's blocks are a single contiguous
byte range of the local `data.log`. Reading them through a pool turns that sequential range into N
concurrent `ReadAt` calls at adjacent offsets — it defeats readahead rather than helping, and the
data was written by this same writer minutes earlier so it is largely still in page cache, i.e.
these are memcpys and not device IO at all.

The cost is one `conc.NewPool(..., WithPreAlloc(true))` per merge task (a segment merged into 10
blocks builds 10 of them), N goroutines, and N separate allocations.

Replacing it with one `ReadAt` over `[first.StartOffset, last.StartOffset+last.BlockSize)` and
slicing per block keeps the existing per-block validation (`parseLocalBlockForCompaction` already
checks size, block number and entry range against the index, and verifies CRC) while collapsing ①
from N allocations to one.

`segmentCompactionPolicy.maxParallelReads` then has no consumer on this path.

### Net effect

| | before | after |
|---|---|---|
| peak copies per merge task | 3 x maxBytes | 2 x maxBytes |
| allocations per merge task | N + 2 | 2 |
| goroutine pools per compaction | 1 + (one per merge task) | 1 |
| read pattern | N concurrent `ReadAt` | one sequential `ReadAt` |

No behaviour change: same validation, same output object, same failure semantics.

### Why this is worth doing before adding any compaction concurrency limit

Compaction memory is `concurrent merge tasks x copies x maxBytes`. Any node-level limit has to be
sized against that constant, so landing the reduction first means the limit is calibrated against
2 rather than 3 — otherwise a third of the budget is permanently reserved for an overhead that
should not exist, and nobody goes back to re-tune it afterwards.

### Caveat to note in the PR

minio-go cannot retry a partially consumed non-seekable reader. With the shipped
`maxBytes` of 2 MB every merged block is far below the multipart threshold and goes out as a single
PUT, so a mid-transfer failure surfaces as an error and the auditor retries the segment on its next
cycle — acceptable, but it should be stated explicitly rather than discovered later.

tinswzy:
## Problem

`processMergeTask` in the staged writer held three copies of every merged block at peak and built
one goroutine pool per merge task to fragment what is a single sequential read of the local file.

Fixes #319

## One payload copy removed

The three copies were the source blocks, the merged payload, and a third buffer that was only the
first two concatenated so `bytes.NewReader` had something contiguous to hand `PutObject`. The
intermediate steps are all subslices, so the source buffers stay alive until the function returns
and all three are live during the upload.

Both leading records are fixed width (`RecordHeaderSize + HeaderRecordSize`,
`RecordHeaderSize + BlockHeaderRecordSize`), so their space is now reserved up front and the data
records are merged straight into the same buffer.

`io.MultiReader` over the pieces would also have removed the copy, but `PutObject` takes an
`io.Reader` and minio-go cannot retry a partially consumed non-seekable body — a transient failure
would have failed the whole compaction and made the auditor redo every merge task of that segment
on a later cycle. Reserving the prefix keeps a seekable `bytes.Reader` and the SDK-level retry.

## The read pool removed

`planMergeBlockTasks` groups a consecutive run of `blockIndexes`, so a merge task's blocks are one
contiguous range of `data.log`. Reading them through a pool turned that into N `ReadAt` calls at
adjacent offsets, each preceded by its own `os.Open`, against data this same writer had just
written and that is normally still in page cache — it fragmented a sequential read rather than
speeding one up.

`readMergeTaskBlocks` now reads the span once and returns a view per block. The span is derived
from the blocks themselves rather than assuming adjacency, so a gap would only cost a few extra
bytes read. `segmentCompactionPolicy.maxParallelReads` no longer has a consumer on this path.

## Net effect per merge task

| | before | after |
|---|---|---|
| payload copies at peak | 3 | 2 |
| allocations | N + 2 | 2 |
| `os.Open` calls | N | 1 |
| ants pools | 1 per task | none |

## Verification

The claim is no behaviour change, so I checked the bytes rather than only the tests. A throwaway
test captured every object compaction uploads through a mocked storage client, with
`segmentCompactionPolicy.maxBytes` and `maxFlushSize` lowered so the segment split into several
blocks and several merge tasks. Running it on this branch and again with the change stashed:

```
r/1/0/footer.blk -> d783065b94ceea05 len=299
r/1/0/m_0.blk    -> 24841a377369ddae len=350
r/1/0/m_1.blk    -> f3c2b4420d65a941 len=325
r/1/0/m_2.blk    -> 0daac62913760f37 len=325
r/1/0/m_3.blk    -> da4dcc98f340bef1 len=325
r/1/0/m_4.blk    -> f424aaad03fb274b len=325
r/1/0/m_5.blk    -> f16a13414d431d08 len=517
```

Identical both ways. `m_0` is 25 bytes larger than its siblings, which is exactly
`RecordHeaderSize + HeaderRecordSize` — the `HeaderRecord` only the first merged block carries, and
a direct check that the reserved prefix width is right.

The two tests that covered `readBlockDataFromLocalFile` now cover `readMergeTaskBlocks`; the
success case additionally reads the segment file independently and asserts each returned view
equals the bytes its index describes, which is what the per-block path used to return. A new test
pins the fixed-width assumption the reserved prefix depends on, so a codec change that broke it
fails loudly instead of leaving a gap of zero bytes between the headers and the payload. The
reserved width is also checked at runtime.

- `go build ./...`: ok
- `go test ./server/...`: 7 packages ok
- `go test -race ./server/storage/stagedstorage/...`: ok
- `golangci-lint run ./server/storage/stagedstorage/...` with the CI-pinned v2.11.4: 0 issues
- `gofmt -l .`: clean
- `go mod tidy`: no diff

### DIFF
--- a/server/storage/stagedstorage/writer_impl.go
+++ b/server/storage/stagedstorage/writer_impl.go
@@ -1599,13 +1599,6 @@ type mergeBlockTask struct {
 	nextEntryID int64                // Next entry ID after this merge block
 }
 
-// blockReadResult represents the result of reading a block from local file
-type blockReadResult struct {
-	blockIndex *codec.IndexRecord
-	blockData  []byte
-	error      error
-}
-
 // mergedBlockUploadResult represents the result of uploading a merged block
 type mergedBlockUploadResult struct {
 	blockIndex *codec.IndexRecord
@@ -1748,28 +1741,14 @@ func (w *StagedFileWriter) processMergeTask(ctx context.Context, task *mergeBloc
 		zap.Int64("mergedBlockID", mergedBlockID),
 		zap.Int("originalBlocks", len(task.blocks)))
 
-	// Create pool for concurrent block reading
-	maxConcurrentReads := w.compactPolicyConfig.MaxParallelReads
-	if maxConcurrentReads <= 0 {
-		maxConcurrentReads = min(8, len(task.blocks)) // Default to min(8, block count)
-	}
-
-	readPool := conc.NewPool[*blockReadResult](maxConcurrentReads, conc.WithPreAlloc(true))
-	defer readPool.Release()
-
-	// Submit all block read tasks to the pool
-	var readFutures []*conc.Future[*blockReadResult]
-	for _, blockIndex := range task.blocks {
-		if ctx.Err() != nil {
-			return &mergedBlockUploadResult{error: ctx.Err()}
-		}
-		// Capture variable for closure
-		blockIndexCopy := blockIndex
-
-		future := readPool.Submit(func() (*blockReadResult, error) {
-			return w.readBlockDataFromLocalFile(ctx, blockIndexCopy), nil
-		})
-		readFutures = append(readFutures, future)
+	// A merge task is planned from a consecutive run of blockIndexes, so its blocks occupy one
+	// contiguous range of the local file. Read that range once and view each block inside it,
+	// rather than issuing one ReadAt per block through a pool: the blocks are adjacent, the data
+	// was written by this same writer moments ago and is normally still in page cache, so the
+	// parallelism only fragmented a sequential read.
+	blockData, readErr := w.readMergeTaskBlocks(ctx, task)
+	if readErr != nil {
+		return &mergedBlockUploadResult{error: readErr}
 	}
 
 	// Collect block data and extract only DataRecords from each block
@@ -1781,21 +1760,15 @@ func (w *StagedFileWriter) processMergeTask(ctx context.Context, task *mergeBloc
 	firstEntryID := int64(-1)
 	lastEntryID := int64(-1)
 
-	for _, future := range readFutures {
+	for i, blockIndex := range task.blocks {
 		if ctx.Err() != nil {
 			return &mergedBlockUploadResult{error: ctx.Err()}
 		}
-		result := future.Value()
-		if result.error != nil {
-			return &mergedBlockUploadResult{
-				error: fmt.Errorf("failed to read block data: %w", result.error),
-			}
-		}
 
-		dataRecords, blockLastEntryID, extractErr := w.extractCompactedDataRecords(result.blockIndex, result.blockData, expectedLastEntryId)
+		dataRecords, blockLastEntryID, extractErr := w.extractCompactedDataRecords(blockIndex, blockData[i], expectedLastEntryId)
 		if extractErr != nil {
 			return &mergedBlockUploadResult{
-				error: fmt.Errorf("failed to extract data records from block %d: %w", result.blockIndex.BlockNumber, extractErr),
+				error: fmt.Errorf("failed to extract data records from block %d: %w", blockIndex.BlockNumber, extractErr),
 			}
 		}
 		if len(dataRecords) == 0 {
@@ -1803,13 +1776,13 @@ func (w *StagedFileWriter) processMergeTask(ctx context.Context, task *mergeBloc
 		}
 
 		allBlocks = append(allBlocks, extractedBlockData{
-			blockIndex:  result.blockIndex,
+			blockIndex:  blockIndex,
 			dataRecords: dataRecords,
 		})
 
 		// Track entry ID range
-		if firstEntryID == -1 || result.blockIndex.FirstEntryID < firstEntryID {
-			firstEntryID = result.blockIndex.FirstEntryID
+		if firstEntryID == -1 || blockIndex.FirstEntryID < firstEntryID {
+			firstEntryID = blockIndex.FirstEntryID
 		}
 		if lastEntryID == -1 || blockLastEntryID > lastEntryID {
 			lastEntryID = blockLastEntryID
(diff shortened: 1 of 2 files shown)


czs007:

**Verified:**
- `server/storage/stagedstorage/writer_impl.go:2023` — each block view is sliced at `StartOffset - spanStart` with length `BlockSize`, so the single-span read returns byte-identical content to the old per-block `ReadAt`, with offsets provably in range.
- `server/storage/stagedstorage/writer_impl.go:1820` — the payload is copied in sorted `allBlocks` order after the sort at `:1796`, so the merged data region matches the concatenation the old append loop produced.
- `server/storage/stagedstorage/writer_impl.go:1829` — `BlockLength`/`BlockCrc` are computed over `completeBlockData[prefixLen:]`, which does not overlap the header prefix, so both values equal the old ones.
- `server/storage/stagedstorage/writer_impl.go:1844` — the reserved prefix relies on fixed-size `HeaderRecord`/`BlockHeaderRecord` encodings (`codec.go:35,53`), and the width guard returns an error before upload in either direction if a codec change breaks that assumption.
- `server/storage/stagedstorage/writer_impl.go:1644` — `processMergeTask` still runs only under `uploadPool`; each task opens the segment file read-only with its own span buffer and the views are never mutated, so removing the read pool introduces no shared mutable state.
- `server/storage/stagedstorage/writer_impl_test.go:270` — the success test independently re-reads the file and asserts each view equals `raw[StartOffset:StartOffset+BlockSize]`, which is exactly the contract the refactor must preserve.

## Nitpick — non-blocking

- **`server/storage/stagedstorage/writer_impl_test.go:1261`** — The stagedstorage tests still assign `SegmentCompactionPolicy.MaxParallelReads` in seven places (`:1261`, `:1342`, `:1383`, `:1476`, `:1579`, `:1690`, `:1785`), but this PR removed the only consumer in the package (`maxConcurrentReads := w.compactPolicyConfig.MaxParallelReads` and its `conc.Pool`), leaving `objectstorage/writer_impl.go:1982` as the sole remaining reader repo-wide. Since most of these sit next to a `MaxParallelUploads = 4/1` assignment, a reader can easily conclude that staged compaction read concurrency is still pinned by these tests. Suggestion: drop the seven now-inert lines, or keep one and add a short comment noting that only the objectstorage path consumes the field. Inert either way — no behavioral impact. (raised by bigsheeper, xiaocai2333)

Issue #318, opened 2026-09-10
enhance(log): a compaction pass has no deadline — one stalled object store halts a log's entire auditor maintenance


tinswzy:
### Summary

A single compaction is bounded server-side, but nothing bounds a compaction *pass*. The auditor
walks every `Completed` segment sequentially, each segment can burn up to
`segmentCompactionPolicy.timeout` on each of its quorum nodes in turn, and the client sets no
deadline of its own. With an object store that hangs rather than fails fast, one auditor cycle for
one log can run for hours, and everything else that log's auditor does is stalled behind it.

### What is already bounded

`server/processor/segment_processor.go:641` wraps the whole `writer.Compact()` call:

```go
timeout := time.Duration(cfg.Woodpecker.Logstore.SegmentCompactionPolicy.Timeout.Seconds()) * time.Second
ctx, cancel := context.WithTimeout(ctx, timeout)   // default 300s
defer cancel()
```

So one `CompactSegment` RPC returns within ~300s. That part is fine.

### What is not

**1. The client sets no deadline.** `woodpecker/client/logstore_client_remote.go:334` passes the
caller's context straight into `CompactSegmentWithExpected`, and that context comes from
`log_writer.go:358`:

```go
ctx, sp := logger.NewIntentCtxWithParent(l.maintenanceCtx, WriterScopeName, ...)
```

`maintenanceCtx` lives as long as the writer, so it carries no deadline. The client's own bound is
therefore whatever the server happens to enforce — it has no independent one, and callers that
reach a server without the timeout configured (or a future path that skips it) get no bound at all.

**2. Quorum retry multiplies the server bound.** `compactSegmentQuorum`
(`woodpecker/segment/segment_handle.go:1647`) tries each quorum node **sequentially** until one
succeeds. Three replicas that each hang until the server deadline is 3 x 300s = **900s for one
segment**.

**3. The pass over segments is unbounded.** `compactCompletedSegments` (`log_auditor.go:55`) loops
over every `Completed` segment in the snapshot with no cap and no pass-level deadline. N segments
in the backlog is N x up-to-900s.

### Consequence

`runAuditor` is one goroutine per log doing, in order per cycle: truncate-state check, segment
snapshot load, snapshot publication to the notify distributor, `compactCompletedSegments`, then the
periodic orphan sweep. A compaction pass that runs long delays the orphan sweep in that cycle and
everything in every subsequent cycle — truncate state stops advancing, no fresh snapshot reaches
[...]

PR #323, opened 2026-09-10
enhance(log): jitter the auditor's first tick and bound the compaction pass

### LINKED ISSUE #317: enhance(log): auditor tickers have no jitter and the compaction pass has no per-cycle bound, so logs compact in synchronised bursts
### Summary

Every log writer runs its own auditor goroutine on a bare `time.NewTicker(auditorMaxInterval)` with
no jitter, and the compaction pass inside that cycle has no per-cycle bound. Writers created
together therefore fire together, forever, and each firing can try to compact an unbounded number
of segments at once. Both halves push a synchronised burst of `CompactSegment` RPCs at the
woodpecker nodes.

### No jitter

```go
// woodpecker/log/log_writer.go:342  and  woodpecker/log/internal_log_writer.go:265
ticker := time.NewTicker(time.Duration(l.auditorMaxInterval * int(time.Second)))
```

One goroutine per log writer, fixed period, phase set by creation time and never perturbed. Any
event that opens many writers at once — a StreamingNode restart, a WAL failover, a pchannel
rebalance — aligns all of their tickers permanently. At a few thousand logs that is a few thousand
auditors waking in the same millisecond every `auditorMaxInterval`.

`segmentRollingPolicy.maxInterval` is likewise a fixed period, so logs opened together also seal
segments together. The two alignments compound: a synchronised wave of segments enters `Completed`,
and the next synchronised auditor tick tries to compact all of them.

### No per-cycle bound on compaction

The same file bounds the notify path deliberately:

```go
// woodpecker/log/log_auditor.go:43
const maxCompactedNotifyPerCycle = 64
```

with a comment explaining the intent — that a large backlog, "e.g. the first cycles after upgrading
a cluster with many pre-existing sealed segments", must not swamp a single cycle. But the same
comment notes compaction is *not* counted, and `compactCompletedSegments` (`log_auditor.go:55`)
walks every `Completed` segment in the snapshot with no cap:

```go
for _, seg := range segs {
    if seg.Metadata.State != proto.SegmentState_Completed { continue }
    ...
    recoverySegmentHandle.Compact(ctx)
}
```

Compaction is far more expensive per unit than a notify — a merge plus object-storage uploads
versus one etcd write — so if either pass deserves the bound it is this one.

The exposure is worst exactly when recovering: while compaction is failing (object storage
unwritable, wrong credentials, a version mismatch), `Completed` segments accumulate. When it
recovers, every log's auditor tries to drain its whole backlog in one cycle, all at the same
instant.

### Why this is worth fixing on the client even though the node should also defend itself

A node-side concurrency limit is the right backstop and is worth having independently. But without
jitter, thousands of auditors still fire simultaneously and burn a full round of RPCs to be told
"busy"; shaping the peak at the source keeps that limit from being the thing that runs hot every
cycle.

### Suggested direction

- Add jitter to both `runAuditor` tickers. Either a randomised initial phase or a per-tick
  perturbation of a few percent is enough to decorrelate writers created together.
- Give `compactCompletedSegments` a per-cycle bound, mirroring `maxCompactedNotifyPerCycle` — the
  remainder is naturally picked up on the following cycle, exactly as the notify path already does.
- Consider the same treatment for `segmentRollingPolicy.maxInterval`, which is the other fixed
  period that keeps logs in lockstep.

Note: #313 is currently changing `compactCompletedSegments` and the surrounding auditor code, so
this is probably best applied after it lands.

tinswzy:
## Problem

Every log writer runs its own auditor on a bare `time.NewTicker(auditorMaxInterval)`, and the
compaction pass inside that cycle had no per-cycle bound. Both push a synchronised burst of
`CompactSegment` RPCs at the woodpecker nodes.

Fixes #317

## Spread the auditor start

A ticker's phase is set by when the writer was created, so anything that opens many writers at once
— a node restart, a WAL failover, a channel rebalance — aligns them, and periodic tickers never
drift back apart. At a few thousand logs that is a few thousand auditors waking in the same
millisecond every interval. `segmentRollingPolicy.maxInterval` is a fixed period too, so logs opened
together also seal segments together and the two alignments compound: a synchronised wave of
segments enters `Completed`, and the next synchronised tick tries to compact all of them.

`runAuditor` now waits a random `[0, interval)` before starting its ticker. The ticker inherits that
offset, so the writers are decorrelated once for the process lifetime: two opened together stay
`d₁ - d₂` apart for as long as the process runs.

Two notes on the shape, both from review on this PR — an earlier revision jittered the *first tick*
instead, by swapping the loop's channel from a one-shot timer to the ticker:

- **The first cycle now lands in `[interval, 2×interval)`** rather than at exactly `interval`. The
  phase spread is identical either way, so "never postpone a cycle" was a constraint we had imposed
  on ourselves, not one the goal implies. Everything the auditor does — compaction, compacted-mark
  distribution, truncated-segment cleanup, the orphan sweep — is background reclamation with no
  latency deadline, and at the default 10s interval the worst case is a first cycle 20s after the
  writer opens. Accepting that leaves the loop exactly as it was: no timer/ticker channel swap, and
  no `if ticker == nil` check sitting in the tick case for a property that only matters once. It
  also removes any chance of a cycle firing at `t≈0`, in the middle of the same startup storm that
  opened the writers.
- **The wait watches `writerClose` and `maintenanceCtx.Done()`**, the same two signals `runAuditor`
  exits on, so a writer closed during it is not held up. That is what the loop-embedded version was
  reaching for, and what a bare `time.Sleep` would have missed;
  `TestLogWriter_RunAuditor_StopsOnWriterClose` covers it.

## Bound the compaction pass

`distributeCompactedMarks` has stopped a backlog from swamping a single cycle since
`maxCompactedNotifyPerCycle` was added, and the comment on that constant notes compaction is not
counted. `compactCompletedSegments` walked every `Completed` segment with no cap, though compaction
is the far more expensive pass — a merge plus object-storage uploads, against one etcd write.

The exposure is worst exactly while recovering: whenever compaction has been failing (object storage
unwritable, wrong credentials, a version mismatch) `Completed` segments pile up, and the moment
storage returns every log tries to drain its whole backlog at once.

A cycle now compacts at most `maxCompactedPerCycle` (64, mirroring the notify path) and reports the
rest as `segmentsDeferred` in the cycle summary — which is also the signal an operator wants when
compaction is falling behind.

**Ordering came with the bound, not as a separate change.** Ranging a map is randomly ordered. That
was harmless while every `Completed` segment was visited each cycle, but under a bound it would let
a segment be skipped cycle after cycle by chance while its local `data.log` stayed on disk. The pass
now walks in ascending segment order, which also drains the data that has occupied local disk
longest.

## Note on what this does not fix

The bound caps how many segments a cycle attempts, not how long the cycle runs: a segment can still
take up to `segmentCompactionPolicy.timeout` on each of its quorum nodes in turn, so a bounded pass
against a hanging object store can still outlast the interval. That is #318, and the two are
complementary — this one shapes the peak, that one bounds the duration.

## Verification

- `go test ./woodpecker/log/`: ok
- `go test -race ./woodpecker/log/ -run 'TestWaitAuditorStartJitter|RunAuditor'`: ok
- `golangci-lint run ./woodpecker/log/...` with the CI-pinned v2.11.4: 0 issues
- `gofmt -l .`: clean
- `go vet ./woodpecker/log/`: clean

New tests cover the bound (only `maxCompactedPerCycle` attempted, remainder reported deferred), the
ordering (oldest first, with the map deliberately populated in descending order so a map-order pass
would be very unlikely to come out sorted), and the start jitter (spread across the interval, plus
both shutdown signals returning immediately instead of waiting the delay out).

The nine `RunAuditor` tests waited 1500ms against a 1s interval, so with the first cycle pushed past
one interval they became coin flips rather than outright failures. They now wait past twice the
interval, which is when a cycle has certainly run.

### DIFF
--- a/woodpecker/log/internal_log_writer.go
+++ b/woodpecker/log/internal_log_writer.go
@@ -262,8 +262,7 @@ func (l *internalLogWriterImpl) WriteAsync(ctx context.Context, msg *WriteMessag
 }
 
 func (l *internalLogWriterImpl) runAuditor() {
-	ticker := time.NewTicker(time.Duration(l.auditorMaxInterval * int(time.Second)))
-	defer ticker.Stop()
+	interval := time.Duration(l.auditorMaxInterval) * time.Second
 
 	// Storage mode is fixed for the writer's lifetime; read it once here rather than per
 	// segment. localMode disables the compaction pass (see compactCompletedSegments), so
@@ -276,10 +275,26 @@ func (l *internalLogWriterImpl) runAuditor() {
 		zap.Int("intervalSeconds", l.auditorMaxInterval),
 		zap.Bool("localStorageCompactionDisabled", localMode))
 
+	// The first tick is jittered so writers created together do not tick in lockstep; the
+	// ticker takes over at the full interval from there. See auditorFirstTickDelay.
+	firstTick := time.NewTimer(auditorFirstTickDelay(interval))
+	defer firstTick.Stop()
+	tickC := firstTick.C
+	var ticker *time.Ticker
+	defer func() {
+		if ticker != nil {
+			ticker.Stop()
+		}
+	}()
+
 	auditCycle := uint64(0)
 	for {
 		select {
-		case <-ticker.C:
+		case <-tickC:
+			if ticker == nil {
+				ticker = time.NewTicker(interval)
+				tickC = ticker.C
+			}
 			if l.maintenanceCtx.Err() != nil || !l.isWriterValid.Load() {
 				l.stopMaintenance()
 				return
@@ -346,6 +361,7 @@ func (l *internalLogWriterImpl) runAuditor() {
 				zap.Int("segmentsProcessed", cs.processed),
 				zap.Int("segmentsCompacted", cs.compacted),
 				zap.Int("segmentsFailed", cs.failed),
+				zap.Int("segmentsDeferred", cs.deferred),
 				zap.Int("truncatedSegments", len(truncatedSegmentExists)))
 
 			// Clean up truncated segments (object-storage data + local files + tombstones).
--- a/woodpecker/log/internal_log_writer.go
+++ b/woodpecker/log/internal_log_writer.go
@@ -275,26 +275,18 @@ func (l *internalLogWriterImpl) runAuditor() {
 		zap.Int("intervalSeconds", l.auditorMaxInterval),
 		zap.Bool("localStorageCompactionDisabled", localMode))
 
-	// The first tick is jittered so writers created together do not tick in lockstep; the
-	// ticker takes over at the full interval from there. See auditorFirstTickDelay.
-	firstTick := time.NewTimer(auditorFirstTickDelay(interval))
-	defer firstTick.Stop()
-	tickC := firstTick.C
-	var ticker *time.Ticker
-	defer func() {
-		if ticker != nil {
-			ticker.Stop()
-		}
-	}()
+	// Spread the start so writers created together do not tick in lockstep.
+	if !waitAuditorStartJitter(l.maintenanceCtx, l.writerClose, interval) {
+		return
+	}
+
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
 
 	auditCycle := uint64(0)
 	for {
 		select {
-		case <-tickC:
-			if ticker == nil {
-				ticker = time.NewTicker(interval)
-				tickC = ticker.C
-			}
+		case <-ticker.C:
 			if l.maintenanceCtx.Err() != nil || !l.isWriterValid.Load() {
 				l.stopMaintenance()
 				return
--- a/woodpecker/log/log_auditor.go
+++ b/woodpecker/log/log_auditor.go
@@ -20,6 +20,9 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"math/rand/v2"
+	"sort"
+	"time"
 
 	"go.uber.org/zap"
 
@@ -42,11 +45,41 @@ import (
 // RPC fanouts. The remainder is picked up on subsequent cycles.
 const maxCompactedNotifyPerCycle = 64
 
+// maxCompactedPerCycle bounds how many Completed segments one auditor cycle compacts. The
+// remainder is picked up on subsequent cycles, mirroring maxCompactedNotifyPerCycle above.
+//
+// Compaction is the more expensive of the two passes by a wide margin -- a merge plus
+// object-storage uploads, against one etcd write -- so a backlog swamps a cycle harder here. The
+// exposure is worst exactly while recovering: whenever compaction has been failing (object storage
+// unwritable, wrong credentials, a version mismatch) Completed segments pile up, and the moment it
+// recovers every log tries to drain its whole backlog at once.
+const maxCompactedPerCycle = 64
+
+// auditorFirstTickDelay returns a random delay in [0, interval) for an auditor's first tick.
+//
+// Every log writer runs its own auditor on a fixed period whose phase is set by when the writer was
+// created, so anything that opens many writers at once -- a node restart, a WAL failover, a channel
+// rebalance -- aligns their tickers, and periodic tickers never drift back apart. At a few thousand
+// logs that is a few thousand auditors waking in the same millisecond every interval, each fanning
+// compaction RPCs at the same nodes. segmentRollingPolicy.maxInterval is a fixed period too, so
+// logs opened together also seal segments together and the two alignments compound: a synchronised
+// wave of segments enters Completed, and the next synchronised tick tries to compact all of them.
+//
+// Jittering only the first tick decorrelates the writers once, for the process lifetime, and the
+// delay is shorter than the interval the first tick used to wait, so no cycle is postponed.
+func auditorFirstTickDelay(interval time.Duration) time.Duration {
+	if interval <= 0 {
+		return 0
+	}
+	return time.Duration(rand.Int64N(int64(interval)))
+}
+
 // compactStats summarizes one compactCompletedSegments pass for the auditor cycle log.
 type compactStats struct {
 	processed int
 	compacted int
 	failed    int
+	deferred  int // Completed segments left for a later cycle by maxCompactedPerCycle
 }
 
 // compactCompletedSegments compacts every Completed segment in the snapshot, sequentially by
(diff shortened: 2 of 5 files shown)


czs007:

**Verified:**
- `woodpecker/segment/segment_handle.go:1654-1709` — `compactSegmentQuorum` walks the quorum nodes sequentially and returns on the first success, so a single failing replica cannot stall a segment's compaction.
- `woodpecker/log/log_writer.go:388` — `CheckAndSetSegmentTruncatedIfNeed` runs ahead of `compactCompletedSegments` in every auditor cycle, so truncation gets first chance to retire stuck candidates.
- `woodpecker/log/log_handle.go:929-959` — Completed segments below the truncation point are moved to `Truncated`, removing them from the compaction candidate set.
- `woodpecker/segment/segment_handle.go:1939-1967` — `Compact` re-reads state under `RefreshAndGetMetadata` from the same round's snapshot, and a lost `doingCompact` CAS returns `nil` transiently rather than wedging the segment.
- `woodpecker/log/log_auditor.go:111-115` — the new ascending sort plus per-cycle cap is deterministic, and the jitter timer→ticker change and its exit paths were checked along with existing test coverage for flakiness.

## Low — non-blocking

- **`woodpecker/log/log_auditor.go:112` — head-of-line blocking can starve later segments.** The cycle now sorts Completed segments ascending and truncates to `maxCompactedPerCycle` (64), so segments that keep failing compaction are always the lowest-numbered ones and re-occupy the same budget slots every cycle. Failure: a log that has never been truncated (`TruncatedSegmentId <= 0`, so `CheckAndSetSegmentTruncatedIfNeed` returns early at `woodpecker/log/log_handle.go:906`) accumulates more than 64 Completed segments whose 64 lowest-numbered ones fail on every quorum replica — e.g. all replicas return `ErrSegmentCompactionDataBehind` after multi-replica local data loss → the entire 64-slot budget is spent re-attempting the same stuck head each cycle, segments 65 and above are never passed to `Compact`, stay Completed indefinitely, their staged local `data.log` is never reclaimed on the quorum nodes, and `segmentsDeferred` grows without bound. Note that the truncation escape hatch only bounds starvation for logs that have a truncation point; for never-truncated logs it never fires. The trigger is narrow — it requires all three replicas of the head segments to fail — so a rotating cursor is not required, but consider documenting the precondition or making the budget skip segments that failed in the immediately preceding cycle. (raised by sijie-ni-0214, bigsheeper)

- **`woodpecker/log/log_auditor.go:49` — the `maxCompactedPerCycle` comment claims parity with `maxCompactedNotifyPerCycle`, but the two paths differ.** The comment states the remainder is "picked up on subsequent cycles, mirroring maxCompactedNotifyPerCycle above", yet the notify path has a terminal exit the compaction path lacks: `distributeCompactedMarks` only consumes budget when `EnsureSegmentNotified` reports `advanced` (`woodpecker/log/log_auditor.go:165-168`), and a record that reaches a settled state (`isSettledNotifyState`, `woodpecker/segment/segment_compacted_notify_manager.go:116-125`) returns `false` at `:191-193` and then hits the in-memory `settled` fast path at `:137-140`, so it stops consuming budget forever. Compaction has no settled or terminal state — a segment leaves the candidate set only by compacting successfully or by being truncated, and truncation is itself gated on a truncation point existing (`woodpecker/log/log_handle.go:906`). Suggestion: reword the comment to state that precondition instead of claiming parity, and optionally add a multi-round test covering the remainder path. (raised by bigsheeper)

tinswzy:
Agreed — simplified in 842c46f. The loop is back to exactly what it was:

```diff
-	firstTick := time.NewTimer(auditorFirstTickDelay(interval))
-	defer firstTick.Stop()
-	tickC := firstTick.C
-	var ticker *time.Ticker
-	defer func() {
-		if ticker != nil { ticker.Stop() }
-	}()
+	if !waitAuditorStartJitter(l.maintenanceCtx, l.writerClose, interval) {
+		return
+	}
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()

-		case <-tickC:
-			if ticker == nil {
-				ticker = time.NewTicker(interval)
-				tickC = ticker.C
-			}
+		case <-ticker.C:
```

Looking back at why I wrote it the other way, I had two reasons and only one of them held.

The first was that waiting before starting the ticker pushes the first cycle out to between one and
two intervals rather than at one. That is true, and it is also not important — you are right that a
startup delay of up to one interval is entirely reasonable for background maintenance on a writer
that has only just opened.

The second was that a bare sleep would not honour the shutdown signals, so a writer closed during
it would be held up; `TestLogWriter_RunAuditor_StopsOnWriterClose` sets a 60s interval precisely to
catch that. That one is real, but it only calls for the wait itself to be interruptible — I jumped
from "must watch writerClose" to "put the delay in the loop's select" and skipped the nearer answer.
`waitAuditorStartJitter` now selects over the timer, `writerClose` and `maintenanceCtx`, with a test
case for each.

One cost worth flagging rather than burying: pushing the first cycle past one interval broke nine
`RunAuditor` tests, which waited 1500ms against a 1s interval. They did not fail outright — they
became coin flips, passing and failing within the same run. They now wait past twice the interval,
which is when a cycle has certainly run. Six consecutive runs are clean, and the reason is in a
comment beside the wait so the next person changing the interval knows the relation.

`go test` and `go test -race` on `woodpecker/log`: ok. `golangci-lint` (CI-pinned v2.11.4): 0 issues.

czs007:

**Verified:**
- `woodpecker/log/log_auditor.go:73` — `waitAuditorStartJitter` returns early when `interval <= 0`, so `rand.Int64N` never gets a non-positive argument. The timer is released with `defer Stop`, and the three-way `select` uses the same exit signals as the main auditor loop (`log_writer.go:453-462`).
- `log_writer.go:705` / `internal_log_writer.go:626` — Close runs `stopMaintenance()`, then sends on the buffered `writerClose` (cap 1), then calls `close(writerClose)`, then `maintenanceWG.Wait()`. If Close runs during the jitter wait, the auditor sees either the sent value or the closed channel, then returns and calls `Done()`. `monitorSession` still gets its signal from the close, as it did before this change.
- `log_writer.go:162` — Session expiry only calls `stopMaintenance()`, which cancels the context. The jitter's `ctx.Done()` branch handles that case.
- `woodpecker/log/log_auditor.go:117` — Completed segments are filtered first, then sorted ascending, then capped at 64. `deferred` is set only when the cap is exceeded. If the loop exits on `ctx.Err()`, the caller returns early (`log_writer.go:412-414`) and doesn't report partial counts.
- `woodpecker/log/log_auditor.go:214` — `runNotifyDistributor` has no ticker of its own; it runs only on auditor snapshots. The first snapshot moves from `interval` to `[interval, 2×interval)`, and background reclamation doesn't depend on that timing.
- `log_writer_test.go:1614-1856` — The 9 tests whose wait changed to 2500ms don't use `.Once()` or `AssertNumberOfCalls` on their mocks, so a second audit cycle inside that window can't break their assertions. The unmodified tests that call `runAuditor` synchronously use `interval=1s`, so the first cycle runs by 2s, inside their 3s bound.

These checks come from reading the code. Reviewers did not run `go test` locally, so CI results are the final word.

PR #324, opened 2026-09-10
enhance(log): bound the compaction pass and each attempt against a node

### LINKED ISSUE #318: enhance(log): a compaction pass has no deadline — one stalled object store halts a log's entire auditor maintenance

tinswzy:
## Problem

A single compaction is bounded server-side, but nothing bounded a compaction *pass*. The auditor
walked every `Completed` segment in sequence, each segment could burn
`segmentCompactionPolicy.timeout` on each of its quorum nodes in turn, and the client set no
deadline of its own.

Fixes #318

Against an object store that hangs rather than failing, one auditor cycle for one log could run for
hours. The auditor runs its passes in sequence, so truncate state, snapshot publication and the
orphan sweep were all stalled behind it — and that is what stops local `data.log` from being
reclaimed, so the failure modes compound in the direction that fills the disk.

## Two bounds

**`client.auditor.compactionAttemptTimeout`** (330s) bounds one attempt against one node. gRPC
carries the deadline to the node, so a node that hangs now costs one attempt rather than the
caller's whole budget, and `compactSegmentQuorum` reaches the remaining replicas — which is the
reason that walk is sequential. Previously a hung first node meant the segment never reached its
healthy replicas, on that cycle or any later one.

**`client.auditor.compactionPassBudget`** (60s) bounds how long one auditor cycle spends *starting*
compactions. Once it is spent the pass starts no further segment; segments it did not reach stay
`Completed` and are reported as `segmentsDeferred` for a later cycle.

## Why the attempt default is 330s and not lower

The client deadline reaches the node over gRPC, so the effective budget is the min of the two, and
**nothing of a cut-off attempt is kept** — unless a complete compacted footer already exists, the
writer restarts from the first block. A client value below the server's own per-segment budget
(`logstore.segmentCompactionPolicy.timeout`, 300s) would therefore fail every compaction that
legitimately needs longer, on every replica, every cycle, and never reclaim the local `data.log` —
a regression in exactly the direction this change sets out to fix, and not something a Milvus
deployment could work around, since `setCustomWpConfig` overrides only `Auditor.MaxInterval`.

330s is the server default plus RPC margin, and cross-field validation rejects a value below the
local `segmentCompactionPolicy.timeout` so the two cannot drift apart.

## Why the pass bound is a budget, not a deadline

An earlier shape of this PR made the pass a `context.WithTimeout`. That cuts off a compaction in
flight, which is the same trap as the attempt timeout above: any segment slower than the budget
dies every cycle and never finishes. It also forced the pass bound above
`attemptTimeout × quorum size` — 990s for three replicas, 1650s for five — which left it too large
to be worth much.

As a budget, the segment already running finishes. A pass lasts at most the budget plus one
segment, needs no relation to the attempt timeout or the quorum size, and cannot cut off a segment
that is legitimately slow.

**What "plus one segment" costs, concretely.** `compactSegmentQuorum` walks the quorum
sequentially, one attempt per node, so the worst case for a single segment is every replica hanging
until its own deadline:

```
one segment, worst case = Wq × compactionAttemptTimeout = 3 × 330s = 990s
one pass,    worst case = 60s (budget) + 990s ≈ 17.5 min
```

That is bounded where it previously was not — the same pass could run for hours against an object
store that hangs rather than failing — but 17.5 minutes is the number to judge, not "budget plus one
segment". Tightening it further would mean either lowering the attempt timeout (not available: below
the server's 300s is the regression described above) or capping how many replicas one segment may
try per pass. Neither belongs in this PR; going from unbounded to bounded is the change worth
making on its own.

Note that "attempt" here is one RPC to one node, not a retry: each replica is tried at most once per
pass, and the retry is the next auditor cycle, as before. Nothing on this path wraps compaction in
`retry.Do`.

## Relationship to #323 (now on master)

The two bounds compose rather than overlap: `maxCompactedPerCycle` caps **how many** segments a
cycle may start, this budget caps **how long** it may spend starting them, and both report their
remainder as `deferred`. Neither subsumes the other — 64 fast segments finish well inside the
budget, and one slow segment exhausts the budget long before the count. `deferred` is now the sum
of both causes, which `TestCompactCompletedSegments_CountBoundAndBudgetCompose` pins.

## Two smaller fixes on the same path

- `compactSegmentQuorum` returns the cancellation instead of walking on. The pool hands back a
  cached client without consulting the context, and `WithTimeout` on a done context yields an
  immediately-expired child, so the remaining nodes would "fail" instantly, land in `nodeFailures`,
  and be counted into `WpSegmentCompactionFailuresTotal` and the all-nodes-failed warning —
  replicas that were never contacted, in the signal used to find a broken one.
- `classifyCompactionFailureReason` now also matches `codes.DeadlineExceeded`. A gRPC status error
  does not wrap `context.DeadlineExceeded`, is not a `woodpeckerError`, and is not
  `codes.Unavailable`, so a remote attempt deadline was labelled `"other"` in exactly the
  deployment where it is what detects a hung node.

## A non-positive value means unbounded

On both settings. That is a guard for a `Configuration` assembled programmatically — several
existing tests build an `AuditorConfig` by hand — not a user-facing escape hatch. `NewConfiguration`
always supplies positive defaults and validation stays strict.

## Verification

- `go test ./common/config/ ./woodpecker/log/ ./woodpecker/segment/`: ok
- `golangci-lint run ./woodpecker/... ./common/...` with the CI-pinned v2.11.4: 0 issues

New tests:

- `TestCompactSegmentQuorum_AttemptDeadlineFailsOverToNextNode` — a hung node and a healthy one,
  with the caller's context carrying no deadline at all (as the auditor's does not). Asserts the
  attempt has a deadline of its own and that the walk reaches the second replica. It finishes in
  almost exactly the attempt timeout, so the bound is what ended the first attempt.
- `TestCompactSegmentQuorum_CancelledCallerDoesNotBlameUntriedNodes` — a cancelled caller must not
  produce failure records for replicas that were never contacted.
- `TestClassifyCompactionFailureReason_GrpcDeadlineIsTransient` — a gRPC `DeadlineExceeded` status
  classifies as a timeout, not `"other"`.
- `TestCompactCompletedSegments_BudgetStopsStartingButNeverCutsOffInFlight` — the pass stops
  starting segments but the one in flight runs to completion and is not cancelled.
- `TestCompactCompletedSegments_ZeroBudgetIsUnbounded` — a non-positive budget must not mean "stop
  before the first segment", which would silently disable compaction.
- `TestCompactCompletedSegments_CountBoundAndBudgetCompose` — both bounds fire in one pass and
  every segment not started is reported as deferred.
- `TestValidate_AttemptTimeoutMustCoverServerCompactionBudget` — the cross-field rule.
- Config: both keys asserted on the YAML and defaults paths, both validation errors covered. Since
  the shipped values equal the in-code defaults, a wrong indentation would be invisible — checked
  with a throwaway config setting them to 777/888, and both came through.

Two hand-built `AuditorConfig` fixtures in the config tests needed the new fields. Validation
rejecting a zero timeout matches how `MaxInterval` is already treated, so completing the fixtures
is the consistent fix.


czs007:

This PR adds client-side attempt and pass deadlines to the auditor compaction path; the deadline plumbing itself is sound, but the default attempt budget sits below the server's own per-segment compaction budget, and a few config-validation and failure-classification edges around the new deadlines need attention.

## High

**`common/config/configuration.go:921` — `compactionAttemptTimeout` default (60s) is five times below the server's `segmentCompactionPolicy.timeout` (300s), so a slow compaction can never finish on any replica**

The new client-side attempt timeout defaults to 60s (`common/config/configuration.go:921`, `config/woodpecker.yaml:37`), while the server budget per segment is 300s (`config/woodpecker.yaml:108`; `server/processor/segment_processor.go:642-644` wraps the inherited gRPC request context in `context.WithTimeout(ctx, 300s)`). The attempt context propagates to the server over gRPC (`woodpecker/client/logstore_client_remote.go:334`; the embed path at `logstore_client_local.go:74-76` passes it straight through), so the effective per-attempt budget is `min(60s, 300s)` = 60s and the configured 300s read/merge/upload budget is unreachable. A cut-off attempt keeps no progress: `StagedFileWriter.Compact` short-circuits only when a *complete* compacted footer already exists (`server/storage/stagedstorage/writer_impl.go:1409` → `compactRemoteFooterIfPresent:1470-1493`, `handled=false` when the footer is missing); otherwise it re-runs `readLocalFileAndUploadToMinio` from the first block (`:1432`), which returns on context cancellation at `:1644`/`:1662` without reusing already-uploaded merged blocks.

Failure scenario: server-side compaction of a `Completed` segment takes between 60s and 300s — e.g. a 256MB segment (`config/woodpecker.yaml:32`) on a throttled or high-latency object store → every quorum node's attempt is cut at 60s with no retained merged blocks, `compactSegmentQuorum` fails on all nodes, the segment stays `Completed`, each auditor cycle repeats the same 3×60s of wasted work, and that node's local staged `data.log` is never reclaimed, so the disk keeps growing.

Before this change the client had no deadline at all (the auditor context derives from `maintenanceCtx` and is cancelled only when the writer closes, `woodpecker/log/log_writer.go:367`), so these segments did complete within the server's 300s budget — this is a behavioural regression, and it regresses in exactly the direction the PR is trying to close. It also cannot be worked around from Milvus: `setCustomWpConfig` in `milvus/pkg/streaming/walimpls/impls/wp/builder.go:94` overrides only `Auditor.MaxInterval`, so both new keys keep their built-in defaults.

Suggestion: size the attempt budget at ≥ the server-side per-segment budget (e.g. derive it from `segmentCompactionPolicy.timeout` plus RPC margin) and recompute the pass budget from it — note the worst-case pass then becomes 3×300s, which the current 300s pass default also cannot accommodate. *(raised by bigsheeper, czs007)*

## Medium

**`woodpecker/segment/segment_handle.go:1654` — no `ctx.Err()` check between quorum nodes, so a pass-deadline expiry is reported as "compaction failed on all quorum nodes"**

The node loop at `segment_handle.go:1654` has no cancellation check at its top, and the new attempt context is derived from the caller's context (`segment_handle.go:1689-1692`). When the new pass deadline (`log_writer.go:412`, `internal_log_writer.go:336`) fires while node A's attempt is in flight, the parent context is already done: `ClientPool.GetLogStoreClient` does not check the context and returns a cached client (`woodpecker/client/logstore_client_pool_remote.go:77-82`), so the loop does not exit early; `WithTimeout` on a done context yields already-expired children for nodes B and C, whose `SegmentCompact` calls fail instantly; then `segment_handle.go:1727-1734` increments `WpSegmentCompactionFailuresTotal` and logs `"Compaction failed on all quorum nodes"` with every node listed in `nodeFailures`, including two healthy replicas that were never contacted.

Failure scenario: a quorum node hangs (object store unresponsive rather than erroring) and several segments each burn a 60s attempt, landing the 300s pass deadline mid-walk on one segment → that segment's remaining healthy replicas are recorded in `nodeFailures` and trigger the all-nodes-failed warning, and `WpSegmentCompactionFailuresTotal` counts never-contacted healthy nodes every cycle, corrupting the signal operators use to identify a broken replica.

Suggestion: add `if ctx.Err() != nil { break }` at the top of the loop, mirroring `compactCompletedSegments` (`woodpecker/log/log_auditor.go:83`); this preserves both the metric's meaning and the invariant that the walk actually reached the remaining replicas. *(raised by czs007)*

## Low

**`common/config/configuration.go:650` — validation rejects the non-positive value the code documents as the unbounded escape hatch**

`log_auditor.go:48-52` and `segment_handle.go:1686-1688` both define a non-positive timeout as "unbounded, i.e. the behaviour before this deadline existed", but `configuration.go:650-655` rejects it outright: both new keys error when `Seconds() <= 0`. The documented escape hatch is therefore reachable only by callers that build a `Configuration` programmatically without going through `NewConfiguration`/`Validate` (e.g. test fixtures); a config-file user wanting pre-upgrade behaviour has to supply an arbitrarily large value instead.

Failure scenario: an operator follows the comments and sets `compactionAttemptTimeout: 0` in `woodpecker.yaml` to sidestep the high finding above → `NewConfiguration` fails validation with `"auditor compaction attempt timeout must be positive, got 0s"` and the client does not start, rather than reverting to unbounded behaviour.

Suggestion: reconcile the two — either accept `0` as "unbounded" in validation, or drop the unbounded branch from the comments. *(raised by czs007)*

**`common/config/configuration.go:82` — the 300s pass default does not satisfy the documented `> attempt timeout × quorum size` relation when 5 replicas are configured**

The comment at `configuration.go:89-91` requires `CompactionPassTimeout` to be strictly greater than `CompactionAttemptTimeout × quorum size`, but the config supports 5 replicas (`configuration.go:130-139`, `GetEnsembleSize` accepts 3 and 5; `config/woodpecker.yaml:68` documents `Valid values: [3,5]`), where 300s is exactly 5×60s and the relation does not hold. The practical window is narrow: the fifth node's attempt is itself already capped at 60s by the attempt timeout (`segment_handle.go:1690-1691`), so the pass deadline only additionally truncates that segment's per-segment setup (`GetRecoverableSegmentHandle`, `RefreshAndGetMetadata`, `GetQuorumInfo`, `GetLogStoreClient`), which is sub-millisecond with a healthy etcd; longer runs are truncated by the attempt timeout instead (the high finding above). Nor can a segment starve permanently — `compactCompletedSegments` iterates a map (`log_auditor.go:82`) in random order, so the segment eventually goes first in some cycle and gets all five attempts.

Suggestion: low-cost hardening — either require `pass > attempt × 5` in validation, or restate the comment in terms of the maximum replica count actually supported. *(raised by sijie-ni-0214, bigsheeper)*

**`woodpecker/segment/segment_handle.go:1760` — attempt timeouts are classified as `other` rather than `transient` in remote quorum mode, mislabelling the very failure mode this PR targets**

`classifyCompactionFailureReason` (`segment_handle.go:1753-1764`) detects timeout aborts via `errors.Is(err, context.DeadlineExceeded)`. In remote mode, however, the new attempt deadline is enforced by gRPC: once the client context expires, `CompactSegmentWithExpected` (`logstore_client_remote.go:334`) returns `status.Error(codes.DeadlineExceeded, "context deadline exceeded")`, which does not wrap `context.DeadlineExceeded` (so `errors.Is` is false), is not a `woodpeckerError` (`werr.IsRetryableErr` only matches those, `common/werr/errors.go:320-326`), and is not matched by `IsTransportError` (`common/werr/utils.go:183-188` only accepts `codes.Unavailable`; the string fallback at `:218-223` does not include `"context deadline exceeded"`). The error therefore falls to the `default` branch and is labelled `other`. Only embed/local mode (`logstore_client_local.go:74-76`, which passes the context through and returns a native `ctx.Err()`) reaches `transient`.

Failure scenario: a node hangs in a remote quorum deployment (not embed) and the new 60s attempt deadline fires → `WpSegmentCompactionFailuresTotal` carries `reason=other` (or `mixed` via `summarizeCompactionFailureReasons` when node reasons differ) instead of `transient`, so alerts and dashboards that split "transient vs needs operator intervention" on that label misjudge the hung-node case.

Suggestion: also match `status.Code(err) == codes.DeadlineExceeded` in the classifier. *(surfaced during verification)*

tinswzy:
Thanks — I checked all four against the code and all four hold. Pushed in 299c28e.

## High: attempt budget below the server's

Confirmed, and it is a regression in exactly the direction this PR was meant to close. Verified each
step: server budget 300s (`segment_processor.go:642`), client attempt 60s (`configuration.go:921`),
the deadline reaches the node over gRPC so the effective budget is the min, and a cut-off attempt
keeps nothing — `compactRemoteFooterIfPresent` returns `handled=false` when the footer is missing,
so the writer restarts from the first block. `setCustomWpConfig` does only override
`Auditor.MaxInterval`, so a Milvus deployment could not have worked around it either.

The root error was mine: I picked 60s so that `pass(300) > 3 x attempt(180)` held, without checking
it against the server's own per-segment budget — a parameter chosen to satisfy an invariant rather
than the system.

Attempt is now **330s** (server default plus RPC margin), and cross-section validation rejects any
value below the local `logstore.segmentCompactionPolicy.timeout`, so the two cannot drift apart.

**On the follow-on you flagged** — that the worst-case pass then becomes 3x300s, which a 300s pass
default cannot accommodate: rather than raise the pass bound to 990s (1650s for five replicas),
which would have left it too large to be worth much, I changed the pass from a deadline to a
**budget**. Once it is spent no further segment is started, but a compaction already in flight runs
to completion.

That resolves the tension rather than trading one side against the other:

- a pass lasts at most `budget + one segment`, so the bound stays small (default now **60s**)
- it needs no relation to the attempt timeout or the quorum size, so the `pass > attempt x quorum`
  invariant is gone — **which also resolves the five-replica finding below**
- it cannot cut off a segment that is legitimately slow, which was the same trap as the attempt
  timeout being too low

`TestCompactCompletedSegments_BudgetStopsStartingButNeverCutsOffInFlight` pins it: a 20ms budget
against 120ms segments starts exactly one and asserts it completed rather than being cancelled.

## Medium: no cancellation check between quorum nodes

Confirmed — `GetLogStoreClient` returns a cached client without consulting the context
(`logstore_client_pool_remote.go:78-81`), so the walk does keep going and untried replicas land in
`nodeFailures`.

I went one step further than the suggested `break`: a plain break still falls through to the
all-nodes-failed metric and warning, which would report every replica as failed when none were
tried. It now returns `ctx.Err()`. Any node that genuinely failed before the cancellation has
already been logged inside the loop.

Note this path is now much rarer anyway, since the pass no longer cancels the context mid-walk —
but writer shutdown still reaches it, and the guard is cheap.

## Low: validation rejects the documented escape hatch

Confirmed, and both halves were mine. The non-positive branch only ever applied to a `Configuration`
assembled programmatically without going through `Validate` (test fixtures), never to a config-file
user. The comments now say that; validation stays strict. `withOptionalTimeout` is gone with the
deadline it served.

## Low: 300s does not satisfy the documented relation at five replicas

Confirmed — `GetEnsembleSize` accepts 5 and the YAML documents `[3,5]`, so 300 = 5x60 exactly. Moot
now: the soft budget removes the relation entirely, so there is no ratio left to violate.

## Low: gRPC deadlines classified as `other`

Confirmed. `classifyCompactionFailureReason` now matches `status.Code(err) == codes.DeadlineExceeded`
as well. `TestClassifyCompactionFailureReason_GrpcDeadlineIsTransient` asserts the premise first —
that a gRPC status error does not wrap the context sentinel — so if that ever changes, the test says
the extra check has become redundant rather than silently keeping it.

## Verification

`go test` and `go test -race` on `common/config`, `woodpecker/log`, `woodpecker/segment`: ok.
`golangci-lint` (CI-pinned v2.11.4): 0 issues. `gofmt`/`go mod tidy`: clean.

Since the shipped values equal the in-code defaults, a wrong key or indentation would be invisible;
checked with a throwaway config setting them to 777/888 and both came through.

czs007:

Adversarial review of this round surfaced four items — two medium (an upgrade-breaking cross-section validation and an unordered budget cutoff) and two low (an unwired external config path and a PR body that no longer matches the diff).

## Medium

**`common/config/configuration.go:623` — new cross-section hard validation rejects startup for deployments that only raised the server-side compaction timeout**
The new check at `validateWoodpeckerConfig()` errors when `client.auditor.compactionAttemptTimeout < logstore.segmentCompactionPolicy.timeout`, and it sits on the `Validate()` main path called unconditionally by `NewConfiguration()`. Because YAML is unmarshalled onto an already-defaulted struct, an existing `woodpecker.yaml` that never mentions `compactionAttemptTimeout` silently picks up the new 330s default, while an operator-tuned `segmentCompactionPolicy.timeout` (say 600) is still read — and the two are compared. Failure: an existing deployment with `segmentCompactionPolicy.timeout: 600` and no `compactionAttemptTimeout` key restarts after this PR → `NewConfiguration()` returns "auditor compaction attempt timeout (330s) must be at least the segment compaction timeout (600s)", `cmd/main.go:105-108` hits `log.Fatalf`, and nodes fail to come up one by one during the rolling upgrade until config is edited by hand. Two further points widen the blast radius: in service mode the check only sees the *local* logstore settings, so a logstore process raising its own purely server-side knob is killed by this process's client-auditor default even though its client auditor never initiates compaction; and conversely the real compacting client validates against its local default (300), which says nothing about the remote server's actual setting — so the gate both false-positives and fails to give the guarantee it implies. Suggestion: the strict behaviour appears deliberate, so at minimum call this upgrade break out by name in the release note ("raising `segmentCompactionPolicy.timeout` now requires raising `compactionAttemptTimeout` in step"); alternatively downgrade to a startup warning when the attempt timeout is at its default, i.e. unset by the operator. In-repo configs are unaffected (`config/woodpecker.yaml:37` attempt=330 ≥ `:108` timeout=300; `cmd/external/user_example.yaml:244-247` sets no timeout). *(raised by bigsheeper)*

**`woodpecker/log/log_auditor.go:92` — the pass budget truncates a randomized Go map walk, so the deferred segment set is random each round**
`compactCompletedSegments` iterates `map[int64]*meta.SegmentMeta`, whose iteration order Go randomizes. Before this PR that was harmless: every Completed segment in the snapshot was processed each round, so order did not affect the outcome. With the budget `break`, which segments a round covers is decided entirely by that round's random map order — there is no sort by `segNo`, no cursor, and no memory of prior progress. Failure: a log accumulates a large Completed backlog (e.g. object storage slows down, ~10s per compaction, 200 pending) so the 60s `compactionPassBudget` only starts ~6 per round → which 6 is random, a segment (possibly the oldest) can be skipped for many consecutive rounds while its node's local `data.log` stays on disk, and `segmentsDeferred` reports only a count so operators cannot tell a repeatedly-skipped segment from a persistently-failing one. This is a reclamation-latency and predictability problem, not data corruption, but it undercuts precisely the goal this PR states. Suggestion: the pass already walks the map once to compute `total` — collect the Completed `segNo`s there, sort them, and iterate in order; consider logging a few deferred segment ids alongside the count. *(surfaced during verification)*

## Low

**`cmd/external/user_config.go:289` — the two new auditor keys are not wired into the external user-config merge and are silently ignored**
This PR adds `client.auditor.compactionAttemptTimeout` / `compactionPassBudget` and documents them in `config/woodpecker.yaml:37-38`, but the `cmd/external` path (used by `cmd/main.go:110-117`) merges by explicit field assignment — as its README states — and `mergeWoodpeckerConfig` copies only `MaxInterval` under `Auditor`. Failure: an external user follows the new comments and sets `woodpecker.client.auditor.compactionPassBudget: 300` (or `compactionAttemptTimeout: 600`) in a Milvus-style `user.yaml` → the value is dropped with no log line and no validation error, the runtime keeps the 60s/330s defaults, and the operator believes the tuning took effect. In fairness, the same function also omits `SegmentCompactionPolicy.Timeout` (`user_config.go:339-348`), so this is a standing habit of the merge function rather than a pattern this PR invented — but the two lines for the newly introduced keys belong in this PR. Suggestion: add the two `if src.Client.Auditor.<Field>.Seconds() > 0` blocks next to the `MaxInterval` one. *(surfaced during verification)*

**`woodpecker/log/log_auditor.go:80` — the PR description documents an earlier design that does not match the merged diff**
The body describes a second knob named `client.auditor.compactionPassTimeout` (300s) that "wraps the context at the call site", plus `withOptionalTimeout`, `TestWithOptionalTimeout` and `TestCompactCompletedSegments_StopsAtPassDeadline`. The diff has `compactionPassBudget` (60s), which neither wraps the context nor cancels in-flight compactions; `withOptionalTimeout` does not exist anywhere in the tree, and the tests are `TestCompactCompletedSegments_BudgetStopsStartingButNeverCutsOffInFlight` / `_ZeroBudgetIsUnbounded`. The body's "attempt timeout 60s" and the "pass timeout > attemptTimeout × quorum size (300 vs 3×60)" invariant are also the exact inverse of the shipped defaults (attempt 330 / budget 60, `common/config/configuration.go:945-947`). Per this repo's house rules the PR body is read as the design description and is often lifted into release notes. Suggestion: rewrite the "Two bounds", "The invariant", and "Verification" sections against the final implementation. *(surfaced during verification)*

tinswzy:
Rebased onto master now that #322 and #323 have merged, and force-pushed (`c44034e`). The two review commits are squashed into one, since resolving the same conflicts twice on an intermediate commit whose code the next one deletes is not worth the risk of getting it wrong — the earlier commit messages are folded into the single one.

The conflict was with #323, which bounds the same pass by count. They compose rather than overlap:

```go
if len(completed) > maxCompactedPerCycle {
    st.deferred = len(completed) - maxCompactedPerCycle   // dropped by the count bound (#323)
    completed = completed[:maxCompactedPerCycle]
}
...
if !deadline.IsZero() && !time.Now().Before(deadline) {
    st.deferred += len(completed) - st.processed          // not started before the budget ran out
    break
}
```

`deferred` is now the sum of both causes rather than either one, and the field comment says so. `TestCompactCompletedSegments_CountBoundAndBudgetCompose` covers the case where both fire in one pass, which neither PR's own tests reached. #323's oldest-first ordering and per-cycle cap are kept as merged; this PR only inserts the time check into that loop.

The PR description was stale in a way that would have misled review — it still described the first shape of this change (60s attempt, a hard 300s pass deadline, `withOptionalTimeout`, and a claim that the branch does not touch `compactCompletedSegments`). Rewritten to the state that is actually on the branch.

Re-verified after the rebase: `go test ./common/config/ ./woodpecker/log/ ./woodpecker/segment/` ok, `golangci-lint run ./woodpecker/... ./common/...` 0 issues, `gofmt -l .` clean, `go mod tidy` no diff.

tinswzy:
Rebased again (`d7c758d`) now that #325 and #326 have merged. One conflicted file this time, `woodpecker/segment/segment_handle_test.go` — both branches appended tests at the end, and the import block needed `require` (from #326) alongside `codes`/`status` (from here). Both sets of tests are kept; nothing else in that file changed. `configuration.go`, `woodpecker.yaml` and `segment_handle.go` auto-merged, and I checked the result carries both #325's `syncScheduler.maxWorkers` and this PR's two auditor keys.

Also clarified two things in the description that a reader could reasonably get wrong:

**"Attempt" is one RPC to one node, not a retry.** The `for _, node := range quorumInfo.Nodes` walk is the pre-existing failover — a replica whose data is behind returns `ErrSegmentCompactionDataBehind` and the walk moves on. Each replica is tried at most once per pass, nothing here wraps compaction in `retry.Do`, and the retry is still the next auditor cycle.

**The concrete worst case is now in the description.** "A pass lasts at most the budget plus one segment" understated it:

```
one segment, worst case = Wq × compactionAttemptTimeout = 3 × 330s = 990s
one pass,    worst case = 60s + 990s ≈ 17.5 min
```

Still bounded where it was not before, but 17.5 minutes is the number to judge. Tightening it means either lowering the attempt timeout (unavailable — below the server's 300s is the regression this PR's second commit fixed) or capping replicas tried per segment per pass. I'd rather not fold either into this one.

Re-verified after the rebase: `go test ./common/config/ ./woodpecker/log/ ./woodpecker/segment/` ok, `golangci-lint run ./woodpecker/... ./common/...` 0 issues, `gofmt -l .` clean, `go mod tidy` no diff.

czs007:

One medium-severity issue survived verification: the new pass budget combined with the deterministic oldest-first ordering can let a single slow-failing segment starve every newer segment in the same log.

## Medium

**`woodpecker/log/log_auditor.go:143`** — Head-of-line starvation from fixed oldest-first ordering plus the new pass budget.

The budget check added in this PR (`log_auditor.go:129-137` builds the deadline, `:143-146` breaks out of the loop) is evaluated only before starting the next segment, so the previous segment's elapsed time counts against the budget whether it succeeded or failed. Iteration order is deterministic ascending (`log_auditor.go:123`, `sort.Slice`), and a failed segment is neither dequeued nor backed off: when `compactSegmentQuorum` fails on all nodes, `Compact` returns the error without advancing state (`segment_handle.go:2090-2100`), so the segment stays `Completed` (state gate at `segment_handle.go:2051`) and sorts first again on the next cycle.

Failure scenario: in service mode, the lowest-numbered `Completed` segment of a log fails on all 3 replicas — e.g. it is near `maxSize` and each replica hits `logstore.segmentCompactionPolicy.timeout=300s` (`server/processor/segment_processor.go:642-643`, `config/woodpecker.yaml:115`), giving roughly 900s across the sequential quorum walk, or the nodes are unresponsive and each replica consumes the new `compactionAttemptTimeout=330s` for ≈990s → with the default `compactionPassBudget=60s`, every cycle starts only that one segment (the first iteration always passes the deadline check), then breaks and marks all newer segments `deferred`; the same segment leads the next cycle, so every newer `Completed` segment in that log is never compacted and its staged `data.log` on the logstore nodes is never reclaimed — the exact disk-growth direction this PR sets out to fix.

Trigger conditions are genuinely narrow: it requires the same segment to fail on *all* replicas, consume at least the whole budget in a single cycle, and repeat every cycle. On a healthy cluster compaction is fast and many segments start per cycle.

Suggestion: any of (a) a resume cursor that remembers where the previous cycle broke and continues from there, (b) per-segment backoff for consecutively failing segments, or (c) charging only successful compaction time against the budget.

*Prior-round context:* the earlier fairness item on this file (random map iteration order, round 1) is genuinely resolved — #323 made the order ascending — and this finding is the new interaction that ordering creates with the budget added here. The author's 09-11 rebase comment states only that #323's oldest-first ordering is preserved and that this PR merely inserts a time check; it does not address the compounding of fixed ordering with the budget, so that specific point remains unanswered.

(raised by bigsheeper, xiaocai2333)

tinswzy:
Reworked after review, in `ad64864`. The two compaction bounds were being treated as one; they are not.

**What was wrong.** The PR added a cross-field check requiring
`client.auditor.compactionAttemptTimeout >= logstore.segmentCompactionPolicy.timeout`. That reads
the relationship backwards. A logstore node serves many `bucketName/rootPath` tenants from a single
configuration (`segmentProcessors` is keyed by log key, `getOrCreateSegmentProcessor` takes the
bucket and root path per request), so its timeout is a property of *the node* — a ceiling that stops
any one request holding its compaction resources indefinitely. It cannot express what a particular
tenant wants, and the client has no way to know it: in service mode the two sides are separate
processes, so validating the client's own copy of a server setting proved nothing about the node it
talks to, and under multiple tenants there is no single value it could have proved anything about.

**The mechanism was already correct** — the client sends its deadline with the request,
`segmentProcessor.Compact` wraps the incoming context with the node's ceiling, and the shorter wins.
Only the framing was wrong. So this removes the check and names which side owns each bound:

| Setting | Owner | Means |
|---|---|---|
| `client.auditor.compactionAttemptTimeout` | tenant | how long I wait for one node to answer one request |
| `logstore.segmentCompactionPolicy.timeout` | node | nobody's compaction runs longer than this here |
| `client.auditor.compactionPassBudget` | tenant | after this, start no further segment this cycle |

Setting the client value below the node's ceiling is now legal and means what it says: give up
before the node would have. The comment keeps the warning that nothing of an abandoned attempt is
kept, so a value below what compaction actually needs makes every attempt futile — but that is a
tenant's decision to get wrong, not an invariant to enforce.

**Why the client bound is not redundant**, since that was the question that started this. The three
cover different edges and none substitutes for another:

- the node's ceiling cannot bound a node that stops answering at all — it is enforced by the node,
  so a wedged, black-holed or GC-stalled one never reaches it, and the auditor's context carries no
  deadline of its own, so the call would block forever;
- the pass budget cannot either — it is checked *between* segments, and a call that never returns
  never gets back to the check;
- the client deadline cannot bound a cycle — it only covers one RPC, while a cycle compacts up to
  64 segments sequentially.

**Comments rewritten on all three**, including the two that were misleading on their own terms:
`compactionPassBudget` now says it is a *start* deadline rather than a cap on the cycle, with the
real worst case (the budget plus one segment — minutes, not the configured 60s) and the steady-state
throughput it implies when compactions are slow; `segmentCompactionPolicy.timeout` is named as the
node's ceiling rather than a per-compaction policy.

**Tests:** `TestSegmentProcessor_Compact_HonoursTheCallerDeadline` pins the combination that was only
implied before — a shorter caller deadline reaches the compaction and beats the node ceiling, and a
caller with no deadline (which is what the auditor has) is bounded by the ceiling. The removed
cross-field check takes its test with it.

`go test ./common/config/ ./server/processor/ ./woodpecker/log/ ./woodpecker/segment/` ok,
`golangci-lint` 0 issues, `gofmt` clean, `go mod tidy` no diff.

**Not changed here, worth a separate look:** with the shipped defaults the node's 300s ceiling is
always shorter than the client's 330s, so it is the node — not the tenant — that decides in every
normal case. If the ceiling is meant as a guardrail rather than the policy, it is probably set too
low, and a tenant asking for 600s cannot get it. That is a behaviour change and does not belong in
this PR.

czs007:

This PR bounds auditor compaction time per attempt and per pass. Four findings remain: the new pass budget can hold back newer segments behind one that keeps failing, one timeout path is still counted as `other`, one subtest fails when run alone, and the external config merge skips the new settings.

## Medium

**`woodpecker/log/log_auditor.go:138`: A Completed segment that keeps failing can block every other Completed segment**
- **Problem:** `compactCompletedSegments` sorts Completed segments oldest-first (`:122`). The new budget check (`:136-146`) stops starting segments once the budget is spent, but the first segment always starts. A failed segment only hits `continue` (`:151-155`), stays Completed, and there is no skip or backoff. If the oldest segment fails every time and each attempt takes at least the budget, each cycle compacts only that segment and defers the rest. The next cycle sorts the same way and picks it again. Before this PR, the loop moved on to later segments after a failure.
- **Failure:** in remote mode with the default 60s `compactionPassBudget`, the oldest Completed segment needs more than the node's 300s ceiling, so every replica aborts at 300s → every cycle defers all newer Completed segments, they never reach Sealed, their local `data.log` is never reclaimed, and node disk usage keeps growing until someone fixes that segment by hand.
- **Suggestion:** do one of the following:
  - move segments that failed in this pass or recently to the end of the order;
  - track failure counts and apply backoff;
  - keep starting later segments within the budget after a failure, instead of letting one slow failure use the whole budget.
- **History:** an earlier round raised the related concern that budget truncation gives no fairness or priority guarantee. The ordering is now deterministic, but the author has not responded on the case of a segment that keeps failing, and `:122-146` still behaves as described.

(raised by bigsheeper)

## Low

**`woodpecker/segment/segment_handle.go:1835`: The node's 300s compaction timeout is still classified as `other` instead of transient**
- **Problem:** with the defaults (client attempt 330s, node ceiling 300s), the node-side timeout fires first. The error goes through these steps:
  1. The writer returns `ctx.Err()`, wrapped with `%w` (`writer_impl.go:1637/1655/1764` → `:1437`).
  2. `service.go:634` converts it with `werr.Status`, which sets code 8002 and `Retriable=false`, because `IsRetryableErr` is false for an error that is not a `woodpeckerError` (`errors.go:320-325`).
  3. The client rebuilds it at `logstore_client_remote.go:338-340` as `woodpeckerError{code: 8002, retryable: false}`.

  That value is not a gRPC status and does not wrap `context.DeadlineExceeded`, so every check at `:1847`, including the new `isDeadlineExceeded`, misses it.
- **Failure:** in remote mode with default config, a compaction takes more than 300s on a node (slow object store or large segment) → the `nodeFailures` log shows `node:other`, and `WpSegmentCompactionFailuresTotal` counts it under `other`/`mixed` instead of `transient`. That is the wrong label for the hung-node case this PR targets.
- **Suggestion:** add `errors.Is(err, werr.ErrTimeoutError)` to `isDeadlineExceeded`. `woodpeckerError.Is` compares codes (`errors.go:272-273`), so it matches the rebuilt error. Prefer this to `werr.IsTimeoutError`, which also treats `context.Canceled` as a timeout. Also add a classification test for `werr.Error(werr.Status(fmt.Errorf("x: %w", context.DeadlineExceeded)))`. `TestClassifyCompactionFailureReason_GrpcDeadlineIsTransient` currently covers only the gRPC status form and the bare sentinel.
- **History:** an earlier round reported that attempt timeouts in remote quorum mode were labeled `other`. The author responded by adding `isDeadlineExceeded` and the gRPC deadline test. That response is not sufficient: under default config the node-side ceiling fires first, and that path still produces the rebuilt 8002 error the new check does not match.

(raised by sijie-ni-0214, bigsheeper) (provenance unclear — please confirm whether this PR introduces it)

**`server/processor/segment_processor_test.go:1183`: Subtest 2 depends on a variable that only subtest 1 sets**
- **Problem:** `nodeCeiling` is declared as `0` at `:1140` and assigned only inside subtest 1 (`:1144`). Subtest 2 ("without a caller deadline the node ceiling applies") reads it at `:1183`.
- **Failure:** `go test -run 'TestSegmentProcessor_Compact_HonoursTheCallerDeadline/without_a_caller_deadline_the_node_ceiling_applies' ./server/processor/` → `nodeCeiling` stays 0 and `InDelta(0, ~300, 5)` fails even though the code under test is correct.
- **Suggestion:** read the ceiling from subtest 2's own `sp.cfg`, or compute it once before either `t.Run`.

(raised by bigsheeper)

**`cmd/external/user_config.go:289`: The external user config merge ignores `compactionAttemptTimeout` and `compactionPassBudget`**
- **Problem:** `mergeWoodpeckerConfig` copies fields one at a time, and for the auditor it copies only `MaxInterval` (`:289-291`). The two `AuditorConfig` fields added by this PR are never copied, so values set in the external user config are silently dropped.
- **Failure:** a deployment using `LoadUserConfig` + `ApplyToConfig` sets `woodpecker.client.auditor.compactionAttemptTimeout: 600` to match a larger node ceiling → the setting is ignored with no error or log, the client still cuts the attempt at 330s, and that segment's compaction fails every cycle.
- **Suggestion:** add `if src.Client.Auditor.CompactionAttemptTimeout.Seconds() > 0 { ... }` and the same check for `CompactionPassBudget`.
- **Pre-existing, follow-up only:** `logstore.segmentCompactionPolicy.timeout` is also not mapped (`:339-348`). That gap predates this PR and should be tracked separately (no tracking issue yet).
- **History:** this was raised in an earlier round. The author has not responded, and `:289-291` is unchanged.

(raised by bigsheeper)

PR #325, opened 2026-09-10
enhance(logstore): size the shared flush pool from config, not the host's cores

### LINKED ISSUE #314: enhance(logstore): sync worker count is derived from host CPU count, so it cannot saturate device IOPS
### Summary

The staged-storage sync worker pool is sized as `runtime.NumCPU() * 2`. Those workers block in `fdatasync`, so CPU count is the wrong variable to size them by — and `runtime.NumCPU()` reads the **host** core count, not the pod's cgroup limit. The practical effect is that the pool caps concurrent fsync well below what the underlying volume can serve: we measured a per-volume write ceiling of ~2,126 IOPS against a gp3 baseline of 3,000, with the surplus load turning into queueing (p99 write latency) instead of throughput.

This matters most for the many-logs case: each log writes its own file, so the number of concurrent fsyncs a node can sustain is what bounds how many active logs a single pod can carry at a given per-message write latency.

### Where it is

```go
// server/logstore.go:145   (line numbers against master)
syncScheduler: stagedstorage.NewSyncScheduler(runtime.NumCPU() * 2),

// server/storage/stagedstorage/sync_scheduler.go:82,89
defaultSyncScheduler = NewSyncScheduler(runtime.NumCPU() * 2)
...
if workerCount <= 0 { workerCount = runtime.NumCPU() * 2 }
```

Each worker runs a job that ends in a blocking `w.file.Sync()` (`server/storage/stagedstorage/writer_impl.go:779`). Go hands the P back to the scheduler for the duration of a blocking syscall, so these goroutines consume essentially no CPU — they cost an OS thread, not a core. Sizing a pure blocking-IO pool by core count has no mechanical basis.

### Evidence

**1. The pool, not the device, is the ceiling.** A load test at three stages, same log set (2,199 logs / 851 active) on 5 pods whose scheduler capacity was 8:

| stage | throughput/pod | per-volume write IOPS | write p99 |
|---|---|---|---|
| 1 | 2.08 MB/s | 1,363 | 65.9 ms |
| 2 | 5.24 MB/s | 2,075 | 152.3 ms |
| 3 | 11.49 MB/s | **2,126** | **351.3 ms** |

From stage 2 to 3 throughput rose 2.2x, IOPS rose 2%, and p99 rose 2.3x. The plateau matches the pool exactly: 2,126 / 8 workers = 266 fsync/s per worker, i.e. ~3.76 ms per fsync — a plausible device round trip. Everything above that queues.

**2. Concurrency is the lever that buys IOPS.** A separate scan on a higher-IOPS volume class, one op = `WriteAt` + `Fdatasync`:

| threads | 8 | 16 | 32 | 64 | 128 |
|---|---|---|---|---|---|
| write IOPS | 8,753 | 15,112 | 22,980 | 32,622 | 42,786 |

Still not saturated at 128. Achievable IOPS tracks `concurrency / fsync latency`, not device capability, until the device caps.

**3. Host core count makes the value non-deterministic.** Across a fleet of identically-specced pods we observed `sync_scheduler_capacity` of 8 / 32 / 64, depending purely on whether the pod landed on a 4-, 16- or 32-core node. Raising the pod's CPU limit does not change it: on the same 4-core node, pools with `limit: 2` and `limit: 4` both reported capacity 8. So the same deployment gets an 8x spread in flush concurrency — and therefore in tail latency — as a function of scheduling.

The repo already has a cgroup-aware helper, `common/hardware/hardware_info.go:GetCPUNum()` (reads `GOMAXPROCS`), but `logstore.go` uses the raw `runtime.NumCPU()`.

### A related key exists, but its scope is different — do not reuse it

`logstore.segmentSyncPolicy.maxFlushThreads` (`common/config/configuration.go:189`, shipped as `32` in `config/woodpecker.yaml:101`) looks like the right knob, but it is **per segment writer**, not node-shared:

```go
// server/storage/objectstorage/writer_impl.go:143
// inside NewMinioFileWriterWithMode(ctx, bucket, baseDir, logId, segId, ...)
pool:             conc.NewPool[*blockUploadResult](cfg...MaxFlushThreads, conc.WithPreAlloc(true)),
flushingTaskList: make(chan *blockUploadTask, cfg...MaxFlushThreads),
```

Every `(logId, segId)` writer builds its own pool and its own buffered channel of that size, so the process-wide bound scales with the number of active segment writers. The `SyncScheduler`, by contrast, is constructed once in `NewLogStore` and handed to every segment processor (`server/logstore.go:374`); its metrics are labelled by `NodeID` (`server/logstore.go:1176`), which is the correct scope for it.

The two are also reached through mutually exclusive branches in `server/processor/segment_processor.go:564-610` — `IsStorageService()` uses the staged writer with the shared scheduler, the `else` branch uses the Minio writer with the per-writer pool. So a single key would mean "32 goroutines per segment" under one `storage.type` and "32 for the whole node" under another, from the same YAML value. That is a trap for anyone tuning it, in both directions: raising it for local fsync concurrency would multiply the object-storage pools by the active segment count, and lowering it to bound object-storage memory would starve the node's fsync pool.

Note separately that the in-code default for `maxFlushThreads` is `8` (`common/config/configuration.go:935`) while the shipped YAML says `32`; those should agree regardless of this issue.

### Proposal

1. Add a **new, node-scoped** config key for the shared scheduler — e.g. `logstore.syncScheduler.maxWorkers`. It should not live under `segmentSyncPolicy`, which is a per-segment policy section; putting a node-shared value there is what makes the scope ambiguous in the first place.
2. **Make it a plain absolute number, not a multiplier of anything.** A multiplier of core count reintroduces the scheduling-dependent drift above, and the quantity has no relationship to cores. A plain number gives the same flush concurrency on every node shape, so a deployment can be tuned to its volume's IOPS budget once and behave the same everywhere.
3. Default `32`. At the measured ~3.76 ms fsync this offers ~8,500 IOPS of headroom — enough to saturate a gp3 baseline volume (3,000) several times over, and to make use of provisioned-IOPS or higher-class volumes on other clouds without a rebuild.
4. Wire it in `NewLogStore`, which already has `cfg` in scope at the call site.
5. Drop the `runtime.NumCPU()` fallback in `NewSyncScheduler`; if a caller passes `<= 0`, use the configured default.
6. Decouple the queue depth from worker count. `queueSize := workerCount * 4096` allocates two channels of that size up front; at 256 workers that is ~33 MB per scheduler, and it scales with the wrong quantity.

### Tuning guidance to document

Achievable fsync throughput follows Little's Law: `workers ≈ target volume IOPS × fsync latency`. Past the point where the device saturates, additional workers buy queueing rather than throughput — and shrink the per-block batching that currently absorbs load — so the value should be set from the volume's IOPS budget, not raised indefinitely.

### Expected outcome

On a node whose pods were getting 8 workers, the flush pool becomes 32 and the volume can be driven to its actual IOPS ceiling instead of a pool-imposed one. That converts today's queueing back into throughput, which is what lets a single pod host more concurrent active logs while keeping per-message write latency low — and it makes flush concurrency a property of the configuration rather than of which node the pod happened to land on.

Happy to send a PR for this.

tinswzy:
## Problem

The staged-storage sync worker pool was sized as `runtime.NumCPU() * 2`.

Fixes #314

Its workers block in `fdatasync` (`writer_impl.go`, `w.file.Sync()`), so core count is the wrong
quantity to size them by: Go hands the P back for the duration of a blocking syscall, so a worker
parked in `fdatasync` costs an OS thread, not a core. And `runtime.NumCPU()` reports the **host's**
cores rather than the pod's cgroup limit, so identically specced pods ended up with pools differing
by a factor of eight depending on where they were scheduled — and raising a pod's CPU limit did not
change it.

The practical effect: the pool, not the volume, capped concurrent fsync. A per-volume write ceiling
of ~2,126 IOPS against a gp3 baseline of 3,000, with load above that turning into queueing rather
than throughput. Because each log writes its own file, the sustainable number of concurrent fsyncs
also bounds how many active logs a pod can carry at a given per-message write latency.

## `logstore.syncScheduler.maxWorkers`

A plain count, default 32, wired in `NewLogStore` where `cfg` was already in scope.

**A plain number, not a multiplier.** A multiplier of core count reintroduces exactly the drift
above, and the quantity has no relationship to cores. A plain number gives the same flush
concurrency on every node shape, so a deployment is tuned to its volume's IOPS budget once.

**Not under `segmentSyncPolicy`.** That section is per-segment policy, while this pool is shared by
every writer on the node. `segmentSyncPolicy.maxFlushThreads` also already means something
different: it sizes a pool per object-storage *segment writer*, so the process-wide bound there
scales with the number of active segment writers. Reusing that key would have made one YAML value
mean "per segment" under one `storage.type` and "per node" under another.

The YAML comment carries the tuning rule: `workers ≈ target volume IOPS × fsync latency`, and that
past the point where the device saturates more workers buy queueing rather than throughput — so it
should be raised toward the volume's budget, not indefinitely.

## Queue depth decoupled

`queueSize := workerCount * 4096` tied a multi-megabyte preallocation to a quantity that has
nothing to do with it. The queues hold *pending* work, which is bounded by the number of writers
rather than by how many are being serviced: the `syncScheduled` and `syncTaskSubmitted` CAS gates
allow at most one schedule request and one submitted job per writer, so one slot per writer is the
real ceiling.

It is now a constant 65,536 slots:

| host cores | old workers | old slots | old prealloc | new workers | new slots | new prealloc |
|---|---|---|---|---|---|---|
| 4 | 8 | 32,768 | 1.0 MB | 32 | 65,536 | 2.0 MB |
| 8 | 16 | 65,536 | 2.0 MB | 32 | 65,536 | 2.0 MB |
| 16 | 32 | 131,072 | 4.0 MB | 32 | 65,536 | 2.0 MB |
| 32 | 64 | 262,144 | 8.0 MB | 32 | 65,536 | 2.0 MB |

The depth matters rather than being a free parameter: a full `scheduleCh` blocks its caller, and
the caller is the append path — `WriteDataAsync` schedules the next sync check before it returns.
65,536 is at or above the previous value for every host shape except the 16- and 32-core ones,
where the old sizing was simply larger than anything the CAS gates can fill.

## Also

A non-positive worker count falls back to the documented default rather than to the core count, so
`NewSyncScheduler` and the config default cannot disagree. The two benchmarks that passed
`runtime.NumCPU()` now pass the same constant, which also makes their results comparable across
machines.

## Verification

- `go test ./common/config/ ./server/ ./server/storage/stagedstorage/`: ok
- `go test -race ./common/config/ ./server/storage/stagedstorage/`: ok
- `golangci-lint run ./server/... ./common/config/...` with the CI-pinned v2.11.4: 0 issues

New tests:

- `TestNewLogStore_SyncSchedulerUsesConfiguredWorkers` — the end-to-end claim: the pool a logstore
  builds has the capacity the config asked for, across several values.
- `TestNewSyncScheduler_WorkerCountIsExplicit` — explicit counts are honoured; non-positive falls
  back to the default rather than the core count.
- `TestNewSyncScheduler_QueueDepthIsIndependentOfWorkers` — a 1-worker and a 64-worker scheduler
  get the same channel depth.
- Config: `maxWorkers` asserted on the YAML and defaults paths, plus the validation error. Since
  the shipped value equals the in-code default, a wrong indentation would be invisible — checked
  with a throwaway config setting it to 4242, which came through.

Two hand-built `Configuration` fixtures in the config tests needed the new field, the same as for
every other validated logstore setting.

### DIFF
--- a/common/config/configuration.go
+++ b/common/config/configuration.go
@@ -196,6 +196,24 @@ type SegmentCompactionPolicy struct {
 	Timeout            DurationSeconds `yaml:"timeout"`
 }
 
+// SyncSchedulerConfig stores the node-wide staged-storage sync scheduler configuration.
+type SyncSchedulerConfig struct {
+	// MaxWorkers is the number of workers in the node's shared flush pool, as a plain count.
+	//
+	// Each worker blocks in fdatasync while a flush is in flight, so this bounds how many
+	// concurrent syncs the node can have outstanding, not how much CPU it uses -- Go hands the
+	// P back for the duration of a blocking syscall, so a worker parked in fdatasync costs an
+	// OS thread rather than a core. Sizing it by CPU count therefore has no mechanical basis,
+	// and reading the host's core count (rather than the cgroup limit) made the effective value
+	// depend on which machine a pod happened to land on.
+	//
+	// Tune it from the volume's IOPS budget: workers ~= target IOPS x fsync latency. Past the
+	// point where the device saturates, more workers buy queueing rather than throughput, and
+	// shrink the per-block batching that absorbs load, so raise it toward the budget and not
+	// beyond.
+	MaxWorkers int `yaml:"maxWorkers"`
+}
+
 // RetentionPolicyConfig stores the data retention policy configuration.
 type RetentionPolicyConfig struct {
 	TTL int `yaml:"ttl"` // Time to live for truncated segments before eligible for GC
@@ -344,6 +362,7 @@ type MinioConfig struct {
 type LogstoreConfig struct {
 	SegmentSyncPolicy       SegmentSyncPolicyConfig      `yaml:"segmentSyncPolicy"`
 	SegmentCompactionPolicy SegmentCompactionPolicy      `yaml:"segmentCompactionPolicy"`
+	SyncScheduler           SyncSchedulerConfig          `yaml:"syncScheduler"`
 	SegmentReadPolicy       SegmentReadPolicyConfig      `yaml:"segmentReadPolicy"`
 	RetentionPolicy         RetentionPolicyConfig        `yaml:"retentionPolicy"`
 	FencePolicy             FencePolicyConfig            `yaml:"fencePolicy"`
@@ -800,6 +819,9 @@ func (c *Configuration) validateLogstoreConfig() error {
 	}
 
 	// Validate SegmentCompactionPolicy
+	if logstore.SyncScheduler.MaxWorkers <= 0 {
+		return fmt.Errorf("sync scheduler max workers must be positive, got %d", logstore.SyncScheduler.MaxWorkers)
+	}
 	if logstore.SegmentCompactionPolicy.MaxBytes <= 0 {
 		return fmt.Errorf("segment compaction policy max bytes must be positive, got %d", logstore.SegmentCompactionPolicy.MaxBytes.Int64())
 	}
@@ -933,6 +955,9 @@ func getDefaultWoodpeckerConfig() WoodpeckerConfig {
 				RetryInterval:              DurationMilliseconds{Duration: Duration{duration: 2000 * 1000000}}, // 2000ms
 				MaxFlushSize:               ByteSize(16000000),
 				MaxFlushThreads:            8,
+			},
+			SyncScheduler: SyncSchedulerConfig{
+				MaxWorkers: 32,
 			},
 			SegmentCompactionPolicy: SegmentCompactionPolicy{
 				MaxBytes:           ByteSize(32000000),
--- a/common/config/configuration_test.go
+++ b/common/config/configuration_test.go
@@ -72,6 +72,7 @@ func TestNewConfiguration(t *testing.T) {
 	assert.Equal(t, 1000, config.Woodpecker.Logstore.SegmentSyncPolicy.RetryInterval.Milliseconds())
 	assert.Equal(t, int64(2000000), config.Woodpecker.Logstore.SegmentSyncPolicy.MaxFlushSize.Int64())
 	assert.Equal(t, 32, config.Woodpecker.Logstore.SegmentSyncPolicy.MaxFlushThreads)
+	assert.Equal(t, 32, config.Woodpecker.Logstore.SyncScheduler.MaxWorkers)
 	assert.Equal(t, int64(2000000), config.Woodpecker.Logstore.SegmentCompactionPolicy.MaxBytes.Int64())
 	assert.Equal(t, 4, config.Woodpecker.Logstore.SegmentCompactionPolicy.MaxParallelUploads)
 	assert.Equal(t, 8, config.Woodpecker.Logstore.SegmentCompactionPolicy.MaxParallelReads)
@@ -178,6 +179,7 @@ func TestNewConfiguration(t *testing.T) {
 	assert.Equal(t, 2000, defaultConfig.Woodpecker.Logstore.SegmentSyncPolicy.RetryInterval.Milliseconds())
 	assert.Equal(t, int64(16000000), defaultConfig.Woodpecker.Logstore.SegmentSyncPolicy.MaxFlushSize.Int64())
 	assert.Equal(t, 8, defaultConfig.Woodpecker.Logstore.SegmentSyncPolicy.MaxFlushThreads)
+	assert.Equal(t, 32, defaultConfig.Woodpecker.Logstore.SyncScheduler.MaxWorkers)
 	assert.Equal(t, int64(32000000), defaultConfig.Woodpecker.Logstore.SegmentCompactionPolicy.MaxBytes.Int64())
 	assert.Equal(t, 4, defaultConfig.Woodpecker.Logstore.SegmentCompactionPolicy.MaxParallelUploads)
 	assert.Equal(t, 8, defaultConfig.Woodpecker.Logstore.SegmentCompactionPolicy.MaxParallelReads)
@@ -475,6 +477,9 @@ func TestQuorumConfigValidation(t *testing.T) {
 							MaxFlushSize:               NewByteSize(16000000),
 							MaxFlushThreads:            8,
 						},
+						SyncScheduler: SyncSchedulerConfig{
+							MaxWorkers: 32,
+						},
 						SegmentCompactionPolicy: SegmentCompactionPolicy{
 							MaxBytes:           NewByteSize(32000000),
 							MaxParallelUploads: 4,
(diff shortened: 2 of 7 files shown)


czs007:

Adversarial review of this PR surfaced one correctness gap in the external-config merge path plus two smaller documentation/readability issues in the newly added `SyncScheduler` code.

## Medium

**`cmd/external/user_config.go:335` — external config merge omits `Logstore.SyncScheduler.MaxWorkers`**
The logstore block in `mergeWoodpeckerConfig` ends at `SegmentSyncPolicy.MaxFlushThreads` (`:335-337`) and continues straight into `// Logstore - SegmentCompactionPolicy` (`:339`); `src.Logstore.SyncScheduler.MaxWorkers` is never copied to the destination config (no occurrence anywhere in the file). Failure: starting logstore with `-externalUserConfigFile` and `woodpecker.logstore.syncScheduler.maxWorkers: 128` in the external YAML — exactly the tuning this PR documents in `config/woodpecker.yaml` — silently discards the value, leaving the default 32 and building a 32-worker pool, so the flush-concurrency limit this PR targets is unchanged on that deployment path while the process starts without any error. Because the base config comes from `config.NewConfiguration(*configFile)` (`cmd/main.go:105`), defaults are already populated, so the symptom is a silent fall-back to 32 rather than a startup failure.
Suggestion: add the merge line alongside the neighbouring logstore fields, and guard it with `> 0` as they are — `ApplyToConfig` (`user_config.go:45-67`) does not call `Validate()`, and `cmd/main.go:115` does not re-validate afterwards, so an unguarded assignment would let a partial external YAML write a `0` that nothing re-checks and that `NewSyncScheduler`'s `<= 0` fallback would mask as the default. *(raised by czs007, sijie-ni-0214)*

## Low

**`server/storage/stagedstorage/sync_scheduler.go:78` — queue-depth comment understates in-flight jobs per writer**
The new comment (`:82-84`) states that each writer can have at most one schedule request and one submitted sync job outstanding, concluding that "one slot per writer is the real ceiling". There are two independent paths into `jobCh`, gated by two separate atomics: `startFlushTasksQueueProcessing` (`writer_impl.go:531` CAS on `flushTasksQueueProcessing`, `:537` `tryEnqueueJob`) and `enqueueScheduledSyncJob` (`writer_impl.go:601` CAS on `syncTaskSubmitted`, `:621` `tryEnqueueJob`). These are the only two `tryEnqueueJob` call sites in the package, so the per-writer ceiling on `jobCh` is 2, not 1 (`scheduleCh` is single-gated by `syncScheduled` and is indeed 1).
Suggestion: correct the comment to state a per-writer ceiling of two `jobCh` entries. The sizing conclusion is unaffected — 65,536 slots still require roughly 32k concurrently active writers on a single node to fill — but the stated derivation is off by a factor of two. *(raised by czs007)*

## Nitpick

**`common/config/configuration.go:821` — new validation sits under the `// Validate SegmentCompactionPolicy` heading**
In `validateLogstoreConfig`, the new `SyncScheduler.MaxWorkers` check was inserted between the `// Validate SegmentCompactionPolicy` section comment and the first `SegmentCompactionPolicy` check it labels, so that comment now covers a check that does not belong to it.
Suggestion: move the new check above the comment, or give it its own `// Validate SyncScheduler` section. *(raised by czs007)*

PR #326, opened 2026-09-11
fix(segment): index the fence LAC by the replicas that may be missing it

### LINKED ISSUE #320: [Bug] Partial Fence replies lower the completion boundary and lose acknowledged data after compaction
## Summary

A fresh writer can recover an acknowledged segment as empty when one data-bearing replica's Fence RPC fails. With `Es=3, Wq=3, Aq=2`, durable tails `[0, 0, -1]` become successful Fence replies `[0, -1]`; `resolveFenceLAC` selects `-1`. Completion succeeds and publishes `Completed.LastEntryId=-1`. Compaction then seals an empty object, and a fresh public SDK reader cannot retrieve the acknowledged entry.

Reproduced **6/6 times** against real Woodpecker services: two runs, each rotating the empty replica and failed-Fence replica through all three nodes. This is an isolated service reproduction; occurrence in a production Milvus deployment has not been established.

## Environment

- Production source: `966a5d4a0bcc855227e85788588688db4920f87d`.
- Kubernetes v1.34.1, Linux/amd64; three LogStore processes with separate persistent volumes, etcd and MinIO, deployed with Helm.
- BuildKit-built image digest: `sha256:da95b04a61fd33db1042d82e2c502c9155ee14144b7bce2861d8e401076b24b6`.
- RPC fault proxies reject selected requests; successful responses come from real servers. Production code was unchanged.

## Reproduction

1. Create a fresh log. Reject append RPCs to node C; allow nodes A and B to persist entry `(segment=0, entry=0)`. Wait for application success and durably record the returned ID and payload SHA-256 independently.
2. Exit the writer process without closing the writer, leaving the segment Active. Keep the append fault on C and reject Fence RPCs to B; allow all Complete and Compact RPCs.
3. Open a writer in a fresh SDK process to trigger recovery. A returns Fence tail `0`, C returns `-1`, and B returns an RPC error. Recovery selects `-1`.
4. All three replicas finalize successfully. Their returned local tails `[0, 0, -1]` all satisfy `localTail >= targetLAC`, because the target is `-1`. Metadata becomes Completed at `-1`.
5. Run auditor maintenance. Nonempty replicas reject the empty compaction boundary; C successfully compacts an empty segment. Metadata becomes Sealed at `-1`.
6. Open a fresh SDK reader at `(0,0)` and compare against the independent ACK ledger. The read times out and reports `lost_ack`.

An immediate local-replica read before compaction can still succeed, so it does not establish ACK preservation through recovery and sealing. No process restart is required for this reproduction.

## Evidence

Both runs on 2026-09-10, starting at 07:10:38 UTC and 07:29:06 UTC, reproduced all three replica rotations. Each case had successful write/recovery exits and a failing post-compaction ACK check.

Representative application-level failure:

```json
{"error":"context deadline exceeded","event":"lost_ack","expected":{"id":{"SegmentId":0,"EntryId":0},"sha256":"6489dfdb74c4fd01a5660b375b28e3f0815c940d34252ccd38d22a2e6a7f441c"},"index":0}
```

The object reader parsed `indexRecordCount=0` and `expectedTotalBlocks=0` for this sealed segment. Recorded Fence replies, Complete replies, metadata revisions and ACK ledgers corroborate the sequence. A bounded TLA+ model independently finds an `AcknowledgedPrefixPreserved` violation; the real-service result is the implementation evidence.

## Cause and related issues

[`resolveFenceLAC`](https://github.com/zilliztech/woodpecker/blob/966a5d4a0bcc855227e85788588688db4920f87d/woodpecker/segment/segment_handle.go#L1510-L1626) takes the Aq-th largest successful reply. A Fence responder quorum intersects the original ACK quorum, but need not contain Aq copies of the acknowledged entry. Missing responses cannot establish that the acknowledged suffix is absent.

A fresh handle has no knowledge of the previous writer's in-memory confirmed LAC. [`completeSegmentQuorum`](https://github.com/zilliztech/woodpecker/blob/966a5d4a0bcc855227e85788588688db4920f87d/woodpecker/segment/segment_handle.go#L1247-L1335) proves replica coverage only of the already-lowered target.

Related: #252, #249 and #242. Here completion qualification and compaction boundary validation are present, but recovery has already selected an unsafe boundary before either check runs.

## Expected behavior / regression criteria

- Recovery must preserve every application-acknowledged entry through completion and sealing; insufficient evidence must not publish a lower boundary.
- Cover the history above across all three replica rotations, checking ACK IDs and payload hashes after compaction with a fresh reader.
- Retain controls for truly empty segments and unacknowledged minority-only writes, so a fix also addresses recovery liveness. Selecting the maximum observed tail alone is not established as a correct fix.

tinswzy:
## Problem

`resolveFenceLAC` indexed the sorted fence replies at `len(results)-ackQuorum`, which is the
acknowledged boundary only when every replica answered. One failed reply shifts the index down a
position and publishes a boundary below data the client was told was durable.

Fixes #320

With tails `[0,0,-1]` and `Aq=2`, entry 0 had reached ack quorum. A failed fence RPC to the second
replica leaves replies `[0,-1]`, and `len-Aq` selects `-1`. Completion succeeds at `-1`, compaction
seals an empty object, and a fresh reader cannot retrieve the acknowledged entry.

## The index

The boundary is a fact about the replicas, not about the replies. An entry is acknowledged once at
least `Aq` replicas hold it, so at most `Wq-Aq` of them can be without it — the same tolerance the
append path spends before giving up at `Wq-Aq+1` failures (`segment_handle.go:444`). A replica that
did not answer is unknown, not empty, so that count does not depend on how many answered, and
neither does the index:

```go
missingTolerated := writeQuorum - ackQuorum
return sorted[missingTolerated], nil
```

`len(results)-ackQuorum` equals this only when `len(results) == Wq`.

`Wq` rather than `Es` because `quorumInfo.Nodes` is the write set — the append path already
expresses its tolerance as `Wq-Aq+1` over the same list, indexed by the same `serverIndex`. The two
are equal today (`GetWriteQuorumSize` returns `GetEnsembleSize`), but `Wq` is the quantity the rule
is actually about.

## This reverses a deliberate trade, and that should be reviewed as such

The previous rule guaranteed the target was covered by `ackQuorum` **responders**, so completion
could always finalize. `TestResolveFenceLACCoverageInvariant` asserted exactly that, with the
rationale that a target above it "can never be met — a behind replica has no way to catch up to an
entry it never received — so completion would retry forever and wedge the log". A table case
records `[0,-1,-1]` as a production wedge that the current rule was written to fix.

The two properties are incompatible once a reply is missing: proving coverage among responders
requires treating silence as absence, which is what loses acknowledged data. This change prefers to
fail and retry rather than lower the boundary.

**The earlier wedge is not reintroduced.** It had all three replicas answering, where old and new
agree — `[0,-1,-1]` still resolves to `-1`. The loss only ever occurred when a reply was missing,
so the two cases do not overlap. There is a test for each, named for which one it is.

## Second commit: completion had to change with it

The open question in the first commit — what happens when a completion now fails — turned out to
have a bad answer, and adversarial review found it independently. Completion is **not** free of side
effects, so "fail and retry" is not available:

- `completeSegmentQuorum` sends `CompleteSegment(lac)` to every node and counts qualifying replicas
  only afterwards (`1259-1261`, `1288`).
- `segmentProcessor.Complete` calls `writer.Finalize(ctx, lac)` without checking `lac` against the
  local tail (`server/processor/segment_processor.go:200`), and in service mode that writer is
  always a `StagedFileWriter`.
- `StagedFileWriter.Finalize` burns the requested LAC into a durable footer even on a replica whose
  tail is below it (`stagedstorage/writer_impl.go:1252`), and refuses any later `Finalize` at a
  different LAC with `ErrInvalidLACAlignment` (`1175-1181`). The footer survives restart (`2183`).

So a completion that fails the old `qualifiedResults >= Aq` check leaves the replicas finalized at
the target while metadata stays Active. The next fence gets every reply, resolves lower, and every
finalized footer rejects that lower target. The segment can never close — and `OpenLogWriter`
fences and completes every active segment before handing back a writer (`log/log_handle.go:268`,
`:298`), so one wedged segment stops writes on the whole log.

The check is now `len(finalizedResults) >= ackQuorum`.

`Aq` is the threshold a write had to clear to be *acknowledged*. It is a fact about the write path,
not a requirement on how the data is distributed once the writer is gone, and reading it as the
latter is what made the check unsatisfiable. What completion has to establish is that enough
replicas durably recorded *this* boundary, so a competing writer cannot close the segment at a
different one — that is the finalized count. Entries between a replica's own tail and the target
were never acknowledged, so nothing is owed on them; they become durable when compaction uploads
the segment. Coverage below `Aq` is logged as a warning instead of failing.

**Corner case this leaves, filed as #327:** if the replica holding the target dies between the fence
and the complete, the segment closes at a boundary no surviving replica covers — compaction cannot
run and a reader waits at the tail until that replica returns or an operator forces it. No
acknowledged entry is lost. The alternative is the write outage above, in a case a three-replica
quorum is supposed to absorb.

## Also fixed

A malformed quorum whose ack requirement exceeds its write quorum makes `Wq-Aq` negative and would
have indexed before the start of the slice — a panic in the recovery path. Config derives `Aq` as
`Wq/2+1` and cannot produce one, but it is now refused explicitly.

I found this because `TestFenceSegmentQuorum_FewerNodesThanAckQuorum` built exactly such a quorum
(`Es=1, Wq=1, Aq=2`). That was not carelessness: with a valid quorum every fence result is either an
error or a success, so too few successes always implies at least one node error, and fence prefers
the node error as the root cause — making `return -1, resolveErr` unreachable from
`fenceSegmentQuorum`. Fabricating an impossible quorum was the only way to reach it.

That test now uses a valid three-replica quorum with two failing nodes and asserts the node error
surfaces, and the guards are covered directly on `resolveFenceLAC`, where they are reachable.

## Tests

- `TestResolveFenceLACNeverDropsAcknowledgedData` walks every combination of replica tails and
  answering subsets for write quorums 3 through 7, asserting the selected target is never below the
  acknowledged boundary, and is exactly it when every replica answers — which also pins that the
  earlier wedge stays fixed. Odd and even write quorums both appear because they differ: at `Wq=4`
  the tolerance is 1 while `ackQuorum-1` would be 2. Only 3 and 5 are configurable today.
- `TestResolveFenceLACIssue320` is the reported history, with the empty replica and the lost reply
  rotated.
- `TestResolveFenceLACRejectsInconsistentQuorum` covers the malformed-quorum guard.
- The table test gains a `writeQuorum` column. Two cases change expectation, and their comments
  changed with them — one previously read "Entry 0 lives on a single replica, so it was never
  acked", which is an inference the evidence does not support when only two of three replied.

- `TestFenceAndComplete_PartialFenceDoesNotWedgeFinalizedReplicas` drives fence then complete
  against replicas that model `StagedFileWriter`'s footer — `CompleteSegment` records a LAC and
  refuses a later one that differs. Tails `[1,0,0]`, `Wq=3`/`Aq=2`, one replica silent at fence. It
  asserts the first pass completes, then runs the counterfactual second pass with every reply
  present (target drops to 0) and asserts those footers reject it, which is the wedge made visible.
- The complete response matrix gains the five cells this flips. `PPP` is documented as unreachable:
  the target always comes from a fence reply, so some replica reported a tail at or above it, and a
  replica's tail never shrinks.
- `TestCompletionManager_CompleteRetriesExhaustedDoesNotPublishCompleted` (from #250) keeps its
  subject — no Completed metadata when completion never lands — but its failure is now no replica
  finalizing rather than a finalized replica reading short. The latter is the case it was written
  for and is no longer a failure; it moved to
  `TestCompletionManager_PartialReplicaFinalizeStillPublishesCompleted`.

Checked that the property test fails against the previous implementation, so it is load-bearing
rather than merely green. Its first counterexample is the reported history:

```
tails=[-1 0 0] replies=[-1 0]: selected -1, below the acknowledged boundary 0
```

## Verification

- `go test ./woodpecker/segment/` and `go test -race ./woodpecker/segment/`: ok
- `golangci-lint run ./woodpecker/...` with the CI-pinned v2.11.4: 0 issues

This addresses the recovery boundary and the completion check that depends on it. The issue's other
asks — controls for truly empty segments and for unacknowledged minority-only writes — are not
covered here.

### DIFF
--- a/woodpecker/segment/completion_manager.go
+++ b/woodpecker/segment/completion_manager.go
@@ -120,7 +120,7 @@ func (cm *completionManager) executeWithRetry(ctx context.Context) error {
 	}
 
 	// Phase 3 (with lock): CAS, fast-fail ops, stop executor, update metadata.
-	// Reaching this point proves Aq finalized replicas locally cover targetLAC.
+	// Reaching this point proves Aq replicas finalized at targetLAC.
 	s.Lock()
 	defer s.Unlock()
 	closeErr := s.doCloseWritingAndUpdateMetaIfNecessaryUnsafe(ctx, targetLAC)
@@ -147,7 +147,8 @@ func (cm *completionManager) failAndCloseWithoutMeta(ctx context.Context, target
 	s.NotifyWriterInvalidation(ctx, fmt.Sprintf("segment:%d complete failed after retries", s.segmentId))
 
 	// Stop this local writer so it cannot accept more appends, but do not publish
-	// Completed metadata: the required Aq qualified replicas were not established.
+	// Completed metadata: fewer than Aq replicas finalized at targetLAC, so nothing
+	// establishes where the segment ends.
 	s.Lock()
 	s.doCloseWritingWithoutMetaUnsafe(ctx, targetLAC)
 	s.Unlock()
--- a/woodpecker/segment/completion_manager_test.go
+++ b/woodpecker/segment/completion_manager_test.go
@@ -157,20 +157,17 @@ func TestCompletionManager_AllRetriesExhausted(t *testing.T) {
 }
 
 // TestCompletionManager_CompleteRetriesExhaustedDoesNotPublishCompleted covers
-// the case where fence establishes a target LAC but every Complete pass has too
-// few replicas whose returned local tail covers that target. The local writer is
-// stopped, while metadata must remain Active for recovery rather than advertising
-// an uncompacted Completed segment.
+// the case where fence establishes a target LAC but no replica ever finalizes at
+// it. The local writer is stopped, while metadata must remain Active for recovery
+// rather than advertising a segment no replica has closed.
 func TestCompletionManager_CompleteRetriesExhaustedDoesNotPublishCompleted(t *testing.T) {
 	seg, mockClient := newTestSegmentForCompletion(t)
 
 	const targetLAC = int64(5)
 	mockClient.EXPECT().FenceSegment(mock.Anything, mock.Anything, mock.Anything, int64(1), int64(1)).
 		Return(targetLAC, nil).Once()
-	// Aq=1 in this fixture, but the only replica is a successfully finalized
-	// partial replica and therefore does not qualify.
 	mockClient.EXPECT().CompleteSegment(mock.Anything, mock.Anything, mock.Anything, int64(1), int64(1), targetLAC).
-		Return(targetLAC-1, nil).Times(3)
+		Return(int64(-1), assert.AnError).Times(3)
 
 	invalidated := false
 	seg.SetWriterInvalidationNotifier(context.Background(), func(ctx context.Context, reason string) {
@@ -185,6 +182,33 @@ func TestCompletionManager_CompleteRetriesExhaustedDoesNotPublishCompleted(t *te
 	assert.Equal(t, int64(-1), seg.segmentMetaCache.Load().Metadata.LastEntryId)
 	writable, _ := seg.IsWritable(context.Background())
 	assert.False(t, writable)
+}
+
+// TestCompletionManager_PartialReplicaFinalizeStillPublishesCompleted pins the
+// contract completion switched to in #320: an ack quorum of replicas must have
+// durably recorded the target LAC, not covered it locally.
+//
+// The target comes out of the fence, so some replica reported a tail at or above
+// it. That replica can be gone by the time CompleteSegment runs, leaving only
+// replicas whose own tail is short. Those entries were never acknowledged, so
+// nothing is owed on them, and refusing to publish Completed here would be worse
+// than publishing it: the segment stays Active, and OpenLogWriter fences and
+// completes every active segment before it returns a writer, so the whole log
+// would stop accepting writes until the missing replica came back.
+func TestCompletionManager_PartialReplicaFinalizeStillPublishesCompleted(t *testing.T) {
+	seg, mockClient := newTestSegmentForCompletion(t)
+
+	const targetLAC = int64(5)
+	mockClient.EXPECT().FenceSegment(mock.Anything, mock.Anything, mock.Anything, int64(1), int64(1)).
+		Return(targetLAC, nil).Once()
+	// Finalized at the target, but the replica's own tail stops one entry short.
+	mockClient.EXPECT().CompleteSegment(mock.Anything, mock.Anything, mock.Anything, int64(1), int64(1), targetLAC).
+		Return(targetLAC-1, nil).Once()
+
+	seg.completionMgr.TriggerCompletion()
+	require.NoError(t, seg.completionMgr.WaitForCompletion())
+	assert.Equal(t, proto.SegmentState_Completed, seg.segmentMetaCache.Load().Metadata.State)
+	assert.Equal(t, targetLAC, seg.segmentMetaCache.Load().Metadata.LastEntryId)
 }
 
 // TestSegmentHandle_SetRollingReady_PropagatesCompletionError verifies that when
--- a/woodpecker/segment/segment_handle.go
+++ b/woodpecker/segment/segment_handle.go
@@ -1244,10 +1244,11 @@ func (s *segmentHandleImpl) completePrepared(ctx context.Context, quorumInfo *pr
 	return s.completeSegmentQuorum(ctx, quorumInfo, lastAddConfirmed)
 }
 
-// completeSegmentQuorum sends CompleteSegment requests to all quorum nodes and
-// counts only finalized replicas whose returned local tail covers lac. A behind
-// replica may still finalize successfully for read failover, but cannot count
-// toward the Aq proof required before publishing Completed metadata.
+// completeSegmentQuorum sends CompleteSegment requests to all quorum nodes and requires
+// ackQuorum of them to finalize at lac before Completed metadata may be published. A
+// replica whose local tail stops below lac still counts: it has durably recorded the
+// boundary, which is what stops a competing writer from closing the segment elsewhere.
+// How many replicas locally cover lac is tracked separately and reported, not enforced.
 func (s *segmentHandleImpl) completeSegmentQuorum(ctx context.Context, quorumInfo *proto.QuorumInfo, lac int64) error {
 	nodeCount := len(quorumInfo.Nodes)
 	ackQuorum := int(quorumInfo.Aq)
@@ -1565,9 +1566,10 @@ func (s *segmentHandleImpl) resolveFenceLAC(results []int64, writeQuorum, ackQuo
 	// (#320).
 	//
 	// Because a missing reply is not evidence, a target chosen this way may turn out to be held by
-	// fewer than ackQuorum replicas. completeSegmentQuorum rejects that rather than lowering the
-	// boundary, and the attempt is retried once more replicas answer: failing to complete is
-	// recoverable, publishing a boundary below acknowledged data is not.
+	// fewer than ackQuorum replicas. That is allowed: completeSegmentQuorum requires ackQuorum
+	// replicas to finalize at the target, not to cover it, precisely so this target can be used.
+	// The entries above the covered tail were never acknowledged, so nothing is owed on them;
+	// publishing a boundary below acknowledged data, by contrast, is unrecoverable.
 	// A quorum where Aq exceeds Wq is not satisfiable -- no entry can ever collect more acks than
 	// the replicas it was written to -- and would make the index below negative. Config derives
 	// Aq as Wq/2+1 so this cannot arise from a valid configuration, but recovery must not panic
--- a/woodpecker/segment/segment_handle.go
+++ b/woodpecker/segment/segment_handle.go
@@ -1302,9 +1302,25 @@ func (s *segmentHandleImpl) completeSegmentQuorum(ctx context.Context, quorumInf
 		}
 	}
 
-	// Check if we have enough successful responses for ack quorum
-	if len(qualifiedResults) < ackQuorum {
-		logger.Ctx(ctx).Warn("Insufficient qualified responses for quorum complete",
+	// Require the boundary to be durably recorded by an ack quorum of replicas, not to be
+	// covered by one.
+	//
+	// Aq is the threshold a write had to clear to be acknowledged. It says nothing about how the
+	// data is distributed once the writer is gone, and treating it as a requirement on coverage
+	// at recovery time is what made this check unsatisfiable: resolveFenceLAC returns the highest
+	// tail that cannot be shown to be below the acknowledged boundary, and when a fence reply is
+	// missing that tail may be held by a single replica. The entries in question were never
+	// acknowledged, so nothing is owed on them; they become durable when compaction moves the
+	// segment to object storage.
+	//
+	// What completion does have to establish is that enough replicas durably recorded this
+	// boundary, so a competing writer cannot close the segment at a different one. That is
+	// finalizedResults. qualifiedResults is kept for the operator-facing count below: a
+	// completion covered by fewer than ackQuorum replicas is worth seeing, but it is not a
+	// failure -- failing here would leave the log unable to open a writer at all, since
+	// OpenLogWriter fences and completes every active segment first.
+	if len(finalizedResults) < ackQuorum {
+		logger.Ctx(ctx).Warn("Insufficient finalized responses for quorum complete",
 			zap.String("logName", s.logName),
 			zap.Int64("logId", s.logId),
 			zap.Int64("segmentId", s.segmentId),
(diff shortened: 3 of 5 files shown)


czs007:

This PR changes `resolveFenceLAC` so it can select a fence target that fewer than ack-quorum replicas hold. In service mode, the completion path finalizes replica footers irreversibly before it checks quorum coverage, so that target can leave a segment permanently unable to complete. The tests do not exercise this path.

## High

**`woodpecker/segment/segment_handle.go:1566`: A fence target taken from partial replies can permanently block segment completion in service (staged storage) mode.**

`resolveFenceLAC` now returns `sorted[Wq-Aq]`, where Wq is the write quorum and Aq is the ack quorum. When a fence reply is missing, the returned LAC (last add confirmed) can be held by fewer than Aq replicas. The comment at `segment_handle.go:1539-1542` says a failed completion is harmless and gets retried. In service mode it is not harmless:

- `completeSegmentQuorum` sends `CompleteSegment(lac)` to every quorum node (`1259-1261`). It counts qualifying replicas only afterwards (`1288`, `1306`).
- `segmentProcessor.Complete` calls `writer.Finalize(ctx, lac)` without checking `lac` against the local tail (`server/processor/segment_processor.go:200`). In service mode that writer is always a `StagedFileWriter` (`segment_processor.go:564-565`).
- `StagedFileWriter.Finalize` writes the requested LAC into a durable footer, even on a replica whose tail is below it (`stagedstorage/writer_impl.go:1252`, `1260`).
- Any later `Finalize` with a different LAC is rejected with `ErrInvalidLACAlignment` (`writer_impl.go:1175-1181`). The footer is reloaded after a restart (`2183-2184`, `2225`).

**Failure scenario:** Service mode, Wq=3/Aq=2, replica tails `[1,0,0]` (entry 1 was written to one replica and never acknowledged), and the fence RPC to one tail-0 replica fails, so the replies are `[1,0]` → target 1 goes to every node.
- The other tail-0 replica durably finalizes its footer at LAC 1 and does not qualify, so qualified=1 < Aq.
- `completionManager` retries `completePrepared` against the same frozen target, then calls `failAndCloseWithoutMeta` (`completion_manager.go:114-119`), and the segment stays Active.
- Every later `FenceAndComplete` (`log/log_handle.go:400`) gets full replies and targets 0. The replicas finalized at 1 reject 0, and target 1 can never reach Aq because only one replica holds entry 1.
- The segment can never reach Completed, including after restarts, and the replica footers disagree (footer LAC 1 above local tail 0).

The old index `sorted[len-Aq]` picked 0 from `[1,0]` and completed normally.

**Suggestion:** Keep the `Wq-Aq` index, but send `CompleteSegment` only once at least Aq successful fence replies cover the selected target. If they don't, treat the fence as inconclusive and fence again without finalizing any replica. This still fixes #320, which was caused by the index choosing too low a value, not by a coverage requirement. With tails `[0,0,-1]` and replies `[0,-1]`, the target is covered by only one reply, so the node fences again and no data is lost. *(raised by sijie-ni-0214, xiaocai2333)*

## Medium

**`woodpecker/segment/segment_handle_test.go:3406`: No end-to-end test runs a partial fence followed by a full fence through completion, and one table case expects a target that only one replica holds.**

All the new tests (`TestResolveFenceLAC`, `TestResolveFenceLACNeverDropsAcknowledgedData`, `TestResolveFenceLACIssue320`, `TestResolveFenceLACRejectsInconsistentQuorum`) call `resolveFenceLAC` directly and only check that the target is never below the acknowledged boundary. None of them passes the target through `completeSegmentQuorum` and a replica that finalizes, which is where the failure happens. The PR also removes `TestResolveFenceLACCoverageInvariant`, which checked that the target is covered by at least Aq responders. The new table case `"five replicas two silent"` (replies `[9,7,5]`, Wq=5, Aq=3) expects target 9, which one replica holds and completion can never qualify.

**Failure scenario:** The high-severity scenario above (service mode, tails `[1,0,0]`, Wq=3/Aq=2, one fence RPC fails) → the permanent completion failure ships undetected, because no test combines a partial fence with completion against replicas that finalize irreversibly.

**Suggestion:**
- Add a regression test with tails `[1,0,0]`, Wq=3, Aq=2: one fence round with replies `[1,0]`, then one with all replies `[1,0,0]`.
- Use a mock client that tracks each replica's footer LAC and rejects a different LAC after finalization, as `StagedFileWriter.Finalize` does.
- Assert that no replica is finalized at the partial-reply target before Aq coverage is established, and that recovery reaches Completed.
- Restore the coverage property in the exhaustive test: whenever a target is returned, at least Aq replicas hold it, or the function returns no target.
- Fix or remove the `"five replicas two silent"` expectation.

*(raised by xiaocai2333)*

tinswzy:
Thanks — the High finding is real, and the trace through `Finalize` → durable footer → `ErrInvalidLACAlignment` is exactly right. Fixed in 20ec37e, but on the completion side rather than the fence side.

## Why not the suggested gate

The suggestion was to require Aq fence replies covering the target before sending `CompleteSegment`, and otherwise fence again. That is clean in that nothing is finalized, but it makes a *tolerable* failure permanent:

```
Wq=3, Aq=2, tails [1,0,0], one tail-0 replica permanently down
→ fence replies [1,0] → target 1, covered by 1 < Aq → "inconclusive", fence again
→ the replies never change → inconclusive forever
```

The gate trips whenever the surviving replies disagree — and if the down replica is one of the Aq that would have held the boundary, they always disagree. So the condition is not "rare partial fence", it is "single replica down with a ragged tail", which a 3-replica quorum is supposed to absorb.

And it does not stop at the segment. `OpenLogWriter` → `openInternalLogWriter` (`woodpecker/log/log_handle.go:268`, `:298`) calls `fenceAllActiveSegments(ctx)` and returns the error, so a fence that can never conclude means the **whole log can never open a writer**. That trades a corner case that loses no acknowledged data for a common case that stops all writes.

## What landed instead

`completeSegmentQuorum` now requires `len(finalizedResults) >= ackQuorum` instead of `len(qualifiedResults) >= ackQuorum`.

Aq is the threshold a write had to clear to be *acknowledged*. It is a fact about the write path, not a requirement on how data is distributed once the writer is gone. What completion has to establish is that enough replicas durably recorded *this* boundary, so a competing writer cannot close the segment at a different one — that is the finalized count. The entries between a replica's own tail and the target were never acknowledged, so nothing is owed on them; they become durable when compaction uploads the segment. Coverage below Aq is now a warning, not a failure.

In your scenario (`[1,0,0]`, replies `[1,0]`, target 1): both reachable replicas finalize at 1, `finalized=2 >= Aq=2`, the segment reaches Completed, and there is no second fence to be rejected. The wedge is gone because the first pass succeeds, not because the target was avoided.

## Remaining corner case (filed separately, not fixed here)

If the replica holding the target dies *between* fence and complete, the segment closes at a boundary no surviving replica covers: compaction cannot run and a reader waits at the tail until the replica returns or an operator forces it. No acknowledged entry is lost. This is the known limitation of the protocol — the same class of thing BookKeeper accepts when recovery resurrects an unacknowledged entry — and the alternative is the write outage above. A follow-up issue will cover the two ways out (synchronous backfill at finalize vs. deferred client-quorum compaction).

## Medium

- **Added the end-to-end test you asked for**: `TestFenceAndComplete_PartialFenceDoesNotWedgeFinalizedReplicas` (`segment_handle_test.go`). Tails `[1,0,0]`, Wq=3/Aq=2, one replica silent at fence; the mock replicas model `StagedFileWriter` — `CompleteSegment` records a footer LAC and refuses a later one that differs. It asserts pass 1 completes, then runs the counterfactual second pass with every reply present (target drops to 0) and asserts those footers reject it, which is the wedge made visible.
- **`"five replicas two silent"` is correct as written.** Replies `[9,7,5]`, Wq=5, Aq=3, so two replicas are unknown. If both hold 9, then 9 is held by 3 = Aq replicas and may have been acknowledged. 9 is the lowest target that cannot lose acknowledged data; anything lower is only safe if silence means absence, which is precisely the #320 bug. Completion no longer needs the target to be covered, so the case is consistent.
- **`TestResolveFenceLACCoverageInvariant` stays removed, deliberately.** Proving "at least Aq replicas hold the target" from the replies alone requires treating a missing reply as an empty replica, which is the thing that dropped acknowledged data. The two properties are not simultaneously satisfiable once a reply is missing. The property that replaced it — the target is never below the acknowledged boundary, and is exactly it when every replica answers — is checked exhaustively for Wq 3 through 7 in `TestResolveFenceLACNeverDropsAcknowledgedData`.

One upstream test changed subject as a result: `TestCompletionManager_CompleteRetriesExhaustedDoesNotPublishCompleted` still asserts that no Completed metadata is published when completion never lands, but its failure is now "no replica finalized" rather than "a finalized replica read short" — the latter is the case it was written for (#250) and is no longer a failure. That case moved to `TestCompletionManager_PartialReplicaFinalizeStillPublishesCompleted`.

czs007:

**Verified:**
- `segment_handle.go:1594` — `sorted[writeQuorum-ackQuorum]` is always an element of `results` and never resolves below the acknowledged boundary; the exhaustive property test (`segment_handle_test.go:531`, Wq 3–7 × all tail/reply subsets) holds, and it reduces to the old formula when every replica answers.
- `segment_handle.go:1570` and `:1583` — the `writeQuorum < ackQuorum || ackQuorum <= 0` guard and the `len(results) <= missingTolerated` refusal both prevent a negative index or an under-informed target, and cannot spuriously fire in production since `Wq` is populated on every real quorum (`quorum/discovery.go:228`) and single-node paths set `Wq:1`.
- `segment_handle.go:1322` — gating on `finalizedResults >= Aq` sacrifices no acknowledged entry, because the resolved target is always ≥ the acknowledged boundary and an acknowledged entry is by definition on ≥ `Aq` replicas; `qualifiedResults` remains live solely for the operator warning at `:1335`.
- `completion_manager.go:105-117` — fence freezes the target once and Complete retries against that same target, with already-finalized replicas returning idempotently (`stagedstorage/writer_impl.go:1175-1187`), so retries converge rather than thrash.
- `segment_handle_complete_test.go:89` — the five flipped matrix cells (QPP/QPE/PPE) follow directly from the finalized-count rule, and `PPP` is correctly documented as unreachable since a tail never shrinks and the target originates from a fence reply.

## Nitpick — non-blocking

- **`segment_handle.go:1567`** — Four comments still describe the old coverage-gated semantics ("`ackQuorum` replicas must locally cover `targetLAC`"), which the second commit in this PR inverted. Most importantly, the inline comment added in `resolveFenceLAC` at `:1567-1570` states that "completeSegmentQuorum rejects that rather than lowering the boundary, and the attempt is retried once more replicas answer" — but `:1322` now gates on `len(finalizedResults) < ackQuorum`, and coverage was demoted to the non-fatal warning at `:1335`. The PR's own new test `TestCompleteSegmentQuorum_TargetHeldByOneReplica` asserts `assert.NoError` for exactly the case the comment says is rejected, so the comment contradicts the same PR's code. The same applies to the `completeSegmentQuorum` doc comment at `:1247-1250` ("counts only finalized replicas whose returned local tail covers lac… cannot count toward the Aq proof"), and to `completion_manager.go:123` ("proves Aq finalized replicas locally cover targetLAC") and `:150` ("the required Aq qualified replicas were not established") — those two were accurate before this PR and became false when the gate changed. A reader following these comments will infer the opposite control flow from what `:1322` implements. Suggestion: update `:1567-1570` and `:1247-1250` in this PR, since they are the self-contradictory ones; the two `completion_manager.go` comments can be corrected here as well or in a small follow-up, since that file is otherwise untouched by this PR. *(raised by bigsheeper)*

tinswzy:
Nitpick taken — fixed in 7fdb23c. All four comments described the coverage gate the second commit replaced, and the one in `resolveFenceLAC` contradicted this PR's own `TestCompleteSegmentQuorum_TargetHeldByOneReplica`. Both `completion_manager.go` comments are corrected here rather than deferred, since they are two lines and the file is meaningless to read against the new gate otherwise.

PR #329, opened 2026-09-11
enhance(logstore): bound what a node takes on, and watch its memory while doing it

### LINKED ISSUE #328: enhance(logstore): a node takes on unbounded concurrent compaction, with nothing watching its memory
Every bound on compaction today is per-segment or per-log. Nothing bounds a **node**, which is where the memory is held.

This is the second half of the pair opened with #319 (merged as #322); that one removed a copy from each in-flight merged block, this one bounds how many of them a node takes on at once.

## What is bounded today, and what is not

| Scope | Bound | Where |
|---|---|---|
| one segment | one compaction at a time | `segmentProcessor.Compact` CAS guard |
| one segment | the node's ceiling on its duration | `segmentCompactionPolicy.timeout` |
| one segment | blocks in flight within it | `maxParallelUploads` |
| one log, one cycle | 64 segments | `maxCompactedPerCycle` (#323) |
| one log, one cycle | a time budget | `compactionPassBudget` (#324) |
| **one node** | **nothing** | — |

All of those are client-side and per log. `Compact` is an RPC; a node holds segments belonging to many logs, each driven by its own writer's auditor on its own schedule, and the node has no say in how many arrive at once. It peaks exactly when it can least afford it: after compaction has been failing, every log has a backlog of `Completed` segments, and the moment it recovers every auditor starts draining.

## What one compaction costs

Per merged block in flight the node holds the source blocks' byte range and the assembled block handed to `PutObject` — the extracted records in between are subslices of the first, not copies, since #322, and `PutObject` adds none (with `SendContentMd5` off and a `*bytes.Reader`, minio-go wraps it in an `io.SectionReader` on both the single-PUT and multipart paths). `maxParallelUploads` of those run at once:

```
one compaction ≈ maxParallelUploads × 2 × merged block size
               = 4 × 2 × 2MB = 16MB      (shipped defaults)
```

A segment with fewer merge tasks than the upload pool is wide — every small segment — costs proportionally less, since only those tasks can ever be resident.

## The gate: an absolute ceiling, not a share of the pod

**Useful compaction concurrency saturates long before memory does.** A hundred concurrent compactions is about 1.6GB here, but four hundred simultaneous object-storage uploads; the network gives out first. So a proportional ceiling on a large pod hands out an allowance nothing can use:

| pod | 70% of it | actually reachable |
|---|---|---|
| 4GB | 2.8GB | ~1GB |
| 16GB | 11GB | ~1GB |
| 64GB | 45GB | ~1GB |

An absolute number covers every realistic concurrency without being tuned per pod, and — unlike a ratio — **cannot be inflated by a misread limit**. That matters, because `hardware.GetMemoryCount()` falls back to the host's total when the cgroup limit cannot be read, and returns it indistinguishably from a real limit:

```go
limit, err := getContainerMemLimit()
if limit > 0 && limit < stats.Total { return limit }
if err != nil { ...Warn... }
return stats.Total          // ← host memory
```

On a 4GB pod on a 256GB machine that is a ceiling 45× too large, on every admission, for the life of the process. A ratio-based design has that failure at its centre; an absolute one does not have it at all.

## The second gate: memory pressure

An absolute ceiling knows only about compaction. It cannot see that the node is already at 90% because of write buffers, and compaction added on top is what tips it over. So a second check refuses new work once the node's memory usage passes a watermark, whatever is using it.

This one does need the limit, so it runs **only where a container limit can actually be read**. Where it cannot — bare metal, no limit set, or detection failure — it is skipped rather than approximated, because approximating it means the misread-limit failure above. Telling those three apart is unnecessary: the conservative answer is right for all of them.

Degrading that way is the point: the absolute ceiling holds on its own, so an unreadable limit costs the pressure check and nothing else.

## Why both, and why neither alone

They fail in opposite directions, and each covers the other's blind spot:

- **Reservation alone** is blind to everything that is not compaction. Writes at 60% plus a full compaction allowance is 130%.
- **Live memory alone** is a lagging signal. Callers arriving together all read the same low value and are all admitted; memory only climbs afterwards. Admission and allocation are not atomic, so a reservation that moves the instant a compaction is taken is the only thing that stops concurrent overshoot.

Three cases, and what catches each:

| | reservation gate | pressure gate |
|---|---|---|
| estimate ≈ actual | bounds it | confirms it |
| estimate > actual | over-refuses, costs throughput | idle |
| **estimate < actual** | admits too much | **catches it** — usage climbs and the next admission is refused |

The third is the one that matters. Note what the pressure gate can and cannot do there: it stops the *next* admission, not the ones already taken. A systematically low estimate is not made safe by it; it is made *self-limiting*.

## Refuse, don't queue

A compaction over either bound is refused, not waited for. Queuing would spend the caller's per-attempt deadline on a wait, and the caller has better things to do with it: `compactSegmentQuorum` walks on to the next replica, which may have room, and failing that the auditor retries next cycle with the segment still `Completed` and nothing lost.

The refusal must be retryable so `classifyCompactionFailureReason` labels it `transient` rather than `other` — a node at capacity is not an unexplained failure, and that is the distinction `WpSegmentCompactionFailuresTotal` exists to make.

## One compaction is always taken

Whatever either gate says, a node running none takes the next one. A ceiling below a single compaction's footprint, or memory over the watermark because of something else entirely, would otherwise stop compaction on this node for good: no segment leaving local disk, no space reclaimed, and eventually the disk watermark throttling writes as well. Briefly exceeding a bound is the lesser failure.

## Configuration

```yaml
logstore:
  segmentCompactionPolicy:
    maxInflightMemory: 1GB        # reservation ceiling, absolute
    memoryHighWatermark: 0.7      # pressure gate, skipped without a trustworthy limit
```

Neither needs tuning per pod: the first saturates, the second is a fraction with physical meaning.

## Observability

Refusals are counted separately by which gate refused them — the ceiling means compaction is at capacity, pressure means the node is busy with something else, and they call for different responses. Reserved bytes, its peak, and the running count are exported alongside; the peak is what says whether the ceiling is close to binding, since a 15s scrape of the current value misses it.

**Separately worth doing:** `compactStats.deferred` — how many segments a cycle left undone — exists only as a log field today. It is the most direct "compaction is falling behind" signal and cannot be alerted on as it stands.

## Not in scope

`maxParallelReads` is unused by staged storage after #322 but still live for object storage, so it is left alone.

tinswzy:
## Problem

Every bound on compaction today is per segment or per log. None is per **node**, which is where the
memory is held.

Fixes #328

| Scope | Bound |
|---|---|
| one segment | one compaction at a time; the node's duration ceiling; `maxParallelUploads` blocks in flight |
| one log, one cycle | 64 segments (#323); a time budget (#324) |
| **one node** | **nothing** |

Those are all client-side and per log. `Compact` is an RPC; a node holds segments belonging to many
logs, each driven by its own writer's auditor on its own schedule. It peaks exactly when the node
can least afford it — after compaction has been failing, every log has a backlog and every auditor
starts draining the moment it recovers.

## Two gates, failing in opposite directions

**Primary: an absolute ceiling on reserved bytes.** Deliberately not a fraction of the pod, for two
reasons.

Useful concurrency saturates long before memory does — a hundred concurrent compactions is ~1.6GB
here but **four hundred simultaneous object-storage uploads**, and the network gives out first. A
proportional ceiling on a large pod hands out an allowance nothing can use:

| pod | 70% of it | actually reachable |
|---|---|---|
| 4GB | 2.8GB | ~1GB |
| 64GB | 45GB | ~1GB |

And an absolute number **cannot be inflated by a misread limit**, which a ratio can:

```go
// hardware.GetMemoryCount
limit, err := getContainerMemLimit()
if limit > 0 && limit < stats.Total { return limit }
if err != nil { ...Warn... }
return stats.Total          // ← host memory, indistinguishable from a real limit
```

On a 4GB pod on a 256GB machine that is a ceiling 45× too large, on every admission, for the life of
the process. A ratio-based design has that failure at its centre.

**Secondary: memory pressure.** The reservation total knows only about compaction; it cannot see
that writes already have the node at 90%. So a live reading (`GetUsedMemoryCount`, benchmarked at
**3.8µs** — cheap enough per admission) refuses new work past a watermark, whatever is using the
memory.

This one needs the limit, so it runs **only where a container limit can actually be read**.
`trustedMemoryLimit` treats a value that is not below host memory as no limit at all, which lumps
together bare metal, a pod with no limit, and a detection failure — the three do not need
separating, because the conservative answer is right for all of them. The absolute ceiling holds on
its own, so an unreadable limit costs the pressure check and nothing else.

## Why neither alone

- **Reservation alone** is blind to everything that is not compaction. Writes at 60% plus a full
  compaction allowance is 130%.
- **Live memory alone** is a lagging signal. Admission and allocation are not atomic: callers
  arriving together all read the same low value and are all admitted, and memory only climbs
  afterwards. Only a reservation that moves the instant a compaction is taken stops that.

| | reservation gate | pressure gate |
|---|---|---|
| estimate ≈ actual | bounds it | confirms it |
| estimate > actual | over-refuses, costs throughput | idle |
| **estimate < actual** | admits too much | **catches it** |

On the third: the pressure gate stops the *next* admission, not the ones already taken. A
systematically low estimate is not made safe by it — it is made self-limiting.

## Charged from the plan, not a flat rate

`Writer.CompactionMemoryEstimate` runs the same planner `Compact` does, so a segment holding a few
hundred KB is charged for itself rather than for a pool's width of full blocks. A flat rate would be
wrong twice: it is set by the merged block size, and the two shipped values differ by 16×
(`woodpecker.yaml` 2MB, in-code default 32MB).

## Refuse, don't queue — and always take the first

A compaction over either gate is refused, not waited for: `compactSegmentQuorum` walks on to the
next replica and the auditor retries next cycle with the segment still `Completed`. The refusal is a
retryable `woodpeckerError`, so `classifyCompactionFailureReason` labels it `transient` rather than
`other` — a node at capacity is not an unexplained failure.

Whatever either gate says, a node running none takes the next one. A ceiling below one compaction's
footprint, or memory over the watermark for reasons of its own, would otherwise stop compaction on
this node for good: no segment leaving local disk, no space reclaimed, and eventually the disk
watermark throttling writes as well.

## Configuration

```yaml
logstore:
  segmentCompactionPolicy:
    maxInflightMemory: 1000000000   # reservation ceiling, absolute
    memoryHighWatermark: 0.7        # pressure gate; skipped without a trustworthy limit
```

Neither needs tuning per pod: the first saturates, the second is a fraction with physical meaning.
Both are overridable from an embedding deployment, for the same reason `maxBytes` and
`maxParallelUploads` are — raising either of those raises what one compaction costs.

## Observability

`compaction_admission_{running,max_inflight_bytes,reserved_bytes,peak_reserved_bytes}` plus
`compaction_admission_rejected_total`, counted separately by which gate refused: the ceiling means
compaction is at capacity, pressure means the node is busy with something else, and they call for
different responses. The peak is what says whether the ceiling is close to binding — a 15s scrape of
the current value misses it. What was resolved, and whether the pressure gate is active at all, is
logged once at startup.

## Verification

- `go test ./common/config/ ./common/werr/ ./common/hardware/ ./server/... ./cmd/external/`: ok
- `go test -race ./server/processor/`: ok
- `golangci-lint run ./server/... ./common/... ./cmd/...`: 0 issues
- `gofmt -l .` clean, `go mod tidy` no diff

Tests worth calling out:

- `TestMemoryAdmission_ConcurrentAcquireNeverExceedsTheCeiling` — 64 goroutines under `-race`,
  tracking the peak reservation; it also asserts the run admitted something, so a gate that refused
  everything could not pass by doing nothing.
- `TestMemoryAdmission_PressureGate` — refuses over the watermark, admits under it, and is *skipped*
  rather than closed without a trustworthy limit.
- `TestMemoryAdmission_AlwaysTakesTheFirst` — both the misconfigured-ceiling and the
  memory-already-high cases.
- `TestTrustedMemoryLimit` — an untrusted limit is reported as none, **never as host memory**.
- `TestMemoryAdmission_StrayReleaseDoesNotRaiseTheCeiling` / `_IdleResetsTheReservation` — the
  failure a credit-based bound has and a mutex does not.
- `TestCompactionMemoryEstimate_TracksTheSegmentNotTheConfig` and `TestCompactionPeakBytes` — the
  estimate follows the segment, and is pinned at both shipped block sizes.
- `TestSegmentProcessor_Compact_SmallSegmentChargesLess` — three small segments share a ceiling that
  would have fitted exactly one under a flat rate.

## Follow-ups, not in this PR

- **`compactStats.deferred`** — how many segments a cycle left undone — exists only as a log field.
  It is the most direct "compaction is falling behind" signal and cannot be alerted on as it stands.
  Filed against #328.
- **The 2MB / 32MB default mismatch** between `woodpecker.yaml` and `configuration.go`: the
  plan-based estimate absorbs it, but the two should still agree, and which is right is a separate
  call.
- **Milvus has no paramtable key** for either setting, so an embedding deployment gets the defaults
  and cannot tune them — while it *can* tune `maxSize` and `maxParallelUploads`, which raise what
  one compaction costs.


czs007:

**Verified:**
- `common/config/configuration.go:492-522`: defaults are applied before the YAML is unmarshalled. An existing deployment YAML without `maxConcurrentCompactions` keeps the default of 8 and passes the new non-positive check, so upgrades are not broken.
- `server/processor/segment_processor.go:650-669`: the per-segment `compacting` claim is taken before `TryAcquire`. On refusal, the deferred `Store(false)` frees the claim, so the next cycle gets `NodeBusy` again instead of `AlreadyCompacting`. `Release` is deferred right after a successful acquire, so the slot is returned on every exit path, including writer-creation failure, a `writer.Compact` error and a panic.
- `server/processor/compaction_scheduler.go:71-93`: the semaphore is a buffered channel with non-blocking send and receive. It has no blocking path, so it cannot deadlock. A stray `Release` on an empty semaphore does nothing, so it cannot raise the limit.
- `server/service.go:634` → `common/werr/utils.go:87,100`: the retryable flag on `ErrSegmentCompactionNodeBusy` (code 2112) survives the RPC round trip. The client moves on to the next replica without dropping the cached gRPC connection (`woodpecker/client/logstore_client_remote.go:77`).
- `woodpecker/segment/segment_handle.go:2051-2060`: a busy refusal returns before the metadata update. `log_auditor.go:140-143` counts it and moves on. The segment stays `Completed` and is retried in the next auditor cycle.
- `server/logstore.go:377-379`: this is the only production caller of `NewSegmentProcessor`, and it now passes both `WithSyncScheduler` and `WithCompactionScheduler`. Existing test callers compile unchanged.

czs007:

The two findings below are both low severity, and both concern how the new compaction memory budget handles unusual inputs: an out-of-range ratio from the external config, and a test that depends on the runner's memory limit.

**Verified:**
- `server/processor/compaction_scheduler.go:164`: admission and charging both happen under `mu`, so concurrent admissions cannot go over the budget.
- `server/processor/segment_processor.go:662,673`: each release is paired with its charge through the captured `compactionEstimate`, so the in-flight byte count is decremented by the same amount it was incremented.
- `server/processor/segment_processor.go:650-656`: the per-segment CAS is taken first and cleared by its defer on refusal, so a refused compaction leaves the segment guard free.
- `server/logstore.go:651` and `writer_registry.go:79`: these are the only callers of `writer.Compact` (the RPC and admin `ForceCompact`), and both go through the budget check.
- `segment_handle.go:1734-1749`: on refusal the client moves to the next replica with no retry loop. `logstore_client_remote.go:338` keeps the retriable flag when it rebuilds the error, so code 2112 is classified as `transient`.

## Low (non-blocking)

- **`cmd/external/user_config.go:355`**: `maxMemoryRatio` from the external user config is never range-checked.
  - **Problem:** the new merge accepts any `MaxMemoryRatio > 0`.
    - `ApplyToConfig` runs in `cmd/main.go:115`, after `NewConfiguration` has already called `Validate()`, and nothing validates the config again after it.
    - `server.NewServerWithConfig` only calls `ValidateMinioConfig()`.
    - `resolveCompactionMemoryBudget` (`compaction_scheduler.go:111`) rejects only values `<= 0`.
    - As a result, the `(0,1]` check at `configuration.go:851` never runs on this path.
  - **Failure:** the external user config sets `maxMemoryRatio: 10` (meaning 10%) and leaves `maxMemoryBytes` at 0 → the node starts without error and the budget becomes 10 × the node/cgroup memory limit. The startup log still says `source=node_memory_ratio`, so the node-wide memory bound this PR adds never applies during a backlog drain.
  - **Suggestion (any one of these):**
    - In `resolveCompactionMemoryBudget`, treat `MaxMemoryRatio > 1` like `<= 0` (fall back) or clamp it to 1.
    - Reject a ratio outside `(0,1]` in `mergeWoodpeckerConfig`.
    - Call `cfg.Validate()` after `ApplyToConfig` in `cmd/main.go`. This also covers every other field merged there.
  - (raised by czs007)

- **`server/processor/compaction_scheduler_test.go:234`**: the subtest "the shipped defaults resolve to a usable budget" asserts `budget >= estimate`, and that result depends on the runner's memory limit.
  - **Problem:**
    - The test calls `config.NewConfiguration()` with no file, so it uses the in-code defaults: `MaxBytes=32000000` and `MaxParallelUploads=4` (`configuration.go:989-990`), not the `maxBytes: 2000000` shipped in `config/woodpecker.yaml`. That makes the estimate 256,000,000 bytes.
    - The budget is `0.1 × hardware.GetMemoryCount()`. That function returns the cgroup limit when it is lower than host memory (`hardware_info.go:83-84`) and also reads `MEM_LIMIT` (`container_linux.go:62`).
  - **Failure:** running `go test -race -short -failfast ./server/processor/...` in a container with a memory limit below about 2.56 GB, or with `MEM_LIMIT=2Gi` exported → the assertion at `:242` fails (budget ≈ 214 MB < estimate 256 MB) although the code has no bug, and `-failfast` skips the rest of the package.
    - The current CI (`unit-test.yaml:31`, `runs-on: ubuntu-latest`, no container) does not hit this.
    - Production behavior is unaffected, because the exemption for an idle node still admits the first compaction.
  - **Suggestion:** make the test independent of the host's memory. For example:
    - Set an explicit `MaxMemoryBytes` in the test config, or lower `MaxBytes`/`MaxParallelUploads`, so the comparison does not depend on `hardware.GetMemoryCount()`.
    - Or assert only that the resolved budget is positive and comes from the expected source.
  - (raised by czs007, bigsheeper)

czs007:

Adversarial review of this PR found two medium-severity correctness issues in the new compaction memory admission path, both introduced by this change; no low-severity or pre-existing hazards survived verification.

## Medium

- **`server/storage/stagedstorage/writer_impl.go:1654`** — The compaction memory lease is released while merge workers are still running. On the failure path the result loop returns on the first failing future (or on `ctx.Err()`) without calling `.Value()` on the remaining futures, and the deferred `uploadPool.Release()` only reaps idle workers and never waits for running ones (ants `Release`, not `ReleaseTimeout`). Since `Submit` blocks until a worker is free, up to `MaxParallelUploads` tasks can still be mid-read/merge/upload, each holding its source range and assembled merged block, when `Writer.Compact` returns. The new `defer s.compactionAdmission.Release(compactionEstimate)` in `segment_processor.go:699` then credits the full estimate back immediately, so the next `TryAcquire` can admit another compaction against memory those buffers still occupy. `MinioFileWriter.streamMergeAndUploadBlocks` (`objectstorage/writer_impl.go:1843-1887`) has the identical shape. *Failure:* a segment with more merge tasks than `MaxParallelUploads` where an early task fails (e.g. `PutObject` error on merged block 0) while the last parallel tasks are still in flight, on a node with `maxInflightBytes` near two compactions and no cgroup limit → `Compact` returns the error, the deferred `Release` frees the full estimate, the next compaction is admitted, and resident compaction memory transiently exceeds `maxInflightBytes` by up to one compaction's peak footprint; with the pressure gate disabled (no trusted limit via `trustedMemoryLimit()`), nothing else bounds it, so a backlog of repeatedly failing compactions can push the process past the operator's budget. The overshoot is bounded to roughly one compaction's footprint and drains within one task's duration, hence medium. **Suggestion:** keep draining every submitted future after the first error (or use `ReleaseTimeout` / a `WaitGroup`) so the lease is released only after the pool has quiesced; apply the same fix to the object-storage backend. (raised by sijie-ni-0214, xiaocai2333)

- **`server/processor/memory_admission.go:166`** — The pressure gate compares a cgroup limit against the process's private RSS, which are mismatched scopes. `memoryLimitBytes` comes from `trustedMemoryLimit()`, which resolves to the container limit via `hardware.GetMemoryCount()`, but `hardware.GetUsedMemoryCount()` on Linux returns `memInfo.RSS - memInfo.Shared` from the process's own `/proc` entry. The cgroup charge the OOM killer enforces also includes active/dirty page cache from the WAL write path, shmem, and any sibling processes in the pod, none of which the gate sees, so it systematically under-reads and fires late. *Failure:* a logstore pod with a cgroup memory limit under sustained append load, where charged cgroup usage (private RSS plus active page cache from local WAL writes) is above `limit * highWatermark` while RSS-minus-shared alone is still below it → `TryAcquire` admits another compaction, its read/merge buffers land on top of cgroup usage already at the watermark, and the pod can be OOM-killed even though the gate never refused anything. **Suggestion:** a matching-scope reader already exists in-tree as `getContainerMemUsed()` (`common/hardware/container_linux.go:95-112`, usage minus `inactive_file`, the same formula `docker stats` uses). Add an exported wrapper in package `hardware` with a fallback to the RSS reading when the cgroup read fails, and use it for the watermark comparison. (raised by sijie-ni-0214, xiaocai2333)

czs007:

**Verified:**
- `server/processor/segment_processor.go:674-699`: the per-segment compare-and-swap (CAS) is taken before `TryAcquire`. A refused request returns right away, and the already-registered `defer Store(false)` frees the flag. `Release` is deferred immediately after a successful acquire, so every return path runs it.
- `server/processor/memory_admission.go:577-604`: the check and the reservation happen under one `a.mu` lock, and a refused request reserves no bytes. The `-race` concurrent test at `memory_admission_test.go:871` asserts `admitted > 0`, so it cannot pass without admitting anything.
- `server/processor/memory_admission.go:626-638`: `Release` does nothing when `tasks == 0`, and `reserved` goes back to zero when the node is idle. An unmatched or wrongly sized release therefore cannot raise the ceiling or drift the count.
- `server/storage/stagedstorage/writer_impl.go:1679`: the staged estimate calls the same `planMergeBlockTasks(target, expectedLastEntryId)` as `readLocalFileAndUploadToMinio` (`:1615`), so the estimate and the real merge use the same plan.
- `common/config/configuration.go:1002-1009`: older YAML files without the new keys get defaults that pass the new validation. The Milvus-side builder starts from `NewConfiguration()` and overrides only individual fields, so an upgrade does not fail validation at startup.
- `werr/errors.go:258`: error code 2112 is unique and retryable. `classifyCompactionFailureReason` sends it to the transient branch, so capacity refusals do not count as compaction failures.

## Low (non-blocking)

- **`cmd/external/user_config.go:355`**: nothing checks that `memoryHighWatermark` from the external user config is in range. The `(0,1]` check (`common/config/configuration.go:863`) runs only inside `NewConfiguration` (`cmd/main.go:105`). `ApplyToConfig` merges external values afterwards (`cmd/main.go:115`) and only requires `> 0`, and nothing validates the config after the merge.
  - **Failure:** setting `woodpecker.logstore.segmentCompactionPolicy.memoryHighWatermark: 70` in `user.yaml` (meant as 70%) → startup succeeds, the pressure threshold becomes 70× the container memory limit, and the pressure gate never refuses a compaction. Only `maxInflightMemory` still limits memory.
  - **Suggestion:** reject or clamp values outside `(0,1]` during the merge, or call `cfg.Validate()` again after `ApplyToConfig`.
  - **Earlier rounds:** raised in the previous round. The author has not responded, so it is still open.
  - (surfaced during verification)

- **`server/storage/stagedstorage/writer_impl.go:1655`**: on the error path, the memory reservation is released while merge/upload workers are still running. The future-collection loop returns on `ctx.Err()` or on the first failed future (`:1655-1661`). `defer uploadPool.Release()` (`:1632`) only calls ants' `inner.Release()` (`common/conc/pool.go:115-116`), which does not wait for workers that are already running. `server/storage/objectstorage/writer_impl.go:1844` / `:1885-1887` has the same structure. Before this PR, returning early did not affect any accounting. With this PR, `Release(compactionEstimate)` (`segment_processor.go:699`) returns budget that those workers still occupy. The overshoot is bounded: `defer cancel()` (`segment_processor.go:660`) runs right after `Release`, leftover workers exit at their next context check, and the excess cannot exceed this task's own estimate.
  - **Failure:** with `maxParallelUploads=4`, one merge task's upload fails while the other 3 workers are still reading local blocks or uploading → the failed `Compact` returns the full estimate at once, a concurrent compaction for another log is admitted against that budget, and actual compaction memory briefly goes above `maxInflightMemory`.
  - **Suggestion:** on the error path, cancel the context and wait for the in-flight futures before returning, in both writers. Alternatively, release the reservation only after all pool workers have exited.
  - **Earlier rounds:** raised in the previous round. The author has not responded, so it is still open. It is rated low here because of the bound described above.
  - (surfaced during verification)

- **`server/processor/memory_admission.go:166`**: the pressure gate compares this process's RSS (Resident Set Size) minus shared memory (`hardware.GetUsedMemoryCount()`, `common/hardware/mem_info.go:44-51`) with the cgroup memory limit (`common/hardware/hardware_info.go:81-84`). These two numbers measure different things:
  - The process-level reading leaves out other processes in the same cgroup, shared anonymous memory (shm/tmpfs) and kernel memory.
  - The new `config/woodpecker.yaml` comment says the gate measures the node's memory usage "whatever is using it".
  - The error only makes the gate fire later than it should. It never refuses work wrongly.
  - If the reading fails, `GetUsedMemoryCount` returns 0 and the gate lets everything through.
  
  Switching to `getContainerMemUsed` (`common/hardware/container_linux.go:95-111`, `usage - inactive_file`) has its own downside. It counts active file cache as used, and staged-mode `data.log` writes and compaction read-back both create that cache. A write-heavy node could then have compactions refused because of memory the kernel can reclaim.
  - **Suggestion:** at minimum, change the config comment to say what is actually measured (process RSS minus shared). If you want node-wide coverage, use a cgroup-level reading that excludes reclaimable file cache. Also consider logging when the memory reading fails instead of silently letting everything through.
  - **Earlier rounds:** raised in the previous round. The author has not responded, so it is still open.
  - (surfaced during verification)

PR #334, opened 2026-09-16
fix(operator): make the init container image configurable

mutoulion:
## What problem does this PR solve?

  The operator hardcoded the init-topology init container image to
  `curlimages/curl:8.7.1`, which cannot be pulled in restricted network
  environments.

  ## What is changed and how it works?

  Add an optional `initImage` field to `WoodpeckerClusterSpec`, following
  the same pattern as the existing `image` field (kubebuilder default
  `curlimages/curl:8.7.1`, so behavior is unchanged by default). The
  `buildInitContainers()` now uses `cluster.Spec.InitImage`.

  ## Checklist

  - [x] Code complies with the existing style
  - [x] Unit tests added/updated (`make test` passes)
  - [x] CRD manifests regenerated (`make manifests generate`)


mutoulion:
@tinswzy Hi, could you take a look? Thanks!

tinswzy:

This PR makes the init-topology container image user-configurable via `spec.initImage`, and the findings below all concern paths that this new user-supplied value opens up.

## Medium

**`deployments/operator/internal/controller/reconcile_statefulset.go:249` — `InitImage` has no consumer-side fallback and no schema constraint.**
The init container is now built with `Image: cluster.Spec.InitImage` and nothing guards an empty value: `types.go:34-37` carries only `+kubebuilder:default` (no `MinLength`), and `validateWoodpeckerCluster` (`internal/webhook/v1alpha1/woodpeckercluster_webhook.go:89-111`) checks only replicas/servicePort/gossipPort/metricsPort. A CRD structural default fills the field only when it is absent or null. Failure: a user explicitly sets `spec.initImage: ""`, or the new operator image rolls out ahead of the updated CRD so the apiserver prunes the unknown field → `buildInitContainers` emits an init container with `Image: ""`, the apiserver rejects the StatefulSet with `spec.template.spec.initContainers[0].image: Required value`, and reconcile retries in a loop with no pods ever starting.
Suggestion: do both — add a package-level `defaultInitImage` constant fallback inside `buildInitContainers` (`img := cluster.Spec.InitImage; if img == "" { img = defaultInitImage }`), and add `+kubebuilder:validation:MinLength=1` at `woodpeckercluster_types.go:35` so an explicit empty string is rejected at admission. Note that adding a branch to the webhook `Default()` would not help here: that code path does not run in any deployment configuration (see the pre-existing section below). *(raised by xiaocai2333, czs007)*

**`deployments/operator/internal/controller/reconcile_statefulset.go:220` — the init script hard-depends on `curl`, and a missing `curl` exits 0 with empty REGION/AZ.**
`NODE_JSON=$(curl -sf --max-time 10 ... || true)` implements deliberate fail-soft behaviour for a failed API query, but `|| true` swallows `sh: curl: not found` (exit 127) just as readily, and a failure inside `$(...)` is not caught by the leading `set -e`. Before this PR the image was pinned to `curlimages/curl:8.7.1`, so the code itself guaranteed `curl` was present; that guarantee is now the user's responsibility, and neither the script nor the field comment states the contract. Failure: a restricted-network user sets `spec.initImage: internal-registry/alpine:3.20` (or busybox, neither of which ships `curl`) → the init container prints "Init complete" and exits 0, the pod goes Ready, but `topology.env` has empty `REGION`/`AVAILABILITY_ZONE` → `topology.Scope()` returns `ScopeUnknown` for every peer, AZ-local replica selection in `orderedQuorumReadCandidates` degrades to no locality, read traffic crosses AZs at random, and every traffic metric collapses to `scope="unknown"` — with no error anywhere.
Suggestion: separate "unusable image" from "API query failed" by probing before the query, e.g. `command -v curl >/dev/null 2>&1 || { echo 'initImage must provide curl' >&2; exit 1; }`; and document the contract (the image must provide `/bin/sh` and `curl`) in the field comment at `woodpeckercluster_types.go:34`, since that comment becomes the CRD description and is the only place users will see it. *(raised by czs007)*

**`deployments/operator/internal/controller/reconcile_statefulset.go:246` — the init container sets no `ImagePullPolicy`, unlike the main container.**
The init container literal (lines 246-263) sets only Name/Image/Command/Env/VolumeMounts, while the main container uses `ImagePullPolicy: cluster.Spec.ImagePullPolicy` (lines 118-121). Previously the pinned `curlimages/curl:8.7.1` tag made Kubernetes infer `IfNotPresent`, which happened to match the CR default; with a user-supplied image the inference can now differ. Failure: an air-gapped cluster pre-loads the init image on every node and sets `spec.imagePullPolicy: Never` with `spec.initImage: internal-registry/curl-tools:latest` → the main container starts from the local image as instructed, but the init container, having no explicit policy and a `:latest` tag, is implicitly set to `Always`, the pull fails with no registry reachable, and the pod is stuck in `Init:ImagePullBackOff` indefinitely.
Suggestion: add `ImagePullPolicy: cluster.Spec.ImagePullPolicy` to the init container literal so both containers honour the same CR field. *(raised by xiaocai2333, czs007)*

## Low

**`deployments/operator/internal/controller/reconcile_statefulset_topology_test.go:101` — the only risky path, "`initImage` unset", is untested.**
The existing case now sets `InitImage: "curlimages/curl:8.7.1"` itself and then asserts `Contains(initCs[0].Image, "curlimages/curl")`, i.e. it asserts the value it just wrote; the new `TestBuildInitContainers_UsesCustomInitImage` (lines 111-125) likewise covers only an explicitly assigned value. The CR-omits-`initImage` path that relies on the CRD default — the entry point for the empty-image issue above — has zero coverage. Note also that `woodpeckercluster_webhook_test.go:51-63` enumerates seven default-value assertions and includes no `InitImage`, so a green CI does not mean this chain was exercised.
Suggestion: `suite_test.go:69-72` already starts a real apiserver with `CRDDirectoryPaths` pointed at `config/crd/bases`, so CRD defaulting genuinely applies there — add a case that creates a CR without `initImage`, reconciles, and asserts `InitContainers[0].Image != ""`. No new test infrastructure is needed. *(raised by xiaocai2333, czs007)*

- `deployments/operator/internal/controller/reconcile_statefulset.go:102` — `spec.imagePullSecrets` is declared in the API type but never propagated to the PodSpec: `buildPodSpec` (lines 102-113) sets ServiceAccountName/InitContainers/Containers/Volumes/Affinity/Tolerations/NodeSelector/TopologySpreadConstraints and nothing else, and a repo-wide search finds `ImagePullSecrets` only in the types file, generated deepcopy, CRD YAML, and docs — zero hits under `internal/controller`. Failure: an operator puts the main image or `initImage` on a credentialed private registry and sets `spec.imagePullSecrets: [{name: my-regcred}]` per the CRD docs → the generated PodSpec carries no pull secrets, kubelet's anonymous pull is refused, both containers land in `ImagePullBackOff`, and the pod never becomes Ready while the CR field appears accepted. Unrelated to this PR because `buildPodSpec` is untouched by the 27-line diff and the same omission exists at base `dc9666e`, where the field was already declared. Since it affects the same private-registry scenario this PR targets, we recommend merging that separate fix before this PR. *(raised by xiaocai2333, czs007)*
- `deployments/operator/config/manager/manager.yaml:70` — the defaulting/validating webhook is disabled on every deployment path, making `Default()` effectively dead code: `manager.yaml:69-71` sets `ENABLE_WEBHOOKS: "false"`, `cmd/main.go:122` derives `enableWebhooks` from it and skips both the webhook server and `SetupWoodpeckerClusterWebhookWithManager` (lines 124-126, 198-203), and `config/default/kustomization.yaml:21-22` leaves `- ../webhook` commented out, so no Mutating/Validating webhook configuration exists in the cluster. `woodpeckercluster_webhook.go:52-80` and `89-111` therefore execute only under unit tests. Unrelated to this PR because `manager.yaml`, `cmd/main.go`, `config/default/kustomization.yaml`, and `internal/webhook/**` are untouched by the diff and the webhook has been disabled since base `dc9666e`. It is listed here because it determines how the first finding above must be fixed — the "CRD default plus webhook safety net" assumption does not hold, so a fix must go through CRD schema validation and a consumer-side fallback. *(raised by czs007)*

PR #336, opened 2026-09-17
enhance(operator): fall back to a known init image, and document initImage

### LINKED ISSUE #335: enhance(operator): harden the new spec.initImage — controller-side fallback + docs (follow-up to #334)
## Summary

#334 makes the `init-topology` init container image configurable via a new `spec.initImage` field, defaulted through a `+kubebuilder:default` marker. That PR is correct and should merge as-is — this issue tracks two hardening follow-ups that were deliberately left out to keep it at 27 lines.

Both are follow-ups, not defects in #334.

## 1. Add a Go-layer fallback for `spec.initImage`

`buildInitContainers()` reads `cluster.Spec.InitImage` with no fallback (`deployments/operator/internal/controller/reconcile_statefulset.go:249`). The only thing that fills it is the CRD's structural-schema default.

**Why this new field needs a fallback when `spec.image` arguably does not.** The two are not equally exposed:

- **Old CRs are already fine.** CRD defaulting applies when reading from etcd, not just on write, so a `WoodpeckerCluster` created before this field existed comes back from the API server with `initImage: curlimages/curl:8.7.1` already filled in. Nobody has to re-apply anything.
- **A stale CRD is the real window.** If the operator image is upgraded without re-applying the CRD, the installed schema has no `initImage` property at all — no default, and a user-supplied value gets pruned. `spec.image` is immune to this because its default has shipped since day one; `spec.initImage` is not, precisely because it is new.

**Failure mode.** `InitImage == ""` produces a StatefulSet with an empty `spec.template.spec.initContainers[0].image`, which the API server rejects (`Required value`). Running pods are untouched, so this is not an outage — the reconcile loop just wedges, and subsequent config changes or scaling stop taking effect.

**Note on the webhook.** `WoodpeckerClusterCustomDefaulter.Default()` (`deployments/operator/internal/webhook/v1alpha1/woodpeckercluster_webhook.go`) mirrors most of the marker defaults, but adding an `InitImage` branch there would not help: `config/default/kustomization.yaml` has `- ../webhook` commented out (it requires cert-manager), so the mutating webhook is not deployed in a standard install. The CRD marker is the only defaulting path that actually runs. The fix belongs in the controller.

**Proposal.**

```go
const defaultInitImage = "curlimages/curl:8.7.1"

// in buildInitContainers()
img := cluster.Spec.InitImage
if img == "" {
    img = defaultInitImage
}
```

One constraint worth respecting: `+kubebuilder:default=` only accepts a literal, so the marker and the constant cannot share a source. Comment them as a pair, so that bumping the curl version does not silently leave two different defaults behind — that would be worse than today's single hardcoded value.

## 2. Document `spec.initImage`

The field is invisible to the people it was written for. Restricted-network users are exactly the ones who read the getting-started guide and copy the spec block out of it, and neither the guide nor the samples mention it:

- `docs/admin-guides/operator/operator-guide-getting-start.md:148` — the example spec lists `image` / `imagePullPolicy`, not `initImage`
- `deployments/operator/config/samples/*.yaml` — same (5 files)

Worth a sentence explaining *why* it exists: the init container queries the Kubernetes API for the node's `topology.kubernetes.io/zone` and `region` labels, which is why it needs a curl-capable image and why an air-gapped registry needs to mirror one.

## Scope

Small, self-contained, no behavior change in the default path. Should land after #334 merges.

tinswzy:
## What problem does this PR solve?

Fixes #335. Follow-up hardening for #334, which added `spec.initImage`.

#334 is correct — this is not a fix for a defect in it. It carries the default in a `+kubebuilder:default` marker, which lives in the CRD schema. That leaves one gap:

**An operator upgraded against a CRD that predates the field reads an empty string.** The schema has no `initImage` property, so nothing fills it and a user-supplied value gets pruned. The resulting StatefulSet has an empty `spec.template.spec.initContainers[0].image`, which the API server rejects (`Required value`). Existing pods keep running, so this is not an outage — the reconcile loop just wedges, and subsequent config changes or scaling stop taking effect.

Old CRs are *not* part of this problem: CRD defaulting applies when reading from etcd, so a `WoodpeckerCluster` created before the field existed already comes back with `initImage` filled in. Only a stale CRD opens the window.

`spec.image` is immune to this because its default has shipped since day one. `spec.initImage` is not, precisely because it is new.

The field was also undiscoverable by the people it was written for. Restricted-network users are the ones who need it, and it appeared in neither the getting-started guide nor the samples.

## What is changed and how it works?

**1. Controller-side fallback.** `buildInitContainers()` falls back to a `defaultInitImage` constant when `spec.initImage` is empty, instead of trusting the schema alone.

The defaulting webhook is deliberately not the place for this: `config/default/kustomization.yaml` has `- ../webhook` commented out (it requires cert-manager), so `WoodpeckerClusterCustomDefaulter.Default()` does not run in a standard install. Adding a branch there would look like a fix without being one.

One caveat is called out in a comment at the constant: `+kubebuilder:default=` only accepts a literal, so the marker and the constant cannot share a source. They have to be bumped together, or two different defaults drift apart silently — which would be worse than the single hardcoded value this started as.

**2. Documentation.**

- A troubleshooting entry in the getting-started guide, explaining why an init container exists at all (it reads the node's `topology.kubernetes.io/zone` / `region` labels via the Kubernetes API, hence the need for a curl-capable image) and how to point it at a mirrored registry. It also notes that `imagePullSecrets` is pod-level and so already covers the init container.
- A row in the spec reference table in `docs/wip/operator/woodpecker_operator.md`.
- A commented line in the canonical sample.

The four specialized samples (external SA, three cloud IAM) are left alone — they demonstrate SA and cloud IAM wiring, not the field surface, and the same commented block repeated five times is noise. #335 suggested all five; this is a deliberate narrowing.

## Checklist

- [x] Code complies with the existing style
- [x] Unit test added for the empty-`initImage` path (`TestBuildInitContainers_FallsBackWhenInitImageUnset`)
- [x] `make test` passes — `internal/controller` and `internal/webhook/v1alpha1` both ok under envtest
- [x] `make manifests generate` produces no drift (no API type changes)
- [x] Default behavior unchanged; no CRD change, so no upgrade ordering requirement of its own


czs007:

Adversarial review found one documentation-accuracy issue: a newly added troubleshooting note points users at a CR field that the operator never wires into the pod spec.

## Medium

**`docs/admin-guides/operator/operator-guide-getting-start.md:269`** — The new troubleshooting section tells users that for a private registry they should add the pull secret to `spec.imagePullSecrets` because "it is a pod-level field, so it covers the init container too." The Kubernetes semantics are correct, but the premise is not: the CR field is inert in this repository and is never read by the operator. `buildPodSpec()` (`deployments/operator/internal/controller/reconcile_statefulset.go:108-120`) constructs a `corev1.PodSpec` with only `ServiceAccountName`, `TerminationGracePeriodSeconds`, `InitContainers`, `Containers`, `Volumes`, `Affinity`, `Tolerations`, `NodeSelector`, and `TopologySpreadConstraints` — no `ImagePullSecrets`. The ServiceAccount inheritance path is closed as well: `reconcileServiceAccount()` (`reconcile_serviceaccount.go:31-49`) sets only labels and an owner reference, and the external-SA branch returns early. A repository-wide search for `ImagePullSecrets` hits only the type definition (`api/v1alpha1/woodpeckercluster_types.go:44-46`), the CRD schema, `zz_generated.deepcopy.go`, and docs — zero read sites.

Failure scenario: a restricted-network user mirrors the curl and woodpecker images to an auth-required registry and sets `spec.initImage: my-registry.internal/curl:8.7.1`, `spec.image: my-registry.internal/woodpecker:v0.1.26`, and `spec.imagePullSecrets: [{name: regcred}]` exactly as this section instructs → the rendered StatefulSet pod template and the operator-created ServiceAccount both carry no `imagePullSecrets`, the kubelet pulls anonymously and gets a 401, and `init-topology` stays in `ImagePullBackOff` — the precise symptom this section's heading promises to resolve. Operators following the doc will misdiagnose it as a misconfigured registry or secret.

Note the blast radius is wider than the init container: the section's YAML example also points the main container at `my-registry.internal/woodpecker:v0.1.26` (`:264`), which is equally unable to obtain the pull secret.

Suggestion — either of:
- Wire the field: add `ImagePullSecrets: cluster.Spec.ImagePullSecrets` to `buildPodSpec()`. The field, its deepcopy, and the CRD schema all already exist, so this is a one-line change that makes the note true.
- Correct the note: drop the `spec.imagePullSecrets` sentence and instead direct users to attach the pull secret to the pod's ServiceAccount (`spec.serviceAccountName` already supports an external SA), which is the mechanism that actually applies today.

The unwired API field predates this PR; what is new here is the documentation claim that setting it resolves the `ImagePullBackOff`. *(raised by bigsheeper, xiaocai2333)*

PR #342, opened 2026-09-17
enhance(metrics): export the configured watermark thresholds

### LINKED ISSUE #338: enhance(metrics): watermark thresholds are config-only, so alert rules must hardcode them
## Summary

The thresholds that govern backpressure live only in configuration. Metrics expose the **input** and the **effect**, but not the **line**:

| dimension | input | effect | threshold |
| --- | --- | --- | --- |
| disk | `logstore_disk_usage_ratio` | `logstore_write_backpressure_state`, `logstore_write_reject_probability`, `logstore_write_rejected_total{reason="disk_pressure"}` | **not exposed** (soft 0.80 / hard 0.90 in config) |
| compaction memory | `system_memory_usage_ratio`, `compaction_admission_reserved_bytes` | `compaction_admission_rejected_total{reason="memory_pressure"}` | **ratio not exposed** — the absolute cap `compaction_admission_max_inflight_bytes` (`service_metrics.go:344`) already is |

## What this costs today

Every alert rule that references a watermark has to hardcode the number. `disk_usage_ratio > 0.75` is really "5 points below the soft watermark", but that intent is not in the expression. Change the config and the alert silently keeps thresholding against the old value — a failure mode that produces no error anywhere.

We are rolling this out across 34 regions. Hardcoded thresholds in rule files that must stay in sync with per-region config is exactly the kind of drift that goes unnoticed until it matters.

## Proposed metrics

```
woodpecker_server_logstore_disk_watermark_soft{node_id}
woodpecker_server_logstore_disk_watermark_hard{node_id}
woodpecker_server_compaction_admission_memory_watermark{node_id}
```

Gauges, set once at startup from the resolved config, alongside the existing `compaction_admission_max_inflight_bytes` which already follows exactly this pattern.

Alert rules then read as what they mean:

```promql
woodpecker_server_logstore_disk_usage_ratio
  > on(node_id) woodpecker_server_logstore_disk_watermark_soft - 0.05
```

### Deliberately not proposed: a memory-side `backpressure_state`

It would be the obvious symmetric addition and it is the wrong thing. The disk state machine rejects **writes**; memory pressure only rejects **compactions**. A gauge named to mirror `logstore_write_backpressure_state` but meaning "compaction is being declined" would read as "writes are being declined" to anyone scanning a dashboard. The effect is already observable through `compaction_admission_rejected_total{reason="memory_pressure"}`; what is missing is only the threshold, which is what this issue asks for.

The server dashboards have a **Disk Watermark — Backpressure** row and a **Compaction Admission** row. Add the threshold as a constant line on the existing ratio panels in each, so the distance to the line is visible rather than remembered.

tinswzy:
Closes #338.

## What

Three gauges for the thresholds that govern backpressure, which until now lived only in configuration:

```
woodpecker_server_logstore_disk_watermark_soft{node_id}
woodpecker_server_logstore_disk_watermark_hard{node_id}
woodpecker_server_compaction_admission_memory_watermark{node_id}
```

Alert rules can then say what they mean:

```promql
woodpecker_server_logstore_disk_usage_ratio
  > on(node_id) woodpecker_server_logstore_disk_watermark_soft - 0.05
```

instead of `> 0.75`, which is really "5 points below the soft watermark" with that intent nowhere in the expression — and which keeps thresholding against the old value after a config change, silently.

## Effective, not configured

All three report the threshold the node **actually enforces**:

* The disk pair is gated on `dwActive` — the same condition that decides whether `newDiskWatermarkTask` is registered — not merely on `dwPolicy.Enabled`. A node that keeps no WAL on local disk never applies the watermark, so publishing its configured ratio would hand alert rules a line that can never be crossed. It publishes `0` instead.
* `MemoryAdmission.MemoryHighWatermark()` returns `0` when the pressure gate is inactive (no readable container memory limit, or no watermark configured), mirroring the existing `MaxInflightBytes()` shape including nil-safety.

So a non-zero value is always a line that can actually be crossed, which is the property a rule needs.

## Deliberately not included

A memory-side `backpressure_state` mirroring the disk one. It is the obvious symmetric addition and it is the wrong thing: the disk state machine rejects **writes**, memory pressure only rejects **compactions**. A gauge named to mirror `logstore_write_backpressure_state` but meaning "compaction is being declined" would read as "writes are being declined" to anyone scanning a dashboard. The effect is already observable through `compaction_admission_rejected_total{reason="memory_pressure"}`; only the threshold was missing.

## Dashboards

Soft and hard ratios added as series on **WAL Disk Usage Ratio**; the compaction watermark on **Memory Usage** (which already carries a ratio series). Both the k8s and docker server dashboards, with `manifests/dashboard-configmaps.yaml` regenerated via `-update`.

## Test

`TestMemoryAdmission_MemoryHighWatermark` covers the four cases that matter for the gauge: gate live, no readable limit, no watermark configured, and nil receiver.

```
go test ./server/processor/ -run TestMemoryAdmission   ok
go test ./tests/k8s/monitor                            ok
go vet ./common/metrics/ ./server/ ./server/processor/ clean
```


czs007:

**Verified:**

- `server/processor/memory_admission.go:225` — `MemoryHighWatermark()` routes through `pressureGateEnabled()`, whose `a != nil && ...` check short-circuits on a nil receiver and returns 0 instead of panicking.
- `server/processor/memory_admission.go:163` — the getter's gating condition and value are identical to what `TryAcquire` actually enforces, so the exported watermark can never be a line that never triggers.
- `server/logstore.go:164` — the `dwActive` boolean used to publish the disk gauges is the same expression used to register `newDiskWatermarkTask`, and the inactive branch explicitly sets 0, so there is no registered-but-zero or unregistered-but-nonzero mismatch and no stale series.
- `server/maintenance.go:360` — the task copies `DiskWatermarkPolicy` verbatim with no normalization, and `diskRejectBpsFor` compares against the same `SoftThresholdRatio`/`HardThresholdRatio`, so the gauges export the thresholds in force.
- `common/metrics/service_metrics.go:459` — all three new gauges are registered exactly once inside the existing `Once` block under names that do not collide with existing metrics, so `MustRegister` cannot panic on a duplicate.
- `server/processor/memory_admission_test.go:125` — the new test covers all four getter branches (gate live, no trusted memory limit, watermark unconfigured, nil receiver), matching the `pressureGateEnabled` conditions one-for-one.

PR #343, opened 2026-09-17
enhance(metrics): count nodes returned by selection, so skew is localizable

### LINKED ISSUE #339: enhance(metrics): quorum_selection_skew_total has no node_id, so skew cannot be localized
## Summary

`woodpecker_server_quorum_selection_skew_total` (`common/metrics/service_metrics.go:382`) carries **`mode` as its only label**. It answers "did selection run weighted or uniform-random", which is useful, but it cannot answer the question its name suggests: **which node is being over-selected**.

Skew is inherently a per-node quantity. With only `mode`, the metric is blind to it.

## What this costs today

Our "quorum selection skew" alert has to approximate, using the spread of `logstore_active_segments` across nodes:

```promql
max(sum by (node_id) (woodpecker_server_logstore_active_segments))
/ avg(sum by (node_id) (woodpecker_server_logstore_active_segments)) > 3
```

That conflates two different things — genuinely uneven selection, and even selection across logs of very different sizes. It fires on the second and misses mild cases of the first.

## Why "just add node_id" is the wrong fix

The obvious change — `[]string{"mode"}` → `[]string{"node_id", "mode"}` — breaks the metric's meaning, and the call sites show why (`common/membership/service_discovery.go:1161-1198`):

```go
if !anyKnown {
    metrics.WpQuorumSelectionSkew.WithLabelValues("random_no_load").Inc()
    return sd.randomSelectNodes(nodes, limit)
}
metrics.WpQuorumSelectionSkew.WithLabelValues("weighted").Inc()

picked := weightedSampleIndices(weights, limit, sd.randFloat)
```

It is incremented **once per selection call**, and in the weighted branch the increment happens **before** `weightedSampleIndices` runs — the chosen nodes are not known at that line. Adding `node_id` would silently change the unit from *selections* to *node-selections* (a factor of `limit`, normally 3), invalidating any existing comparison between the two mode series.

## Proposal

Leave `quorum_selection_skew_total` alone and add a second counter recording the actual outcome:

```
woodpecker_server_quorum_node_selected_total{node_id, mode}
```

Incremented once per node in the returned set, after selection — i.e. over `out` in `selectLowestLoadNodes`, and correspondingly in `randomSelectNodes`. Roughly four lines. Skew then becomes a direct query with no proxy:

```promql
max(rate(woodpecker_server_quorum_node_selected_total[15m]))
/ avg(rate(woodpecker_server_quorum_node_selected_total[15m]))
```

### Open question for review

`ServiceDiscovery` lives in `common/membership` but emits metrics in the `server` subsystem. Worth confirming which process actually runs the selection path in the storage-service deployment, since that determines which dashboard the panel belongs on — and, if it is the client process, whether the `server` subsystem prefix is right for these two metrics at all.

tinswzy:
Closes #339.

## What

```
woodpecker_server_quorum_node_selected_total{node_id, selected_node_id, mode}
```

incremented once per node in the set `selectLowestLoadNodes` returns. Skew becomes a direct query:

```promql
  max(sum by (selected_node_id) (rate(woodpecker_server_quorum_node_selected_total[15m])))
/ avg(sum by (selected_node_id) (rate(woodpecker_server_quorum_node_selected_total[15m])))
```

instead of the proxy we use today — the spread of `logstore_active_segments` across nodes — which conflates genuine skew with logs of legitimately uneven size.

## Why not just add `node_id` to the existing counter

`quorum_selection_skew_total` is incremented **once per selection call**, and on the weighted path **before** `weightedSampleIndices` runs (`service_discovery.go:1188-1193`) — the chosen nodes are not known at that line. Relabelling it would also silently change its unit from *selections* to *node-selections*, a factor of `limit` (normally 3), invalidating any existing comparison between the two mode series. It is left untouched.

## Label choice

`node_id` keeps the family convention — the node emitting the sample, i.e. the one that ran the selection, since a `SelectNodes` RPC is served by whichever logstore the client reached. `selected_node_id` is the outcome. Two identities that are genuinely different, so they get two labels rather than one overloaded `node_id` that would mean "this node" on every other server metric and "some other node" on this one.

Cardinality is bounded by (selectors × selected × 3 modes); at our largest planned cluster, 12 × 12 × 3.

## A blind spot worth knowing about

The mode set gains `load_disabled` for the `sd.loadAware == false` path. The existing counter does not record that path at all — with `loadAwareEnabled: false`, `quorum_selection_skew_total` is simply **silent**, which is easy to mistake for "no selections are happening". Not changed here (that would alter the existing counter's meaning), but the new counter does cover it.

## Scope

Go only. There is no skew panel in either server dashboard today, so adding one means allocating a new panel id — that belongs with the alert-aligned panel work rather than colliding with it.

## Test

`TestQuorumNodeSelected_CountsEveryReturnedNode` covers all three modes plus a nil/meta-less candidate. Assertions are on deltas, since the counter is a package-level global shared across tests.

```
go test ./common/membership/   ok  (36s, full package)
go vet ./common/membership/ ./common/metrics/   clean
```


czs007:

Adversarial review of the new `WpQuorumNodeSelected` counter found one metric-correctness defect in where the instrumentation is placed, plus two accuracy gaps in the documented skew query and one doc-comment placement issue.

## Medium

**`common/membership/service_discovery.go:1198` — per-node counting lives in `selectLowestLoadNodes`, a primitive that exhaustive search reuses as a per-combination selector.**
The three new `recordSelection` call sites (`:1167` `load_disabled`, `:1189` `random_no_load`, `:1198` `weighted`) sit at the lowest level, but `selectLowestLoadNode(nodes)` wraps `selectLowestLoadNodes(nodes, 1)` (`:1144-1150`) and is used once per AZ/RG combination by all three exhaustive-search paths (`:484`/`:500`, `:569`/`:585`, `:657`/`:673`). One call therefore increments `len(allValidNodes) + limit` times while returning only `limit` nodes. *Failure: with `single_az_multi_rg` (or the other exhaustive strategies), all 5 random attempts missing, 6 valid AZ-RG combinations and `filter.Limit = 3` → one `SelectNodes` call increments `woodpecker_server_quorum_node_selected_total` 9 times: 3 intermediate candidates that never appear in the returned set are counted as selected, and the 3 actually returned nodes are each counted twice, so the `max(...) / avg(...)` panel shows a skew that does not exist over 3x the real sample volume.* This contradicts both the Help text and the PR description ("once per node in the returned set"), and it distorts exactly the non-random strategies the metric exists to measure. Note the counting is correct when `limit == 0 || limit >= len(allValidNodes)` (`:496`/`:581`/`:669` return early), and phase-1 paths (`:457`/`:542`/`:624`) are unaffected — `numAZs = min(limit, len(candidateAZs))` (`:609`) makes the second convergence at `:631-632` unreachable. Suggestion: record once at the `Select*` (RPC/selection) boundary, or add a non-recording variant (e.g. `selectLowestLoadNodesNoRecord`) for `selectLowestLoadNode` to call. *(raised by bigsheeper, xiaocai2333)*

## Low

**`common/membership/service_discovery.go:1209` — the documented `max/avg by (selected_node_id)` ratio is dominated by topology imbalance under per-group strategies.**
The comment at `:1206-1211` and the Help string at `common/metrics/service_metrics.go:390` present that ratio as a directly readable skew signal ("with no proxy through active_segments, which conflates skew with logs of uneven size"). But most non-random strategies pick one node per group rather than sampling uniformly across candidates: `SelectMultiAzMultiRg` one per selected AZ (`:624`), `SelectSingleAzMultiRg` one per selected RG (`:457`), `SelectMultiAzSingleRg` one per selected AZ (`:542`). When group sizes differ, nodes in small groups are selected far more often (≈1/2 in a 2-node AZ vs ≈1/10 in a 10-node AZ), so the ratio reports topology imbalance as scheduling skew. This is a documentation-accuracy issue, not a runtime defect. Suggestion: state in the comment and Help that the ratio must be normalized by group size, or scope it to strategies that pick `limit` nodes from one pool (`random`, `single_az_single_rg`). *(raised by bigsheeper, xiaocai2333)*

**`common/membership/service_discovery.go:735` — the new counter is permanently empty under the `random-group` strategy, producing a silent metric.**
Every reachable return path in `SelectRandomGroup` (`:710-777`) bypasses `selectLowestLoadNodes`: `:735-737` returns directly when `limit == 0 || limit >= len(allCandidates)`; otherwise `numGroups = len/limit >= 1` (`:745`) and `groupIdx ∈ [0, numGroups-1]` (`:751`) make `end <= len(allCandidates)` (`:755`) always true, so `:757` returns and the only `selectLowestLoadNodes` call site — the partial-group fill branch at `:761-774` — is unreachable. The strategy is reachable in production: `woodpecker/quorum/discovery.go:107-108` maps `"random-group"` to `StrategyType_RANDOM_GROUP` and `server/service.go:752` dispatches to `SelectRandomGroup`. *Failure: with selection strategy `"random-group"` configured and `SelectNodes` returning `limit` nodes normally → `woodpecker_server_quorum_node_selected_total` emits no series at all on that node, and the skew dashboard reads `no data`, indistinguishable from "no selection ever happened" — the same misreading this PR cites as motivation for instrumenting `load_disabled`.* Suggestion: note the coverage gap in the `recordSelection` comment and Help, or move instrumentation up to the `Select*` boundary, which also resolves the medium item above. *(raised by bigsheeper)*

## Nitpick

**`common/metrics/service_metrics.go:378` — the new metric is inserted between `WpQuorumSelectionSkew`'s doc comment and its declaration.**
Lines `:378-381` are the existing doc comment for `WpQuorumSelectionSkew`. The new `WpQuorumNodeSelected` comment and declaration (`:382-391`) now sit between that comment and the `WpQuorumSelectionSkew` declaration at `:392`, so godoc/IDEs attach the old skew comment to `WpQuorumNodeSelected` (no blank line before `:386`), `WpQuorumSelectionSkew` is left without a doc comment, and the two comment blocks each refer to "the metric below". Suggestion: move the new metric together with its own comment below `:397`, or add a blank line after `:381` and reorder. *(surfaced during verification)*

PR #344, opened 2026-09-18
enhance(metrics): report device IOPS and throughput for the WAL volume

### LINKED ISSUE #337: enhance(metrics): no device-level disk IOPS or throughput, so capacity alerts cannot be written
## Summary

`system_metrics_collector.go` exposes CPU, memory and disk **capacity**, but nothing about disk **rate**:

| exists | what it gives |
| --- | --- |
| `woodpecker_server_system_cpu_usage{node_id}` | CPU % |
| `woodpecker_server_system_memory_usage_ratio{node_id}` | memory ratio |
| `woodpecker_server_system_disk_total_bytes` / `_used_bytes{node_id,path}` | capacity |
| `woodpecker_server_logstore_disk_usage_ratio` / `_disk_free_bytes{node_id,path}` | watermark |
| `woodpecker_server_system_io_wait{node_id}` | saturation proxy |

There is no ops/sec and no bytes/sec for the local WAL device. A grep over the non-test tree for `iops|disk_io|IOCounters|ReadCount|WriteCount|read_bytes|write_bytes|throughput` returns only comments in `common/config/configuration.go:236` and prose in `cmd/wpcli`.

## What this costs today

We are standing up capacity/scale-up alerting for a Pulsar → Woodpecker migration across 34 regions. The scale-up decision gates on five per-node signals: CPU, memory, **disk IOPS**, **disk throughput**, network. Two of the five cannot be expressed, so the rule can only run in a degraded three-signal form.

Worse, the missing dimension is the binding one. On a 2 vCPU node with a 100 GiB volume, AWS gp3 and Azure Premium SSD v2 both have a 3,000 IOPS baseline. At the ~60 IOPS per MB/s we measure on a production-shaped write workload (mean write ≈ 16.2 KB), that caps the node at ~50 MB/s — well under the instance's own 81 MB/s throughput ceiling. **IOPS is the first limit reached on every cloud we run on, and it is the one signal we cannot see.**

Closest existing signals, and why none substitutes:

* `system_io_wait` — a saturation proxy, not a rate; cannot be compared against a device limit.
* `flush_bytes_written` — application bytes handed to the WAL flush path. Excludes fsync amplification, compaction reads and recovery reads. Not a device measurement.
* `read_batch_bytes_total` — same, read side.
* node_exporter `node_disk_*` — device level, but not attributable to a pod without separately resolving which device backs the PVC.

## Proposed metrics

```
woodpecker_server_system_disk_read_ops_total    {node_id, path, device}
woodpecker_server_system_disk_write_ops_total   {node_id, path, device}
woodpecker_server_system_disk_read_bytes_total  {node_id, path, device}
woodpecker_server_system_disk_write_bytes_total {node_id, path, device}
```

Counters. Source: gopsutil `disk.IOCounters()` → `ReadCount` / `WriteCount` / `ReadBytes` / `WriteBytes`.

### Implementation notes

* **Where.** `common/metrics/system_metrics_collector.go`, next to `system_disk_total_bytes` / `_used_bytes` (`:64`, `:70`), registered in `RegisterSystemMetrics` (`:85`). Collection rides the existing 15 s goroutine (`cmd/main.go:312`); no new ticker is needed.

* **`path` is not optional, and `device` alone is wrong.** Inside a container `/proc/diskstats` reports *every host block device*, not this pod's. If the metric is exported keyed only by `device`, a node running two logstore pods has **both pods reporting the whole host**, and any `sum()` across pods double counts. Resolve the WAL directory's device instead: `stat` the path → `major:minor` → `/sys/dev/block/<maj>:<min>` → device name, and export only that one device, carrying the same `path` label the existing disk gauges already use. Each PVC is its own cloud volume and therefore its own block device, so this also keeps two co-located pods' series distinct.

* **Prefer a custom Collector over tracking deltas.** Device counters are cumulative since boot and already monotonic. Report them as-is via `prometheus.MustNewConstMetric(desc, prometheus.CounterValue, raw, labels...)` rather than computing a delta and calling `Add()`. The delta approach loses data across scrape gaps and mishandles device counter wrap/reset.

### Acceptance

Two self-checks that need no new test rig:

1. `rate(write_ops) / (rate(write_bytes) / 1e6)` should land near **60**, i.e. mean write size ≈ **16.2 KB**. That is the ratio measured on a production-shaped workload; a materially different number means the collection is picking up the wrong thing.
2. `rate(system_disk_write_bytes_total)` should be **slightly above** `rate(flush_bytes_written)` — the gap is filesystem plus fsync amplification. If it comes out *lower*, the wrong device was selected.

Both server dashboards already have a **System** row; add `Disk IOPS` and `Disk Throughput` panels there:

* `tests/k8s/monitor/grafana/templates/dashboard_server_k8s.json`
* `tests/docker/monitor/grafana/templates/dashboard_server.json`

tinswzy:
Closes #337.

## What

Four counters on the block device backing the WAL directory:

```
woodpecker_server_system_disk_read_ops_total    {node_id, path, device}
woodpecker_server_system_disk_write_ops_total   {node_id, path, device}
woodpecker_server_system_disk_read_bytes_total  {node_id, path, device}
woodpecker_server_system_disk_write_bytes_total {node_id, path, device}
```

Source is gopsutil's `disk.IOCounters`, already a dependency and already used in `common/hardware`.

## Why it matters

IOPS is the dimension that binds first on every cloud we run on. A 2 vCPU node with a 100 GiB volume gets a 3,000 IOPS baseline on both gp3 and Premium SSD v2; at the ~60 IOPS per MB/s a production-shaped write workload produces (mean write ≈ 16.2 KB), that caps the node at ~50 MB/s — comfortably under its own 81 MB/s throughput ceiling. The signal that actually decides capacity was the one with no metric behind it.

## Two decisions worth reviewing

**The device is resolved, not enumerated.** Inside a container `/proc/diskstats` reports *every* device on the host. Exporting keyed by `device` alone would have every logstore pod report the whole host, and with two pods per node — which the small-region layout uses — any `sum()` across pods would double count. Each PVC is its own volume with its own mount entry, so the longest matching mountpoint picks out exactly this pod's disk.

Resolution goes through the mount table rather than `/sys/dev/block/<maj>:<min>`, which avoids `dev_t` arithmetic, needs no `golang.org/x/sys` promotion, and behaves the same anywhere gopsutil runs. The selection is a pure function (`pickDevice`), so the cases that matter are tested without depending on the host — including the one this is all for: two PVCs on one node resolving to different devices, and `/var/lib/woodpecker` not swallowing `/var/lib/woodpecker-2`.

**Read at scrape, reported verbatim.** A custom collector emits the device's own cumulative value via `MustNewConstMetric` instead of sampling on the existing 15s ticker and accumulating deltas. The device counter is already monotonic, so passing it through means a missed scrape loses no increments, and a device counter reset reads as the counter reset it is rather than arriving as a negative delta that `Counter.Add` would reject. Reading `/proc/diskstats` costs microseconds — node_exporter does the same thing.

Failure is all-or-nothing: a device that cannot be resolved, or that stops reporting, emits **nothing** rather than zeros, and clears the cached name so a remount recovers on the next scrape.

## Acceptance

Two checks that need no new rig, from the issue:

1. `rate(write_ops) / (rate(write_bytes) / 1e6)` should land near **60**, i.e. mean write ≈ 16.2 KB.
2. `rate(system_disk_write_bytes_total)` should sit **slightly above** `rate(flush_bytes_written)` — the gap being filesystem and fsync amplification. Coming out *lower* means the wrong device was picked.

## Dashboards

`WAL Disk IOPS` and `WAL Disk Throughput` in the System row of both server dashboards, split read/write and broken out by `device`; configmap regenerated with `-update`.

## Test

```
go test ./common/hardware/ ./common/metrics/ ./tests/k8s/monitor   ok
go vet ./common/hardware/ ./common/metrics/ ./cmd/                 clean
```

`TestDiskIOCollector_CollectIsAllOrNothing` asserts a complete four-series set with the right labels where a device resolves, and skips where none does — so it is meaningful on Linux CI without being flaky on a developer's macOS.


czs007:

This PR adds WAL (write-ahead log) device I/O counters (`woodpecker_server_system_disk_{read,write}_{ops,bytes}_total`) and dashboard panels. The findings below concern how the WAL directory is mapped to a block device, when the collector is registered, and missing test and documentation coverage for the new metrics.

## Medium

**`common/hardware/disk_io.go:43`: The per-node metrics are wrong when several nodes' WAL directories share one block device.**
`pickDevice` takes the mount source from mountinfo (gopsutil v3.23.7 `disk_linux.go:328-333`). For Docker named volumes, hostPath volumes and local-path-provisioner PVCs (PersistentVolumeClaims), that source is the host disk that holds the volume, so every node resolves to the same `device`.
- **Failure:** starting the 4-node `deployments/docker-compose.yaml` setup (each node has a named volume on `/woodpecker/data`, all on one host disk) → all 4 `node_id`s report the full cumulative I/O of that disk, including other nodes and unrelated host processes. The WAL Disk IOPS (I/O operations per second) and Throughput panels show 4 identical series, per-node IOPS is overstated, and any cross-node `sum()` counts the I/O 4 times.
- This is the case the comment at `disk_io.go:57-61` says device resolution avoids ("both pods reporting the whole host"). That only holds when each PVC maps to its own block device. Acceptance check 2 in the PR description only treats values *below* `flush_bytes` as a wrong-device signal, so this case passes it.
- **Suggestion:** `/proc/diskstats` cannot split one device's I/O by node. At minimum, detect a shared device or document it. The docs and dashboard should state that the metric can be read per node only when each node has its own block device. Do not describe the counters as attributed per pod.

(surfaced during verification)

**`common/hardware/disk_io.go:75`: Volumes backed by LVM (Logical Volume Manager) or device-mapper produce no metrics in non-privileged containers, and nothing is logged.**
gopsutil (`disk_linux.go:344-349`) replaces `/dev/mapper/<name>` with `/dev/dm-N` only if `EvalSymlinks` on `HostDev + "/mapper/..."` succeeds. A non-privileged container's `/dev` is its own tmpfs and has no `/dev/mapper/*` links, so the name stays `/dev/mapper/vg-lv`. `filepath.Base` then returns `vg-lv`, but `/proc/diskstats` lists the device as `dm-N`.
- **Failure:** logstore runs in a non-privileged pod with an LVM/device-mapper WAL volume (for example TopoLVM, OpenEBS LVM-LocalPV, or a hostPath/Docker volume on host LVM) → `GetDeviceIOCounters` returns "no I/O counters reported" (lines 84-86) and `Collect` calls `forget()` and returns (`system_disk_io.go:101-103`). The four `system_disk_*` series are never emitted, nothing is logged, and the WAL Disk panels stay empty.
- **Suggestion:** Resolve `/dev/mapper/*` through the mount's `maj:min` using `/sys/dev/block/<maj:min>`, which is readable in containers; this is the method gopsutil already uses for `/dev/root`. Alternatively, map names through `/sys/class/block/*/dm/name` when the lookup fails. Either way, log a warning on the first failed resolution.

(surfaced during verification)

## Low

**`common/metrics/system_disk_io.go:115`: `RegisterDiskIOMetrics` is gated only on `dataPath != ""` and does not check the storage type.**
On the `wp server` path, `dataPath` is never empty:
- the `--data-dir` flag defaults to `/woodpecker/data` and overrides `RootPath` (`cmd/main.go:72,129-130`);
- `configuration.go:930-931` rejects an empty `RootPath`.

As a result, the early return at line 118 and `TestRegisterDiskIOMetrics_NoDataPath` cover a branch that production never reaches.
- **Failure:** starting `wp server` with `storage.type` set to `minio`, `default` or empty (for example the bundled `config/woodpecker.yaml`), where the root path is not a separate mount → `pickDevice` falls back to the root filesystem device. The four counters then report the whole root disk's I/O, including other processes, as WAL I/O. The device resolves successfully, so the "emit nothing rather than the whole host" guard in `Collect` does not apply.
- **Scope:** the bundled deployments all run the server with `type: service`: the image entrypoint default (`start-woodpecker.sh:29`), the operator configmap (`reconcile_configmap.go:34`) and `docker-compose.yaml`. They are therefore unaffected. Only configs that use `minio`, `default` or an empty type hit this.
- **Suggestion:** Match the existing local-WAL check in `server/logstore.go:164-165` (`IsStorageService() || IsStorageLocal()`). For example, compute a bool at `cmd/main.go:309` and pass it in, or pass `cfg`. Update the function comment so it describes the actual behavior.

(surfaced during verification)

**`common/metrics/system_disk_io.go:95`: Failed device resolution is not cached, so every scrape re-reads the mount table.**
The `resolve()` comment (lines 72-74) says resolution "is not repeated per scrape". Two paths contradict this:
1. When `DeviceForPath` fails (for example, the WAL directory is on the overlay root with no separate mount), `c.device` is never set.
2. When the resolved name is not in diskstats (for example, the `/dev/mapper` case above), `forget()` clears the cache.

On both paths, every scrape re-reads `/proc/1/mountinfo` and `/proc/filesystems` and fails the same way. The cost per scrape is small and there is no functional impact.
- **Suggestion:** Add negative caching or backoff after a failed resolution, or correct the comment.

(surfaced during verification)

**`tests/docker/monitor/monitor_test.go:385`: The four new metrics are missing from the Docker monitor integration test and the README metrics table.**
- Both `SystemMetrics` subtests (lines 149-153 and 385-393) assert only the existing cpu, memory, disk_total, disk_used and io_wait metrics.
- `tests/docker/monitor/README.md:206-213` does not list the new metrics either.
- The unit test `TestDiskIOCollector_CollectIsAllOrNothing` calls `t.Skip` when no device can be resolved. That makes this Docker setup the only test that checks, inside a container, that a device resolves and all four series are emitted.
- **Suggestion:** Add `assertMetricExists` (or `assertMetricRegistered`) for `woodpecker_server_system_disk_{read,write}_{ops,bytes}_total` to the subtest at line 385, and add the matching rows to the README table.

(surfaced during verification)

PR #345, opened 2026-09-18
enhance(metrics): count completed segment rolls by what triggered them

### LINKED ISSUE #340: enhance(metrics): segment rolling has no counter, so roll frequency is only visible indirectly
## Summary

There is no counter for segment rolling. Roll frequency is only inferable, from `woodpecker_client_segment_state` transitions, `woodpecker_client_write_frontier_segment` stepping, or `logstore_operations_total{operation="complete"}` — none of which says *why* a segment rolled.

## Why it matters beyond curiosity

Roll size drives compaction volume, and compaction upload is the dominant term in a logstore pod's network egress. On a measured production-shaped workload, per-pod network ran **1.53×** the pod's own disk write rate (40 MB/s to the volume, 44 MB/s in + 17 MB/s out on the wire), and ~78% of that egress is compaction `PutObject` traffic — each logical byte uploaded exactly once, by whichever replica wins `compactSegmentQuorum`.

We use that 1.53 factor for network capacity planning. It is not a constant: it moves with `segmentRollingPolicy` (`maxSize` 256 MB / `maxInterval` 600 s / `maxBlocks` 1000) and the compaction settings. If someone retunes those, our planning constant silently goes stale and **no metric today would say so**. A roll counter broken down by trigger is the first place anyone would look.

## Proposed metric

```
woodpecker_server_segment_rolled_total{node_id, log_ns, log_id, reason}
```

`reason` ∈ `size` | `interval` | `blocks`, matching the three trigger conditions in `woodpecker/segment/rolling_policy.go:66-123`.

Counter, incremented at the point the policy decides to roll.

## Priority

Low. Diagnostic only — not proposed for any alert rule. Filed so the reasoning is recorded rather than rediscovered.

tinswzy:
Closes #340.

## What

```
woodpecker_client_segment_rolled_total{log_ns, log_id, reason}
```

`reason` is `size`, `blocks`, `interval` or `force`.

## Why, beyond curiosity

Roll size drives compaction volume, and compaction upload is the dominant term in a logstore pod's object-storage egress. On a production-shaped workload, per-pod network ran about **1.53×** the pod's own disk write rate — 40 MB/s to the volume against 44 MB/s in + 17 MB/s out on the wire — with roughly three quarters of the egress being compaction `PutObject` traffic.

We use that 1.53 for network capacity planning. It is not a constant: it moves with `segmentRollingPolicy`. If someone retunes `maxSize` or `maxInterval`, the planning number goes stale and **no metric today would say so**. A roll counter broken down by trigger is the first place anyone would look.

## The interface change

The reason has to come from the policy, so `ShouldRollover` now returns `(bool, string)`. The alternative — re-deriving it at the call site from size, blocks and elapsed time — is a second copy of the policy's conditions that would drift from the first.

It touches one production call site (`log_handle.go`) and the existing tests, which keep asserting just the decision through a small helper so that diff stays mechanical. A new test pins the label each branch returns, including the ordering case: **`size` wins when several limits are over at once**, so a segment already past its size budget is not misattributed to `interval`.

## Counted at completion, not at the decision

A roll can still be aborted after the decision — `ErrMetadataRevisionInvalid` or `ErrMetadataSegmentAlreadyExists` both mean another writer took the log over, and the code correctly invalidates the writer instead of rolling. The increment sits at the "segment rolling completed successfully" point so those never show up as rolls.

## Correction to the issue

This is a **client** metric, not a server one. `woodpecker/segment` runs in the client process and emits the `woodpecker_client_*` family, so it carries `log_ns` and `log_id` and no `node_id`. Issue #340 proposed `woodpecker_server_segment_rolled_total{node_id, ...}`, which was wrong.

## Scope

Go only. No dashboard panel — panels are landing together with the alert-aligned panel work rather than colliding with it.

```
go build ./...                                    ok
go vet ./...                                      clean
go test ./woodpecker/segment/ -run Rolling        ok
go test ./woodpecker/log/ -run Rolling            ok
go test ./common/metrics/                         ok
```


czs007:

This PR changes `ShouldRollover` to return `(bool, string)` and adds a per-log `woodpecker_client_segment_rolled_total` counter labelled by roll reason. Three low-severity follow-ups are listed below: one on how retried rolls are labelled, one on test coverage, and one on the metric's Help text.

**Verified:**
- `woodpecker/log/log_handle.go:668`: this is the only production caller of `ShouldRollover`. `DefaultRollingPolicy` (`rolling_policy.go:76`) is its only implementation, and no mock implements `RollingPolicy`, so no call site was missed when the signature changed.
- `woodpecker/log/log_handle.go:562`: the three label values `(l.logNs, logIdStr, rollReason)` match the CounterVec's `{log_ns, log_id, reason}`. `logIdStr` is defined at `:478` and `rollReason` is bound at `:517`, so `WithLabelValues` cannot panic on label count.
- `woodpecker/log/log_handle.go:533-540` / `:547-554`: both takeover-abort branches (`ErrMetadataRevisionInvalid`, `ErrMetadataSegmentAlreadyExists`) return before the increment at `:562`, so a roll taken over by another writer is not counted.
- `woodpecker/segment/rolling_policy.go:96`: every `return true` path (size `:96`, blocks `:104`, interval `:128`, plus force at `log_handle.go:663`) returns a non-empty reason, and every `return false` path returns `""`. No `reason=""` series is emitted.
- `woodpecker/log/log_handle.go:492`: the increment runs inside the `l.Lock()` critical section, `Counter.Inc` is goroutine-safe, and no new shared state is added.
- `common/metrics/client_metrics.go:369`: the metric is registered inside `RegisterOnce`. `reason` has a fixed set of 4 values, and `log_ns`/`log_id` have the same cardinality as the existing per-log client metrics.

## Low severity (non-blocking)

- **`woodpecker/log/log_handle.go:661`**: A retried roll is counted as `reason="force"` and its original trigger is lost. `SetRollingReady` sets `rollingState` to true (`segment_handle.go:2180`), and only the constructors ever reset it. Suppose `createAndCacheWritableSegmentHandleWithID` / `createNewSegmentMetaWithID` then fails, either because `selectQuorumFunc` cannot select a quorum or because of a transient etcd error or ctx timeout in `storeNewSegmentMeta`. The call returns at `:556` and the old handle stays writable. On the next append, `IsForceRollingReady` returns early at `:661-663`. **Failure:** a size-triggered roll fails to create the new segment's metadata once, then succeeds on the next append → `woodpecker_client_segment_rolled_total` records `{reason="force"}` +1 and `{reason="size"}` +0. During node failures or etcd instability, size/interval rolls are therefore undercounted and force rolls overcounted, which would mislead anyone tuning `segmentRollingPolicy` from this metric. The "size wins" case in `rolling_policy_test.go` only checks the order of conditions inside the policy and does not cover this path across two calls.
  **Suggestion:** Use one of these fixes:
  - **(A)** Store a `pendingRollReason` on `logHandleImpl`. Set it when the roll is decided at `:517`, return it from the force branch when it is non-empty, and clear it after the increment at `:562`. All reads and writes happen under `l.Lock()`.
  - **(B)** Have the segment handle record whether `rollingState` was set by `SetRollingReady` or by `HandleAppendRequestFailure` (`segment_handle.go:496`), and return `force` only for the second case.

  Add a test where `selectQuorumFunc` fails once and then succeeds, and assert `{reason="size"}` = 1 and `{reason="force"}` = 0. (raised by czs007)

- **`woodpecker/log/log_handle.go:562`**: No test checks the value of `WpSegmentRolledTotal` from the `log_handle` roll path, and no test confirms that the two abort branches (`:533-540` `ErrMetadataRevisionInvalid`, `:547-554` `ErrMetadataSegmentAlreadyExists`) leave it unchanged. The existing tests only cover the reason returned by the policy (`rolling_policy_test.go`) and whether the metric can be gathered (`metrics_registration_test.go`). If a later change moves `Inc()` to where the roll is decided, or before an abort return, no test would fail.
  **Suggestion:** Add a `log_handle` test that runs one completed roll and both abort paths, and assert the counter delta per `reason` for each. (raised by czs007, xiaocai2333)

- **`common/metrics/client_metrics.go:238`**: The Help text "Segment rolls that completed, by the limit that triggered them" and the `RollReasonForce` constant do not describe what `force` counts. `force` is not a limit. `rollingState` is set in only two places:
  1. `segment_handle.go:496` (`HandleAppendRequestFailure`), when an append still fails after its retries.
  2. `segment_handle.go:2180` (`SetRollingReady`), when an earlier size/blocks/interval roll started but failed to create the new segment and is now being retried.

  Operators reading `reason="force"` cannot tell these two cases apart. Renaming it to something like `append_failure` would also be inaccurate while retries are still counted there.
  **Suggestion:** Document both cases in the Help text and in the `RollReasonForce` comment in `rolling_policy.go`. Alternatively, fix the first item above so retries keep their original reason. `force` would then mean only "append failed", and a more specific name could be chosen. (raised by czs007, xiaocai2333)

PR #347, opened 2026-09-20
enhance(deps): move gopsutil to v4.25.10, matching Milvus

### LINKED ISSUE #346: enhance(deps): align gopsutil with Milvus on v4, we are why v3 is still in its tree
## Summary

Woodpecker pins `github.com/shirou/gopsutil/v3 v3.23.7`. Milvus moved to **v4** on 2024-11-22 (milvus-io/milvus#37905) and its main module now requires `github.com/shirou/gopsutil/v4 v4.25.10` directly.

Since Milvus embeds the Woodpecker client, that split lands in one binary.

## We are the reason v3 is still there

`go mod why -m github.com/shirou/gopsutil/v3`, run in `milvus/pkg`:

```
github.com/milvus-io/milvus/pkg/v3/streaming/walimpls/impls/wp
github.com/zilliztech/woodpecker/common/metrics
github.com/zilliztech/woodpecker/common/hardware
github.com/shirou/gopsutil/v3/cpu
```

Milvus's own modules carry v3 only as `// indirect`:

| module | gopsutil |
| --- | --- |
| woodpecker | `v3 v3.23.7` (direct) |
| milvus (root) | `v4 v4.25.10` (direct), `v3 v3.23.12` and `v3.20.11+incompatible` (indirect) |
| milvus/pkg | `v4 v4.25.10` (direct), `v3 v3.23.12` (indirect) |

So Milvus ships two majors of the same system-metrics library, and one of them is ours.

A comparison of all shared dependencies between the two repos turns up 40 version differences, but this is **the only one where Woodpecker's choice forces an extra major version into Milvus**. The others (`jwt`, `grpc-gateway`) look like splits at a glance but both repos carry both majors identically, and `backoff/v4` is Milvus's own direct requirement. The remaining ~36 are same-major differences that MVS resolves upward in a Milvus build and only affect Woodpecker's standalone builds.

## The migration is small

Woodpecker uses six gopsutil symbols in total:

```
cpu.Percent   cpu.Times   disk.Usage   mem.VirtualMemory   process.NewProcess   process.Process
```

Every one of them is already in use on v4 inside Milvus's own `pkg/util/hardware` — which is where Woodpecker's `common/hardware` was originally copied from. Compatibility is therefore already proven in production by the consumer.

The two symbols added recently for the WAL device counters were checked against v4 directly:

```go
func Partitions(all bool) ([]PartitionStat, error)
func IOCounters(names ...string) (map[string]IOCountersStat, error)
// IOCountersStat still has ReadCount / WriteCount / ReadBytes / WriteBytes
```

Identical signatures and field names.

## Scope

Import path change in the five files under `common/hardware` that touch gopsutil, plus `go.mod`/`go.sum`. No behaviour change intended; anything that does change behaviour is a finding to report, not to absorb.

Target `v4.25.10` — matching Milvus exactly rather than taking the newest release, so the two stay on one version.

## Not in scope

This does not remove v3 from Milvus's tree on its own: `github.com/google/gops` and `github.com/apache/pulsar-client-go` also require it. Pulsar is what Woodpecker replaces, so after that migration `gops` is the only holder left.

tinswzy:
Closes #346.

## What

`github.com/shirou/gopsutil/v3 v3.23.7` → `github.com/shirou/gopsutil/v4 v4.25.10`, matching Milvus exactly rather than taking the newest release, so the two stay on one version.

Import path change in the five files under `common/hardware` that touch gopsutil, plus `go.mod`/`go.sum`. No behaviour change.

## Why this version, and why now

Milvus moved to v4 on 2024-11-22 (milvus-io/milvus#37905) and requires `v4.25.10` directly. Woodpecker stayed on v3, and since Milvus embeds the Woodpecker client, both majors end up in one binary — with `go mod why -m github.com/shirou/gopsutil/v3` in `milvus/pkg` naming us as the reason:

```
github.com/milvus-io/milvus/pkg/v3/streaming/walimpls/impls/wp
github.com/zilliztech/woodpecker/common/metrics
github.com/zilliztech/woodpecker/common/hardware
github.com/shirou/gopsutil/v3/cpu
```

## Compatibility was already proven by the consumer

Woodpecker uses six gopsutil symbols: `cpu.Percent`, `cpu.Times`, `disk.Usage`, `mem.VirtualMemory`, `process.NewProcess`, `process.Process`. Every one is already in use on v4 inside Milvus's own `pkg/util/hardware` — the package `common/hardware` was originally copied from.

The two added for the WAL device counters were checked against v4 before this change: `Partitions(all bool) ([]PartitionStat, error)` and `IOCounters(names ...string) (map[string]IOCountersStat, error)`, with `IOCountersStat` keeping `ReadCount` / `WriteCount` / `ReadBytes` / `WriteBytes`. Identical.

## It aligns more than gopsutil

v4 pulls a newer transitive set, and each one lands on the version Milvus already has:

| module | before | after | Milvus |
| --- | --- | --- | --- |
| `power-devops/perfstat` | `20210106…` | `20240221…` | `20240221…` |
| `tklauser/go-sysconf` | v0.3.11 | v0.3.15 | v0.3.15 |
| `tklauser/numcpus` | v0.6.0 | v0.10.0 | v0.10.0 |
| `yusufpapurcu/wmi` | v1.2.3 | v1.2.4 | v1.2.4 |
| `ebitengine/purego` | — | v0.9.0 | v0.9.0 |
| `shoenig/go-m1cpu` | v0.1.6 | dropped | — |

So one dependency bump closes six of the version differences between the two repos, not one. (v4 replaced the `go-m1cpu` Apple Silicon path with `purego`.)

## Not claimed

This does not remove v3 from Milvus's tree by itself — `github.com/google/gops` and `github.com/apache/pulsar-client-go` also require it. Pulsar is what Woodpecker replaces, so `gops` is the last holder after that migration.

```
go build ./...                                   ok
go vet ./...                                     clean
go test ./common/hardware/ ./common/metrics/     ok
go test ./common/...                             ok
```


czs007:

**Verified:**
- `go.mod:31`, `go.sum:409-410` — only `gopsutil/v4` remains across the module graph (6 files, 12 import sites); no v3 `h1:`/`go.mod` lines survive, so there is no silent two-major coexistence.
- `common/hardware/disk_io.go:88-94` — `PartitionStat` and `IOCountersStat` field sets are byte-for-byte identical between v3.23.7 and v4.25.10, and v4 still uses `sectorSize = 512` with `/proc/diskstats` `fields[2]` as the key, so counter magnitudes and device matching are unchanged.
- `common/hardware/hardware_info.go:118,135` — v4's `disk.Usage` still returns the raw `unix.Statfs` errno without wrapping, so the `errors.Is(err, oserror.ErrNotExist)` checks still match ENOENT and the "missing path yields zero values" contract holds.
- `common/hardware/mem_info.go:44,51` — v4 redefines `VirtualMemoryStat.Used` on Linux, but the Linux path reads `proc.MemoryInfoEx()` (`RSS - Shared`) and the callers at `hardware_info.go:74,100` only read `.Total` (MemTotal, unchanged); the darwin/windows `.Used` formulas are identical to v3.
- `common/hardware/hardware_info.go:132-141` — v4's `UsageWithContext` computes `Free = Bavail * bsize` and `Used = (Blocks - Bfree) * bsize`, matching the comment this code depends on, so df-style ratio semantics are preserved.
- `common/hardware/disk_io.go:24,35` — no gopsutil type crosses the public API surface; the only consumer of `disk.PartitionStat` is the unexported `pickDevice`, so downstream callers are not forced to move majors.

One scope note on the evidence above: no Go toolchain was available in the review environment, so `go build ./...` / `go test` were not executed. The compatibility conclusions come from line-by-line comparison of the v3.23.7 and v4.25.10 sources in the module cache rather than from a build.

PR #348, opened 2026-09-20
enhance(monitor): add an Alerts row plotting what the rules actually threshold

### LINKED ISSUE #341: enhance(monitor): dashboards plot raw metrics, not the quantities alerts actually threshold
## Summary

The client and server dashboards are thorough — 29 panels across 5 rows on the client, 10 rows on the server — and they are good at **exploring**: append latency percentiles, per-scope replica traffic, frontier positions, watermark state.

What they do not contain is **the quantity any alert rule actually thresholds**. Those are derived, and no panel computes them. When a page fires, the on-call opens the dashboard and sees a related series, not the series that crossed the line.

Concrete examples from the rule set we are standing up:

| alert thresholds | nearest existing panel | same quantity? |
| --- | --- | --- |
| share of appends slower than 256 ms, > 1% | `Append Latency P50/P99` | no — a percentile, not a share |
| share of reads slower than 512 ms, > 1% | `Read Latency P50/P99` | no |
| single-volume IOPS as a fraction of the device ceiling, > 50% | — | no panel |
| cluster mean of five signals all > 50% | — | no panel |
| object-storage error ratio > 5% | ops rate by status | close, but the ratio is not plotted |
| slow-write share vs the same window 7 d ago | — | no panel |

## Why this is worth a panel even though it duplicates data

It is deliberately the same series the rule reads, and that is the point. If a panel plots a *near* equivalent instead, investigations turn into arguing about why the alert says one thing and the graph says another. Same expression on both sides means the graph is evidence, not a second opinion.

The practical need is narrow and specific: when an alert fires, answer **from when to when was this out of bounds, and by how much** — which needs the thresholded expression plotted over time with its threshold drawn on it.

## Proposal

Add an **Alerts** row to each dashboard, one panel per rule, each plotting:

1. the rule's expression, verbatim, and
2. the threshold as a second series — read from the recording-rule constants (`wp:vol_iops_limit`, `wp:vol_bw_limit_bytes`, `wp:node_net_limit_bytes`, `wp:replica_factor`) rather than hardcoded in the panel, so panel and rule cannot drift apart.

Split follows the existing dashboard split, which already matches how the rules divide:

* **server dashboard** — node resource and capacity rules, disk watermark rules, object-storage failure rules, flush latency, sync-scheduler queueing, local-data reclamation
* **client dashboard** — per-log append/read latency shares, pending appends, frontier stalls, direct-read failures, compaction failures, active-segment replica count

Files:

```
tests/k8s/monitor/grafana/templates/dashboard_server_k8s.json
tests/k8s/monitor/grafana/templates/dashboard_client_k8s.json
tests/docker/monitor/grafana/templates/dashboard_server.json
tests/docker/monitor/grafana/templates/dashboard_client.json
```

### Note on panel titles

Title each panel with the rule id it mirrors, so a page that names a rule leads straight to its panel with no lookup step.

### Possible test hook

`tests/k8s/monitor/dashboards_test.go` already validates the dashboards. If the alert rules land in-repo, that test is the natural place to assert every rule has a panel whose expression matches — which is what keeps the two from drifting once someone edits a threshold.

## Depends on

Two of the server-side panels cannot be written until the device IOPS/throughput metrics exist. The rest are writable today.

tinswzy:
Closes #341.

## What

An **Alerts** row on each dashboard — **13 panels** on the server, **8** on the client — each plotting the expression an alert rule thresholds, with the threshold drawn on it.

## Why, when the dashboards are already thorough

They are good at exploring. They are not what you want open when a page fires, because **the quantity the rule thresholds is derived and no panel computed it**:

| the rule thresholds | nearest existing panel | same quantity? |
| --- | --- | --- |
| share of appends slower than 256 ms, > 1% | `Append Latency P50/P99` | no — a percentile, not a share |
| share of reads slower than 512 ms, > 1% | `Read Latency P50/P99` | no |
| single-volume IOPS as a fraction of the device ceiling | — | no panel |
| cluster mean of five signals all over 50% | — | no panel |
| object storage error **ratio** | ops rate by status | close; the ratio itself was not plotted |
| slowest node's flush p99 over the median | — | no panel |

Plotting the same series the rule reads is the point. A near-equivalent turns an investigation into an argument about why the alert says one thing and the graph another.

## Normalising to "fraction of the limit"

Every signal with a cloud-dependent ceiling is drawn as a fraction of that ceiling. The line is then a fixed `0.5` or `0.7` on every deployment, and the per-cloud numbers live in three textbox variables — `vol_iops_limit`, `vol_bw_limit`, `node_net_limit` — an operator sets once.

It also makes R2 readable at a glance: five series, one line, and the rule fires only when all five are above it. Without normalisation that panel would mix percent, ratio, IOPS and bytes/s on one axis.

Thresholds use the `thresholdsStyle: dashed` + `thresholds.steps` pattern already present on `WAL Disk Usage Ratio`, so no fake constant series is needed.

## Panels carry the traps

Each panel is titled by the rule ids it mirrors and carries a description saying what the rule is really asking. The ones that have already cost us a dead rule are written down:

- the object-storage and flush histograms are **milliseconds** while `grpc_server_handling_seconds` is **seconds** — the F9 panel says so, because "p99 > 5s" has to be written `5000`
- `append_latency` and `read_latency` top out at **512 ms**, which is why C1/C5 count the tail past a bucket boundary rather than asking for a quantile the buckets cannot express

## Two panels that had no home before

- **S5** divides max by mean over `quorum_node_selected_total` (#343), instead of approximating skew through the spread of `active_segments`, which conflates it with logs of uneven size.
- **R1/R2** get their IOPS and throughput terms from the WAL device counters (#344).

## Safety of the diff

The four dashboard templates take **additions only** — no existing panel is touched. `dashboard-configmaps.yaml` is regenerated with `-update`; its 199 deletions are the diff realigning around inserted content, and I checked that every deleted line reappears elsewhere (net loss: 0 lines).

Network uses `process_network_*` with the same selector as the existing `Network Traffic` panel, rather than cAdvisor, so the series works in the docker stack too.

All 31 metric references in the new panels were checked against the metric definitions in `common/metrics` before committing.

```
go test ./tests/k8s/monitor   ok
```


czs007:
> **Must-fix issues introduced by this PR: 4**
> **Merge recommendation:** Mergeable after the 4 must-fix items below are fixed in this PR.

## Overview

This PR adds 21 alert-threshold panels across four dashboard templates and the generated k8s configmap; several of the new PromQL targets do not parse, filter on labels the metrics do not carry, or apply `delta()` to per-segment counters that reset, so a number of the new panels will not show usable data.

---

## Critical

**`tests/docker/monitor/grafana/setup.go:84` — `{SEL, ...}` placeholder is never expanded, breaking 6 panels' PromQL**
21 bare `SEL` tokens across 18 new target strings land in 4 templates plus the generated configmap (docker `dashboard_client.json` ×3, docker `dashboard_server.json` ×4, `dashboard_client_k8s.json` ×3, `dashboard_server_k8s.json` ×4, `dashboard-configmaps.yaml` ×7), covering 6 distinct panels: C1/C2, C5, C7, F6, F7/F8, S3. Nothing expands it — the docker path does a single `strings.ReplaceAll(d.template, "__DATASOURCE_UID__", dsUID)` at `setup.go:84` (`__PROMETHEUS_URL__` at `:118` only serves the datasource template), and the k8s path is a Grafana sidecar reading the configmap verbatim, with `renderDashboardConfigMaps` (`dashboards_test.go:80-103`) only re-indenting. A repo-wide grep for `SEL` finds no expansion logic; all other hits are substrings of `BASELINE`. In PromQL, `{SEL, le="256"}` is a bare identifier in label-matcher position and Prometheus rejects it with `unexpected "," in label matching`. These 18 targets are also the only new ones missing `namespace=~"$namespace"` / `woodpecker_id=~"$woodpecker_id"`, which indicates the whole selector was swallowed by the placeholder.
*Failure:* importing the dashboard via `tests/docker/monitor` or `tests/k8s/monitor` and opening the "Alerts — what the rules actually threshold" row → all six panels error out at the label matcher and permanently render no data, i.e. exactly the screen the PR describes as "the one you open when an alert fires."
*Suggestion:* either add a real macro expansion step for `SEL` alongside the `__DATASOURCE_UID__` replacement (and mirror it in the configmap generator), or inline the intended selector (`namespace=~"$namespace", woodpecker_id=~"$woodpecker_id"`, matching the surrounding targets) directly into all 18 targets and regenerate `dashboard-configmaps.yaml`.
*(raised by bigsheeper, czs007)*

---

## High

**`tests/docker/monitor/grafana/templates/dashboard_server.json:4094` — F6 filters on gRPC methods that do not exist**
F6 ("gRPC append handling p99") filters `grpc_method=~"AppendEntry|AppendEntries"`, but the proto defines `rpc AddEntry` / `rpc AddEntries` (`proto/logstore.proto:52,57`). The `grpc_method` label is taken straight from the proto method name by go-grpc-prometheus (registered at `server/service.go:247-250`), and no `AppendEntry` method exists anywhere in the repo. This is independent of the `SEL` issue: fixing the selector still leaves F6 empty.
*Failure:* opening F6 on any deployment, or copying the expression into a "gRPC append p99 > 1s" alert rule → the regex matches no series, so the panel is permanently empty and the rule never fires, hiding end-to-end append latency regressions (including sync wait, see `server/service.go:415-465`).
*Suggestion:* change the filter to `grpc_method=~"AddEntry|AddEntries"` in all three places: `dashboard_server.json:4094`, `dashboard_server_k8s.json:4094`, `dashboard-configmaps.yaml:4104`.
*(raised by bigsheeper, czs007)*

**`tests/k8s/monitor/grafana/templates/dashboard_server_k8s.json:3715` — node-level metrics filtered by `log_ns`/`log_id`, which they do not carry**
Several node-scoped metrics have no `log_ns`/`log_id` labels: `system_cpu_usage` / `system_memory_usage_ratio` are `{node_id}` (`common/metrics/system_metrics_collector.go:39,63`), `system_disk_write_ops_total` / `system_disk_write_bytes_total` are `{node_id,path,device}` (`system_disk_io.go:51`), `logstore_disk_free_bytes` is `{node_id,path}` (`service_metrics.go:119`), `sync_scheduler_waiting|capacity` are `{node_id}` (`:344,350`), `compaction_admission_rejected_total` is `{node_id,reason}` (`:396`), `quorum_node_selected_total` is `{node_id,selected_node_id,mode}` (`:411`). The new panels nonetheless add `log_ns=~"$log_ns", log_id=~"$log_id"` to all of them. Affected: 6 of the 13 new server panels — R1 (all three targets), R2 (four of five; only the `job`-filtered network target survives, so the "five signals above 50% at once" rule can never be fully evaluated), R6 (both targets), R7, S1, S5. Note that F4/F5, S4, F9 and S6 use `file_flush_latency`, `object_storage_operation_latency`, `compacted_cleanup_reclaimed_bytes_total` and `file_stored_bytes`, which *do* carry `log_ns`/`log_id` — as do `object_storage_operations_total` (`service_metrics.go:231`) and `logstore_operations_total` (`:92`) — so a blanket removal would over-correct. `woodpecker_id` is a scrape-time target label (`tests/docker/monitor/prometheus.yml:13`, `manifests/podmonitor-server.yaml:27-31`) and should stay.
*Failure:* switching the `log_ns` (or `log_id`) variable from All to a specific log, e.g. while investigating one slow log → R1, R2's four resource targets, R6, R7, S1 and S5 all go to No data at once; because `allValue` is `.*` (which matches the absent label), the default view looks fine, so this only surfaces on drill-down and reads like a scrape problem.
*Suggestion:* drop `log_ns`/`log_id` from the targets of R1, R2 (resource targets), R6, R7, S1 and S5 in both server templates and the configmap, matching the existing panels, which use `{woodpecker_id=~"$woodpecker_id"}` only. Leave the log-scoped panels listed above untouched.
*(raised by bigsheeper, czs007)*

**`tests/docker/monitor/grafana/templates/dashboard_client.json:1709` — C4/C6 apply `delta()` to entry-id gauges that reset every segment**
C4 uses `sum by (log_ns, log_id) (delta(woodpecker_client_write_frontier_entry[5m]))` and C6 uses `sum by (log_ns, log_id, reader_name) (delta(woodpecker_client_read_frontier_entry[10m]))` (`:1834`), and both panel descriptions treat delta≈0 as the stall criterion (C6 states it is the only stuck-reader signal). Entry ids are per-segment and restart at zero — `common/metrics/client_metrics.go:401` says so explicitly ("Entry ids restart at zero in every segment, which is also why lag is measured in segments"), and `segment_handle.go:389` passes the in-segment LAC. `delta()` does no counter-reset handling, so any window spanning a segment roll yields "new small entry id − old large entry id". Rolls are triggered by `maxSize 256MB` / `maxBlocks 1000` / `maxInterval 600s` (`config/woodpecker.yaml:32-34`), far more often than 10 minutes under load. Existing Progress Frontiers panels (`:786,801`) plot the raw gauges and never take a delta; `delta()` is new in this PR.
*Failure:* a steadily written log rolling normally at 256MB/600s with a follower reader, where C6's 10m window spans one roll → `delta(read_frontier_entry[10m])` goes large-negative for the healthy reader, while a reader genuinely stuck inside a segment reads exactly 0 and sorts *above* it on the same chart; read by the panel's own criterion, real stalls are missed and normal rolls produce negative noise.
*Suggestion:* measure displacement on the companion segment gauges (`client_metrics.go:54,90`), e.g. `(delta(*_frontier_segment[5m]) > 0) or delta(*_frontier_entry[5m])`, or `changes(*_frontier_segment[5m])`. C6 additionally needs a baseline series (writer-side frontier or append rate) to separate idle from stuck. Apply to `dashboard_client.json:1709/1834`, the k8s template at the same lines, and `dashboard-configmaps.yaml:6455/6580`.
*(raised by czs007)*

---

## Medium

**`tests/docker/monitor/grafana/templates/dashboard_server.json:4506` — S5 skew ratio reads extreme starvation as perfect balance**
S5 computes `max(sum by(selected_node_id)(rate(quorum_node_selected_total[15m]))) / avg(same)` with thresholds `[null→green, 3→yellow]`. The metric only creates a child series when a node is actually selected: `recordSelection` iterates just the `selected` slice (`common/membership/service_discovery.go:1212-1219`), `WpQuorumNodeSelected` is a plain `CounterVec` (`service_metrics.go:406-411`), and registration is `MustRegister` only — no per-member pre-initialization, unlike `SetNodeDecommissionProgress`, which pre-sets 0 for its three known states (`:487-490`). Nodes that were selected earlier but not in the last 15m do keep a series at rate 0 and are counted in `avg`, so partial skew is still detectable; the blind spot is "never selected since process start".
*Failure:* a 4-node cluster where load-aware weighting means node D has never been returned by `SelectNodes` since startup (or D was just added by scale-out) and the other 3 are selected evenly → the `selected_node_id="D"` child never exists, `max/avg ≈ 1.0` over the remaining 3 series, and S5 shows a flat green line well under the 3.0 threshold, so a node carrying no quorum traffic at all looks perfectly balanced.
*Suggestion:* divide by the actual cluster node count instead, e.g. `max(...) / (sum(rate(...[15m])) / count(count by (node_id)(system_cpu_usage{...})))`, or add a separate "nodes selected 0 times in the last 15m" panel. This panel is also affected by the `log_ns`/`log_id` issue above (`quorum_node_selected_total` has neither label) — both need fixing.
*(raised by bigsheeper)*

**`tests/docker/monitor/grafana/templates/dashboard_server.json:3725` — R1/R2 compare write-only IOPS and bandwidth against read/write-shared device limits**
R1's IOPS target (`:3725`) and R2's IOPS/bandwidth targets (`:3803`, `:3808`) divide `sum by (node_id)(rate(system_disk_write_ops_total[5m]))` / `..._write_bytes_total` by `$vol_iops_limit` / `$vol_bw_limit`, whose defaults (3000 / 81000000) are device-level cloud-volume limits shared between reads and writes. The collector also exports `system_disk_read_ops_total` / `system_disk_read_bytes_total` (`system_disk_io.go:58,60`, all four emitted at `:109-112`), and they are not counted, so the numerator covers roughly half the load and saturation is systematically understated — which undercuts the stated purpose of R1/R2 (plotting usage as a fraction of its own limit). Verified separately: the collector registers one WAL path / one device under `sync.Once` (`system_disk_io.go:117-124`, single call site at `cmd/main.go:309`), so `sum by (node_id)` does not double-count; reads are the only gap.
*Failure:* a 3000-IOPS volume during a read-heavy phase (catch-up reads or compaction reading the WAL device) at 2500 read + 400 write IOPS → R1's IOPS target reads 400/3000 ≈ 0.13, far below the 0.5 yellow line, while the device is at ~97% of its limit and latency is degrading; alert rules built on the same expression stay silent.
*Suggestion:* include the read counters in the numerator, e.g. `sum by (node_id)(rate(system_disk_read_ops_total[5m]) + rate(system_disk_write_ops_total[5m]))` and the equivalent for bytes, in `dashboard_server.json:3725,3803,3808`, `dashboard_server_k8s.json` at the same lines, and the configmap.
*(raised by czs007, bigsheeper)*

---

## Low

**`tests/docker/monitor/grafana/templates/dashboard_server.json:3914` — threshold lines at 0 are invisible for non-negative expressions**
R7 (expression at `:3954`, `increase(...[15m])`, thresholds `:3914-3926`), C7 (`dashboard_client.json:1902`, thresholds `:1862-1872`) and C8 (`:1970`, thresholds `:1930-1940`) all use `thresholds.steps` `[{green,null},{yellow,0}]` with `thresholdsStyle.mode: "dashed"`. That style draws one line at each non-base step value; since these three expressions have range [0, +∞), the line sits at y=0 and coincides with the axis baseline. The intent is "strictly greater than 0", which step semantics (≥ value takes the color) cannot express. This does not match the PR's stated goal of "each plotting the expression an alert rule thresholds, with the threshold drawn on it" (C4, C6 and S6 have `thresholds: null` and likewise draw no line). The data itself plots correctly — only the threshold marker is missing.
*Suggestion:* set the step to a small positive value (e.g. 0.001), or switch to `thresholdsStyle.mode: "area"` so the entire region above 0 is marked.
*(raised by czs007)*

---

- `tests/k8s/monitor/dashboards_test.go:44` — the dashboard tests never inspect `targets[].expr`: `TestK8sDashboards_Valid` (`:30-63`) only checks that the JSON unmarshals, that `__DATASOURCE_UID__` is gone, that `${prometheus}` is present, that `$cluster` is absent, and that four templating variable names exist; `TestK8sDashboardConfigMaps_InSync` (`:108-129`) only byte-compares template against configmap (verified in sync on this PR). Since `tests/k8s` is not in `make unit-test`'s exclusion list (which covers `tests/{integration,benchmark,stability,docker}`), this package is CI's only gate on dashboards, so unparseable PromQL and non-existent metric/method names pass green. Unrelated to this PR: the five changed files are the four dashboard JSONs plus the generated configmap — `dashboards_test.go` is untouched, and the same assertion set at base `c5adc7a` would have admitted a broken expression just as readily. Worth adding an assertion that no bare identifier (one without `=`/`!=`/`=~`/`!~`) appears inside an expression's braces, and ideally full PromQL parse validation plus a metric-name check against the `common/metrics` registry (a manual pass confirmed all 31 referenced metric names exist, so this is mechanizable). No tracking issue yet.

czs007:

Adversarial review of the latest round found one carried-over correctness issue in the new client dashboard panels and one unrelated file change bundled into this PR.

## High

**`tests/docker/monitor/grafana/templates/dashboard_client.json:1709` — C4/C6 apply `delta()` to entry-id gauges that reset to zero on every segment roll**

*Previously raised in an earlier round on this PR. The author has not responded to it in the PR discussion; the two follow-up commits (`39ad729e96`, `97ffa725ff`) only added the `log_ns=~"$log_ns"` matcher to these two expressions and left the `delta()` semantics untouched, so the original problem is unchanged.*

- **Problem:** C4 (`sum by (log_ns, log_id) (delta(woodpecker_client_write_frontier_entry[5m]))`) and C6 (`sum by (log_ns, log_id, reader_name) (delta(woodpecker_client_read_frontier_entry[10m]))`) take `delta()` over gauges that are not monotonic. `common/metrics/client_metrics.go:401-402` states outright: "Entry ids restart at zero in every segment, which is also why lag is measured in segments." `SetWriteFrontier` (`client_metrics.go:404-407`) and `SetReadFrontier` (`:423-426`) `Set` the in-segment entry id directly, and both gauges are labelled only `{log_ns, log_id}` (+ `reader_name`) with no `segment_id` dimension, so a single series drops from a high value back to ~0 at each roll. `delta()` performs no counter-reset correction on gauges, so the drop flows straight into the result.
- **Failure scenario:** any log with sustained writes rolls a segment at least every 10 minutes under the default `segmentRollingPolicy` (`config/woodpecker.yaml:31-33`: `maxSize: 256000000`, `maxInterval: 600`), so C6's 10m window almost always spans a roll and C4's 5m window spans one roughly half the time → a healthy writer is plotted as a large negative spike, or, when the roll lands mid-window, the drop cancels real growth and the value lands near 0, which is exactly the "flat frontier with a non-zero append rate is a stall" criterion the C4 description tells the on-call to act on. C6 describes itself as "the only signal" for a stuck reader, yet a genuinely stuck reader reads 0 and a normal segment roll reads negative — the two states are indistinguishable on this panel.
- **Suggestion:** use segment-level movement instead, e.g. `changes(woodpecker_client_write_frontier_segment[...])` (the `write_frontier_segment` / `read_frontier_segment` gauges already exist — `client_metrics.go:51-56` / `:89-94`), or combine segment and entry into a monotonic quantity before applying `delta()`. At minimum, wrap with `clamp_min(delta(...), 0)` to absorb the resets. The same expression appears in six places and all should be updated together: `dashboard_client.json:1709/1834`, `dashboard_client_k8s.json:1709/1834`, `dashboard-configmaps.yaml:6455/6580`.

*(raised by xiaocai2333, bigsheeper)*

## Low

**`deployments/operator/config/default/kustomization.yaml:1` — unrelated operator kustomization rewrite bundled into this monitoring PR**

- **Problem:** commit `97ffa725ff` also rewrites this file from 229 lines to 43 (7 added, 193 removed), deleting the kubebuilder webhook / cert-manager / network-policy / metrics-with-certs scaffolding comment blocks — including the `# +kubebuilder:scaffold:crdkustomizecainjectionns` and `# +kubebuilder:scaffold:crdkustomizecainjectionname` markers — and moving `apiVersion` / `kind` from the top of the file to the bottom, with `resources:` and `patches:` relocated after their comment blocks. The actual `resources` and `patches` entries are unchanged, the YAML is still valid and `kustomize build` output is unaffected, so this is not a correctness problem. It is simply unrelated to this dashboard PR, and that commit's message only discusses the `AddEntry`/`AddEntries` `grpc_method` label values, saying nothing about the 186 deleted lines.
- **Suggestion:** drop this file from the PR and submit it separately, so the operator's component-enablement documentation and kubebuilder injection anchors do not disappear silently inside a monitoring change.

*(raised by bigsheeper)*

czs007:
> **Must-fix issues introduced by this PR: 2**
> **Merge recommendation:** Mergeable after the 2 must-fix items below are fixed in this PR.

Two carried-over correctness defects in the new client dashboard panels remain unaddressed, and one unrelated file is still in the diff.

## High

**`tests/docker/monitor/grafana/templates/dashboard_client.json:1709` — C4 applies `delta()` to a gauge that resets every segment**
The panel plots `sum by (log_ns, log_id) (delta(woodpecker_client_write_frontier_entry{...}[5m]))`, but that gauge holds the entry id *within the current segment*: `common/metrics/client_metrics.go:57-62` defines it as "Highest entry id acknowledged in the write frontier segment per log", and `client_metrics.go:401-402` states outright that "Entry ids restart at zero in every segment". `SetWriteFrontier` (`client_metrics.go:404-407`) writes it as a bare gauge, and `delta()` performs no reset compensation. Failure: a healthy log whose writer rolls from segment 7 (entry ~40000) into segment 8 inside the 5m window → the panel plots ≈ -40000 alongside a non-zero append rate, which is precisely the stall signature the panel's own description tells the on-call to act on, and the next window understates real append volume. Suggestion: plot the frontier position itself, or fold `woodpecker_client_write_frontier_segment` (`client_metrics.go:51-56`, same label set) progression into the expression so a segment advance counts as forward movement. This was raised in the previous round; the author has not responded to it in the PR discussion, and the only incremental commit since (`53facaf96a`) changes 12 lines, all `le` matchers, leaving C4 untouched. *(raised by xiaocai2333, bigsheeper)*

**`tests/docker/monitor/grafana/templates/dashboard_client.json:1834` — C6 has the same `delta()` defect on the read frontier**
`sum by (log_ns, log_id, reader_name) (delta(woodpecker_client_read_frontier_entry{...}[10m]))` reads a gauge documented at `common/metrics/client_metrics.go:93-98` as the "Read position of each reader within read_frontier_segment", set as a plain gauge by `SetReadFrontier` (`client_metrics.go:423-426`) and subject to the same per-segment reset. Failure: a reader finishes segment 12 at entry 30000 and is delivering entry 800 of segment 13 when the window closes → C6 plots ≈ -29200 for that `reader_name`, and thereafter only the small in-segment offset; meanwhile a reader genuinely wedged at entry 0 of a fresh segment plots 0, identical to a healthy reader that just rolled. The panel description calls this "the only signal" for a silently-retrying stuck reader, so the defect negates the panel's entire purpose in both directions. Suggestion: incorporate `woodpecker_client_read_frontier_segment` (`client_metrics.go:87-92`, same labels including `reader_name`) movement rather than delta-ing the in-segment entry id. The same expression is replicated at `tests/k8s/monitor/grafana/templates/dashboard_client_k8s.json:1834` and in the generated `tests/k8s/monitor/manifests/dashboard-configmaps.yaml:6580`. Raised in the previous round; no response from the author so far. *(raised by xiaocai2333, bigsheeper)*

## Low

**`deployments/operator/config/default/kustomization.yaml:42` — unrelated file rewrite still in the diff (+7 / -193)**
Introduced by `97ffa725ff`, a commit whose message is entirely about gRPC method names, in a PR that is otherwise dashboards only. This was raised in the previous round and the author has not responded; re-raising it because the earlier characterisation of the change as "only a positional reordering … functionally equivalent" understates what the diff contains. Three distinct changes are present: (1) `resources:` / `patches:` keys move below the comment blocks that previously sat inside them — this part genuinely is an equivalent reordering, both lists keep identical entries; (2) roughly 180 lines of commented-out kubebuilder scaffolding (cert-manager CA-injection replacements, webhook patches) are deleted, including the two markers the file itself labels "Do not remove or uncomment the following scaffold marker; required to generate code for target CRD" — `# +kubebuilder:scaffold:crdkustomizecainjectionns` and `# +kubebuilder:scaffold:crdkustomizecainjectionname` — and `deployments/operator/PROJECT` confirms this is a live kubebuilder project where those markers are load-bearing for future scaffold runs; (3) `apiVersion: kustomize.config.k8s.io/v1beta1` and `kind: Kustomization` are appended at the end of the file, neither of which exists anywhere in the base revision `c5adc7a1ac`. Together that is the signature of a YAML load/dump round-trip performed by a tool rather than a hand edit, which seems unintended. `kustomize build` output is unchanged, so nothing breaks today. Suggestion: revert the file out of this PR rather than reasoning about whether the scaffold-marker loss is acceptable; if the kustomization cleanup is wanted, land it as its own change. *(raised by bigsheeper)*

tinswzy:
Thanks — all of it was actionable. Addressed in `c9a67d4`, plus three earlier commits that had already landed before the latest round was posted.

**C4 / C6, `delta()` on the entry gauges** — confirmed and fixed. `client_metrics.go:401` states the reset outright, and I reproduced the inversion on a live cluster: a log rolling 19 segments in five minutes read **−11.6** on the old expression. Both panels now count segments rolled plus entries advanced inside the current one, with `clamp_min` absorbing the reset:

```promql
sum by (log_ns, log_id) (
  changes(woodpecker_client_write_frontier_segment{...}[5m])
+ clamp_min(delta(woodpecker_client_write_frontier_entry{...}[5m]), 0)
)
```

Same log now reads **19**. I did not take the `changes(*_frontier_segment)`-only form, because a log writing steadily inside one segment would report no movement.

**R1 / R2, write-only against a shared ceiling** — correct, reads are now summed in. Worth noting for the record that I could not observe the difference: the test workload has no read traffic, so both forms read 0.054 there. Accepted on the argument, not on evidence.

**S5, starvation reading as balance** — correct, and a sharper failure than I had considered. The denominator is now the mean over the cluster's node count via `system_cpu_usage`, so a node with no child series drags the mean down instead of vanishing from it.

**Threshold steps at 0** — moved to 0.001 on R7, C7 and C8.

**`kustomization.yaml`** — entirely my fault and reverted. The monitor script runs `kustomize edit set image` / `set namespace` during bringup, which round-trips the YAML; a `git add -A` swept the result into a commit about gRPC method names. That explains the signature you spotted, including the appended `apiVersion`/`kind`. The PR now touches only `tests/`.

**On the first three must-fix items** (`{SEL,` placeholder, `AppendEntry` vs `AddEntry`, `log_ns`/`log_id` on node-scoped metrics) — those were already fixed in `39ad729` and `97ffa72` before the review posted, found the same way, by opening the dashboards against a live cluster rather than reading the JSON.

**On the pre-existing gap you flagged** — that dashboard tests never inspect `targets[].expr` — I agree, and `54e6596` in this PR adds it. `TestK8sDashboards_AlertPanelsQuery` runs every Alerts-row expression against live Prometheus: everything must be *accepted* (catches the malformed class), and a short allowlist of panels fed by metrics that cannot be idle under load must also return series (catches the parses-but-matches-nothing class). Error-counter panels are deliberately off that list, since empty is their healthy reading. Verified by reintroducing each of the three original bugs in turn; each fails its own subtest and nothing else.

czs007:

Verified:
- `tests/docker/monitor/grafana/templates/dashboard_client.json:1709` — the C4 write-frontier expression now reads `changes(write_frontier_segment[5m]) + clamp_min(delta(write_frontier_entry[5m]), 0)`, and both gauges carry identical `{log_ns, log_id}` labels (`common/metrics/client_metrics.go:51-62`) written as a pair by `SetWriteFrontier`, so the binary match is one-to-one and a segment roll is counted as progress instead of a large negative delta.
- `tests/docker/monitor/grafana/templates/dashboard_client.json:1834` — C6 uses the same construction aggregated by `(log_ns, log_id, reader_name)`; `read_frontier_segment`/`read_frontier_entry` both carry those three labels (`client_metrics.go:87-98`) and `SetReadFrontier`/`ClearReadFrontier` set and delete them together, so neither side of the join dangles and a single-series panel no longer reads negative on reset.
- `tests/docker/monitor/grafana/templates/dashboard_server.json:3725` — R1/R2 now include `disk_read_ops_total`/`disk_read_bytes_total`, which are emitted by the same collector with the same `{node_id, path, device}` label set (`system_disk_io.go:57-61`), so the utilization ratio is finally measured against the same read+write device ceiling it is compared to.
- `tests/docker/monitor/grafana/templates/dashboard_server.json:4506` — S5 divides by `count(count by (node_id) (system_cpu_usage))`; the `node_id` label is confirmed at `system_metrics_collector.go:34-39`, and both numerator and denominator reduce to label-free scalars, so a node that is never selected still contributes to the node count and extreme skew no longer reads as perfect balance.
- `manifests/dashboard-configmaps.yaml` — regenerated from the templates: zero occurrences of the old inverted expressions remain, and the docker template, `dashboard_client_k8s.json`, and the configmap agree expression-for-expression.
- `deployments/operator/config/default/kustomization.yaml` — fully reverted; the unrelated operator kustomization change no longer appears in the net diff for this PR.

czs007:

**Verified:**
- `tests/docker/monitor/grafana/templates/dashboard_client.json:1709` — C4 now reads `changes(write_frontier_segment[5m]) + clamp_min(delta(write_frontier_entry[5m]), 0)`; both halves carry the label set `{log_ns, log_id}` and are written together in `SetWriteFrontier`, so the `+` cannot silently drop series, and in-segment progress / roll / true stall all resolve correctly.
- `tests/docker/monitor/grafana/templates/dashboard_client.json:1834` — C6 is structurally identical over the read frontier; both halves share `{log_ns, log_id, reader_name}`, and `ClearReadFrontier` deletes them together, so no one-sided series can linger.
- `tests/docker/monitor/grafana/templates/dashboard_server.json:3725` — R1/R2 changed from write-only to `rate(read_*) + rate(write_*)`; the four disk metrics come from one collector with identical label set `{node_id, path, device}`, so the 1:1 vector match holds and does not yield an empty set.
- `tests/docker/monitor/grafana/templates/dashboard_server.json:4506` — the S5 denominator `count(count by (node_id) (woodpecker_server_system_cpu_usage{...}))` is backed by a gauge that does carry `node_id` with one series per node, so it counts present nodes rather than degenerating to 1.
- `tests/k8s/monitor/manifests/dashboard-configmaps.yaml:4516` — the generated configmaps are in sync with the k8s templates (16+/16− matching the 32 changed template lines), with each new expression present exactly once in the generated output.

## Nitpick — non-blocking

- `tests/docker/monitor/grafana/templates/dashboard_client.json:1709` (and `:1834` for C6) — `frontier_segment` and `frontier_entry` are two separately registered `GaugeVec`s (`common/metrics/client_metrics.go:54,60`) gathered independently, so a scrape can observe a position torn across a segment roll; the code comments at `client_metrics.go:400-402` acknowledge this. Failure: a scrape lands between the two `Set` calls in `SetWriteFrontier` (`client_metrics.go:405-406`) during a roll and records `segment=N+1` with the previous segment's large `entry`, and that torn sample happens to be the newest sample inside the `[5m]` (C4) or `[10m]` (C6) window at evaluation time → `clamp_min(delta(frontier_entry), 0)` returns roughly the previous segment's final entry id (potentially tens of thousands) instead of entries advanced in the current segment, drawing an isolated spike far above the real rate and polluting the legend's `max` column. To be precise about scope: `delta()` only looks at the first and last samples of the range, so a torn sample falling in the *middle* of the window has no effect; and the error is one-directional (it can only overstate progress, never render a stalled frontier as advancing), so C4/C6 remain sound as stall detectors. Suggestion: if you want to eliminate it, fold the two gauges into a single custom collector that emits both in one `Collect` call; otherwise a note in the panel description that this spike is ignorable is sufficient. The earlier `delta()`-semantics finding at these lines was addressed in `02a3c3086f`; this is a distinct residual, not a re-raise. (raised by bigsheeper)

tinswzy:
Thanks for the re-verification.

On the remaining nitpick — the torn read between `frontier_segment` and `frontier_entry` — I've taken the lighter of your two options and noted it in the C4/C6 panel descriptions (`37f3100`) rather than restructuring the metrics:

> An occasional isolated spike here is expected and can be ignored: the segment and entry gauges are registered separately, so a scrape landing between the two Set calls during a roll records the new segment beside the old segment's entry id. The error only ever overstates progress, never the reverse, so it cannot hide a stall.

Your scoping is what decided it: `delta()` only reads the first and last samples of the range, so a torn sample mid-window is inert, and the error is one-directional. Since these two panels exist to detect a stall, an artefact that can only ever look like *more* progress cannot produce a false negative — and a false positive here costs one glance at the append-rate series drawn beside it.

Folding the pair into a single collector would remove it properly, but that changes the metrics rather than a dashboard, and it would want its own PR alongside the `client_metrics.go:400-402` comment that already documents the tear.

PR #350, opened 2026-09-20
fix(monitor): give etcd and MinIO volumes, so log ids stay monotonic

tinswzy:
## The symptom

Rerunning `run_monitor_tests.sh` against a cluster kept with `--no-cleanup` fails **every** append:

```
compacted segment 1/0: tombstone present but no compacted footer
in object storage (unexpected state)
```

then `writer lock session has expired`, which reads like a locking problem and is not one — the lock expiring is the last symptom, not the cause.

## The cause

The dependencies and the servers did not have the same durability:

| | storage | after a minikube restart |
| --- | --- | --- |
| woodpecker servers | PVC | data intact |
| MinIO | none | every object gone |
| etcd | none | all metadata gone |

An empty etcd re-seeds `logs/idgen` to `0` (`meta/metadata_etcd.go:305`), and `CreateLog` takes the next id from that counter — `nextID := currentID + 1` (`:424`) — **not from the log name**, which is used only for the duplicate check (`:400`). So the next log gets id 1 again and writes onto segment directories the previous run left on the PVCs, where `files/1/0/data.compacted` says the segment is compacted while the footer it refers to died with MinIO.

The server then refuses to serve that segment. **That refusal is correct** — it is the guard at `writer_impl.go:378`, and it is what stands between a lost footer and silent data loss. Nothing in the server needs changing.

`meta/clear_meta.go` is explicit about why this must not happen:

> logId appears in the data paths of both storage tiers … Restarting the counter makes a new log reuse a directory that an old log may still have objects in … Keeping the counter monotonic makes collisions structurally impossible.

A harness whose etcd forgets is reproducing a state production does not permit.

## The fix

Volumes for etcd and MinIO, so all three tiers have the same lifecycle.

This also matters beyond this script: in the chaos suite, killing etcd should exercise **recovery**, not metadata loss. Without a volume, that test simulates a failure production cannot have.

## On the earlier version of this PR

It tried a timestamped log name instead. That was wrong — the name plays no part in id allocation — and it appeared to work only because the cluster I tested on still had a live etcd, so the counter had already advanced past 1. Thanks for catching both the mechanism and the reason the verification looked green; that was exactly right.

The timestamped name is kept, but only for what it actually does: avoiding the duplicate-name check on a rerun. Its comment now says so instead of claiming to solve id reuse.

## Verification

Reproducing the original trigger, on the same cluster:

| | delete the etcd pod, then create a log under a brand-new name |
| --- | --- |
| before | `logId=1`, **430 failed appends** |
| after | `logId=2` before the restart, `logId=3` after, **zero errors** either side |

A probe key written before the restart also reads back intact, and the key count and `logidgen` value are unchanged across it.

Two measurement notes, since the first round of this PR was verified badly: the "after" case deletes the etcd pod rather than testing against a surviving one, so it reproduces the condition that triggered the bug rather than a neighbouring one; and readiness is awaited with `until etcdctl endpoint health`, because reading too early returns an empty result that is easily mistaken for data loss.

## Nitpick from review

`printf %q` applied to `LOG_NAME` — the remote `bash -c` reparses the string, so a value containing a space would have been split there.


czs007:

Adversarial review of this PR found one correctness issue in the new timestamped-log-name logic plus one shell-quoting nit; a separate, pre-existing hazard that sits directly on the path this PR aims to fix is listed at the end as follow-up.

## High

**`tests/k8s/monitor/run_monitor_tests.sh:90`** — The timestamped log name cannot avoid logId reuse, and the new comment block (lines 83-89, echoed in the PR description) attributes the root cause incorrectly. `logId` is unrelated to `logName`: `CreateLog` reads the global counter `logs/idgen` (`meta/metadata_etcd.go:383`) and assigns `nextID := currentID + 1` to `LogMeta.LogId` (`:424-426`); `logName` is used only for the duplicate check in `CheckExists` (`:400`). When etcd is empty, `InitIfNecessary` seeds `idgen` back to `"0"` (`:305`), and `meta/clear_meta.go:44-51` states in-repo that `logId` appears in the data paths of both storage tiers, so reuse can only be prevented by keeping the counter monotonic. The etcd/MinIO pods sourced from `deployments/operator/test/lib.sh:96-127` have no volumes at all, so a minikube restart wipes them. Failure: keep the cluster with `--no-cleanup`, run `minikube stop && minikube start`, and rerun the script → the emptied etcd seeds `idgen` to 0, the new timestamped log still receives logId 1, writes land on the PVC's leftover `files/1/0/data.compacted` tombstone, the server rejects with `compacted segment 1/0: tombstone present but no compacted footer`, and loadgen fails every append until `writer lock session has expired` — identical to the pre-patch behavior, now with a comment in the tree asserting the opposite. The change's one real benefit is avoiding the `"log already exists"` rerun error while etcd survives (which also explains why manual verification succeeded: `idgen` had already been advanced to 1, so the new name got logId 2). Suggestion: give etcd/MinIO persistent volumes with the same lifecycle as the servers' PVC, or explicitly clear stale `{storage.root}` data on the PVC during bringup; at minimum, narrow the comment and PR description to the "etcd alive, PVC dirty" case rather than claiming id reuse is solved. (raised by bigsheeper, xiaocai2333)

## Nitpick

**`tests/k8s/monitor/run_monitor_tests.sh:100`** — `-log=$LOG_NAME` is expanded locally inside double quotes (no local word splitting), but the whole string is then reparsed by `bash -c` inside the pod, so spaces or glob characters in the value would undergo word splitting / globbing remotely and `-log=` would receive only the first fragment. Real-world impact is limited: the value comes from the caller's `LOG_NAME` env var and defaults to `date +%s`, which contains no special characters — hence nit only. Suggestion: use `-log=$(printf %q "$LOG_NAME")`, or single-quote the value on the remote side. (raised by bigsheeper, xiaocai2333)

- **`tests/k8s/monitor/loadgen/main.go:48`** — A metrics-port bind failure panics the entire loadgen process: `-metrics-addr` defaults to `:29099` (`:38`) and `http.ListenAndServe` panics on error inside a goroutine (`:47-51`), killing the process. The pod runs `sleep infinity` with `restartPolicy: Never` (`manifests/client-loadgen.yaml:15,21`), the write loop only prints `res.Err` without exiting (`main.go:106-108`) so the old process never terminates, `kubectl apply` on an unchanged spec is a no-op (`run_monitor_tests.sh:91`), and the exec never passes `-metrics-addr` (`:100`). Failure: after a previous `--no-cleanup` run leaves the loadgen process alive, rerunning the script without restarting minikube → the new process fails to bind `:29099` and panics before creating the client, so `-log=<timestamp>` never takes effect and `/metrics` is still served by the stale process, letting the assertions pass on the previous run's data. This PR touches only `run_monitor_tests.sh` (new lines 83-90, rewritten 99-101) and leaves `loadgen/main.go` and `client-loadgen.yaml` untouched — the panic reproduces without this patch. No tracking issue yet. Because this sits directly on the rerun path this PR targets, we recommend merging that separate fix before this PR. (raised by bigsheeper)

tinswzy:
You were right on both counts, and the second one is the part worth dwelling on.

**The mechanism.** `CreateLog` takes the next id from `logs/idgen` (`meta/metadata_etcd.go:383`, `:424`); the name is only the duplicate check at `:400`. A timestamped name cannot affect id allocation. The comment I added asserting otherwise was wrong and is gone.

**Why my verification looked green.** Exactly as you deduced — the cluster I tested on still had a live etcd, so the counter had already advanced past 1 and the new name happened to draw id 2. I verified the "etcd alive, PVC dirty" case and shipped it as a fix for the "etcd wiped" case. That is the more useful half of this review: not that the code was wrong, but that my evidence was about a different scenario than my claim.

I reproduced the real trigger afterwards to be sure: delete the etcd pod, create a log under a brand-new name → **`logId=1`, 430 failed appends**. Unchanged from before the patch.

**The fix now.** Volumes for etcd and MinIO, so all three tiers share a lifecycle and the counter stays monotonic — which is what `meta/clear_meta.go` requires, and the invariant my harness was violating. Same sequence after the change: `logId=2` before the restart, `logId=3` after, zero errors either side, and a probe key written beforehand reads back intact.

This also fixes something your review pointed at indirectly: in the chaos suite, killing etcd should exercise recovery. Without a volume it was exercising metadata loss, which production cannot have.

The timestamped name is kept for what it actually does — avoiding the duplicate-name check on a rerun — and its comment now says only that.

`printf %q` applied as suggested.

**On the pre-existing loadgen hazard** (metrics-port bind panic leaving a stale process serving `/metrics`) — agreed it sits directly on this path, and agreed it is separate. It also explains a second way a rerun can look green while measuring the previous run's data. I have not filed it yet; happy to, unless you would rather it go with whoever picks up `loadgen/main.go`.

czs007:

Overall: the persistent-volume approach in `lib.sh` fixes the id-reuse root cause correctly, but the existing early-return guard in `wp_deploy_deps` prevents that fix from applying in the exact scenarios it targets, and makes a partial-pod state fail the whole script.

## Medium

- **`deployments/operator/test/lib.sh:87`** — The new `etcd-data` / `minio-data` PVCs and the `volumes` / `volumeMounts` additions all live inside the heredoc at `lib.sh:100-171`, which sits behind the pre-existing early return `if kubectl get pod/etcd pod/minio &>/dev/null; then log ...; return 0; fi`. As long as both pod objects exist, the entire apply — including the new PVCs — is skipped and the log just prints "etcd and MinIO already running, skipping". That guard was harmless before this PR (it skipped an apply identical to the live state); with this PR it silently disables the fix. Failure: rerunning the script against a cluster kept by `--no-cleanup` (`run_monitor_tests.sh:34/120`) or `KEEP` (`run_network_chaos.sh:204`), or after `minikube stop && minikube start`, still finds both pods → `wp_deploy_deps` returns at `lib.sh:89`, the PVCs are never created, etcd keeps running volume-less, `logs/idgen` re-seeds to 0 after the container is rebuilt, the new log gets `logId=1` and collides with the leftover `files/1/0/data.compacted` on the server PVC, reproducing `compacted segment 1/0: tombstone present but no compacted footer` while Step 4 reports everything as fine. Note this covers both scenarios the PR targets, including the `minikube restart` path described by the new comment at `lib.sh:92-99`: after a restart the pod objects still exist, so the guard still fires. Suggestion: detect whether the existing pods already carry the two volumes and `kubectl delete pod/etcd pod/minio` before reapplying when they do not, rather than simply removing the guard (see the next item for why removing it is not sufficient), and document in the comment that pre-existing clusters need their dependency pods recreated to benefit. *(raised by bigsheeper)*

- **`deployments/operator/test/lib.sh:100`** — The guard at `lib.sh:87` only short-circuits when *both* pods exist; if exactly one is missing it returns non-zero and execution falls through to the `kubectl apply`, which then submits a spec with new `volumes` / `volumeMounts` (and, for etcd, the changed `command` with `--data-dir=/etcd-data`) against the surviving old pod. Pod specs are immutable apart from a few fields such as `image`, so the API server rejects it with `pod updates may not change fields other than ...` and `kubectl apply` exits non-zero. Failure: on a pre-PR cluster whose dependency pods have no volumes, deleting or losing just one of them (e.g. `kubectl delete pod etcd` as in the PR's own Verification section, or minio being evicted) and then rerunning `run_monitor_tests.sh` / `run_network_chaos.sh` / `smoke-test.sh step4` → the apply is rejected and, since all three callers use `set -euo pipefail` and invoke `wp_deploy_deps` bare (`run_monitor_tests.sh:63`, `smoke-test.sh:106`, `run_network_chaos.sh:49`), the script aborts at Step 4 leaving the cluster half-migrated: PVCs created, one pod with volumes, the other still without. Before this PR the same path was safe, because the applied spec matched the live one and kubectl printed `pod/minio unchanged`. Suggestion: run `kubectl delete pod/etcd pod/minio --ignore-not-found` (or a volume-presence check) before the apply so dependency-pod recreation is deterministic. *(surfaced during verification)*

## Low

- **`tests/k8s/monitor/run_monitor_tests.sh:83`** — In the comment rewritten this round (the earlier root-cause attribution was correctly replaced by the persistent-volume fix; only this clause remains inaccurate), the timestamped log name is justified as avoiding the duplicate-name check in `CreateLog`. But `tests/k8s/monitor/loadgen/main.go:68` is `_ = cli.CreateLog(ctx, *logName) // idempotent`: the `ErrMetadataCreateLogAlreadyExists` returned at `meta/metadata_etcd.go:408-412` is discarded and the following `OpenLog` opens the existing log, so a duplicate name was never an obstacle — it only leaves a `log already exists` warning server-side. What the timestamped name actually buys is a distinct `logId` per run, keeping the per-run client metric series (`log_ns` / `log_id` labels, `common/metrics/client_metrics.go:50` onward) from being mixed together. Suggestion: reword the comment to state that purpose. *(raised by bigsheeper)*

PR #353, opened 2026-09-20
enhance(admin): add GET /admin/instance/data so orphaned local data can be found

### LINKED ISSUE #349: enhance(admin): local-disk residue is write-only — the control plane can delete an instance's data but can never ask whether it is gone
### Problem

In `service` storage mode, an instance's node-local WAL data (`<root>/<bucket>/<rootPath>/<logId>/<segId>/data.log`) is reclaimed only when an external control plane calls `POST /admin/instance/delete` on every node that holds it. That broadcast is best-effort, and in practice it is lossy:

- **A node is down or partitioned when the delete is issued.** On restart, `rebuildDeletingSetsFromMarkers` repopulates the deleting-set from on-disk markers only. The node never received the delete, so it has no marker, finds nothing to reclaim, and keeps the data indefinitely.
- **The control plane cannot wait forever.** It has to move an instance to a terminal state on a deadline, whether or not every woodpecker node acknowledged.
- **The cleanup step is skipped entirely.** When an operator force-sets a Milvus instance's state while handling some other incident, the woodpecker deletion never runs at all.

Each of these strands local data that nothing will ever reclaim. Disk is consumed by an instance that no longer exists, and neither woodpecker nor the control plane can notice.

The underlying gap is that **deletion is write-only**: the control plane can act, but it cannot observe. It has no way to ask "does instance X still have data on this node?", so it cannot run a query -> decide -> reclaim reconciliation loop, and it has no path back to a correct state on its own.

### Why the existing endpoints don't answer this

- `GET /admin/node/decommission/progress` exposes a single node-wide `has_local_data` boolean. It cannot be attributed to an instance, and it deliberately ignores segments that already carry a compacted mark.
- `GET /admin/log-health` reports **in-memory observed** read/write health for logs this node has served. A node holding pure residue runs no processor, observes nothing, and reports nothing.
- Metrics have the same blind spot: `woodpecker_server_logstore_active_logs{namespace="bucket/root"}` is driven by live processors, not by what is on disk.
- `GET /admin/logstore/segments` lists in-memory writer snapshots, not disk contents.

None of them distinguishes "this instance is clean" from "this instance has 12 logs of orphaned data sitting here with nothing running".

### Proposal

One read-only, **node-local** endpoint, paired with the delete that already exists under the same noun:

```
GET  /admin/instance/data     ?bucket_name=<b>&root_path=<r>
POST /admin/instance/delete   {"bucketName":"<b>","rootPath":"<r>"}     (already exists)
```

| Question | Call |
|---|---|
| What instances hold data on this node? | no params |
| Does instance `(b, r)` still have data on this node? | `?bucket_name=b&root_path=r` |
| What does the whole cluster hold? | the caller fans out — see below |

Filter semantics follow `/admin/log-health` exactly, since it is the only other endpoint taking these two params: both must be supplied or the filter is ignored, and `bucket_name` / `bucketName` / `bucketname` are all accepted (`firstNonEmpty`). The response echoes the filter that actually took effect as `filter_bucket_name` / `filter_root_path`, so a caller that misspells a param can see it received an unfiltered answer rather than silently mistaking it for one instance's data.

### Response

```json
{
  "node_id": "woodpecker-0",
  "storage_mode": "service",
  "storage_root": "/var/lib/woodpecker",
  "timestamp_ms": 1758300000123,
  "instance_count": 2,
  "total_size_bytes": 4823910,
  "scan_errors": 0,
  "instances": [
    {
      "bucket_name": "a-bucket",
      "root_path": "in01-71ba8da6ff40c33",
      "log_count": 12,
      "segment_count": 37,
      "live_segment_count": 30,
      "size_bytes": 1234567,
      "last_modified_ms": 1758299000456,
      "active_processors": 4,
      "delete_state": "",
      "marked_deleted_at_ms": 0
    },
    {
      "bucket_name": "a-bucket",
      "root_path": "in01-dead0beef000000",
      "log_count": 4,
      "segment_count": 9,
      "live_segment_count": 9,
      "size_bytes": 3589343,
      "last_modified_ms": 1755100000000,
      "active_processors": 0,
      "delete_state": "",
      "marked_deleted_at_ms": 0
    }
  ]
}
```

`filter_bucket_name` / `filter_root_path` are `omitempty` and appear only when a filter took effect. `200` always, `405` for any method other than GET. There is no unhealthy state to report, so no `503`.

Every field either drives a control-plane decision or explains a discrepancy the caller will otherwise hit:

- `scan_errors` — directories that could not be read. **A non-zero value means "not found" does not imply "not present"**, and the caller must not delete on that basis.
- `active_processors`, `last_modified_ms` — the anti-footgun pair. A freshly created instance whose registration has not yet propagated and a genuine orphan are indistinguishable on disk; only live processors and data age separate them. Callers should require `active_processors == 0` plus an age threshold before acting. In the example above, the second instance is the orphan.
- `delete_state` (`""` | `"log"` | `"instance"`), `marked_deleted_at_ms` — already marked and inside the grace window. Do not re-send; do not read as a failed delete.
- `live_segment_count` vs `segment_count` — the former is the `HasLocalSegmentData` predicate's view (segments not yet carrying a compacted mark). Both are reported so that an operator seeing `segment_count: 37` alongside `has_local_data: false` on the decommission endpoint can see why they disagree, rather than filing a bug.
- `storage_mode`, `storage_root` — in `minio` mode there is no node-local data at all, so an empty list means "not applicable", not "clean".
- `size_bytes`, `log_count`, `total_size_bytes` — what the residue actually costs.

### Conventions this follows

Deliberately modelled on `/admin/log-health`, the existing node-level, filterable, list-bearing read: `node_id` + echoed `filter_*` + `timestamp_ms` + node-level config echo (`storage_mode` / `storage_root`, where log-health has `stall_after_ms`) + aggregate counters + the detail list. Time fields are epoch milliseconds with an `_ms` suffix (`timestamp_ms`, `last_modified_ms`, `marked_deleted_at_ms`), matching `started_at_ms`, `last_seen_ms`, `last_modified_ms` elsewhere. The list is wrapped in an object rather than returned bare. Read via GET with snake_case query params; no path params, consistent with every other admin endpoint.

### Cluster-wide view stays with the caller

Every admin endpoint in this codebase answers only for its own node — there is no HTTP client anywhere in `common/http`, `server/`, or `cmd/main.go` that calls a peer. Cluster aggregation lives in wpcli, which resolves the gossip memberlist and fans out client-side (`wp cluster info`, `wp cluster health`, `wp node list`, `wp config diff --all`). This endpoint keeps that split.

A server-side `?scope=cluster` was considered and rejected. It would have been the first node->node admin call in the tree, it invites an N^2 fan-out if a caller queries every pod for the cluster view, and it buys the control plane nothing it does not already need: **there is no cluster-scoped delete, so the control plane must already enumerate nodes and handle an unreachable one in order to send the deletes at all.** A cluster-scoped read paired with a node-scoped write would also be asymmetric in exactly the place where care is required.

**The consequence must be stated plainly, because it is a safety rule and not an implementation detail: the completeness check belongs to the caller.** Before deleting anything, the caller must confirm that *every* node answered — an unreachable node holds intact data, and acting on a partial snapshot is precisely the accident this feature exists to prevent. Node enumeration comes from the caller's own pod inventory or from `GET /admin/memberlist`; wpcli's `--strict` / `StrictFailure()` is the reference implementation of the semantic.

### The loop this enables

1. Enumerate nodes, `GET /admin/instance/data` on each. If any node failed to answer, or any answer has `scan_errors > 0`, alert and take no destructive action this round.
2. Union the results, diff against the live-instance registry to get orphan candidates, then filter by `active_processors == 0 && now - last_modified_ms > threshold`.
3. For each surviving orphan, `POST /admin/instance/delete` to exactly the nodes that reported it.
4. Re-query with `?bucket_name=&root_path=` until the instance is absent everywhere.

Between 3 and 4 the data persists until the grace-based reclaim task runs. `EvictInstance` has no `sync` option (only the log-level delete does), so instance reclamation is always asynchronous. With current defaults (`deleteGracePeriod=5s`, `deleteReclaimInterval=2s`) convergence takes seconds; step 4 should poll rather than expect immediate absence.

### Scope

- `server/local_data_inventory.go` (new): one `WalkDir` over the storage root, reusing the existing top-level `.deleted` skip rule and the `hasCompactedMark` predicate. `HasLocalSegmentData` is left alone — its early exit is a requirement of the decommission polling path — but the "is this `data.log` live?" predicate is extracted and shared so the two can never drift.
- `server/logstore.go`: `LocalDataInventory()` on the `LogStore` interface, merging the disk scan with the `spMu`-guarded `segmentProcessors`, `deletingLogs`, and `deletingInstances` state.
- `common/http/management/instance_data_handler.go` (new), the path constant in `router.go`, the callback field in `AdminCallbacks`, conditional registration in `server.go`, and the closure in `cmd/main.go` — the same five-point shape every other admin endpoint uses.
- No request coalescing and no TTL cache: every request walks, and a reply is never stale.
  `HasLocalSegmentData` already walks the same tree unprotected on every
  `/admin/node/decommission/progress` poll — i.e. every 2s during a drain, far more often
  than a reconciliation loop will ever call this — so adding coalescing here alone would be
  inconsistent, and it has no honest test that does not require a test-only seam in
  production code. Revisit together with that walk if either becomes a measured problem.
- `wp instance data [node] [--all]` in wpcli, fanning out via the shared `NewFanout` + `--strict` path like `wp cluster info`. The control plane does not need it — it speaks HTTP — but an operator holding a PVC that will not drain does, and `--strict` is where the completeness check stops being improvised per incident. Paired with a cookbook recipe: a command operators cannot find is barely better than no command.
- `common/http/README.md`; unit tests for the scan predicates, the filter/echo behaviour, and the handler; an integration test covering query -> delete -> re-query convergence, including the case of a node that missed the original delete.

### Non-goals

- **No cluster-scoped read or delete on the server.** Both stay node-local; see above.
- **No object-storage inventory.** Object cleanup is client-driven off the log's object prefix (`woodpecker/log_object_cleanup.go`); this endpoint covers node-local disk only.
- **No `sync` option on `/admin/instance/delete`.** Noted as a possible follow-up; the polling loop in step 4 works without it.
- **No auth changes.** The endpoint exposes the same tenant identifiers already visible via `/admin/log-health` and `/metrics` on the same port.
- **No log-level drill-down.** The decision unit is the instance; `/admin/logstore/segments` and wpcli already serve per-log debugging.
- **Not fixing** the pre-existing camelCase-body / snake_case-query split between the delete and logstore handler families, nor `/admin/config`'s missing json tags.

tinswzy:
Closes #349.

### What this adds

`GET /admin/instance/data` — the read half of the `/admin/instance` family whose write half (`POST /admin/instance/delete`) already exists.

```
GET /admin/instance/data ?bucket_name=<b>&root_path=<r>
```

| Question | Call |
|---|---|
| What instances hold data on this node? | no params |
| Does instance `(b, r)` still have data here? | `?bucket_name=b&root_path=r` |

```json
{
  "node_id": "woodpecker-0",
  "storage_mode": "service",
  "storage_root": "/var/lib/woodpecker",
  "timestamp_ms": 1758300000123,
  "instance_count": 1,
  "total_size_bytes": 1234567,
  "scan_errors": 0,
  "instances": [
    {"bucket_name": "a-bucket", "root_path": "in01-abc",
     "log_count": 12, "segment_count": 37, "live_segment_count": 30,
     "size_bytes": 1234567, "last_modified_ms": 1758299000456,
     "active_processors": 0, "delete_state": "", "marked_deleted_at_ms": 0}
  ]
}
```

Plus `wp instance data [node] [--all]`, fanning out through the shared `NewFanout`/`--strict` path, and cookbook recipe 12 so operators can find it. The control plane does not need the CLI — it speaks HTTP — but an operator holding a PVC that will not drain does, and `--strict` is where the completeness check stops being improvised per incident.

---

## How to use

Two interfaces over the same node-local answer: the REST endpoint, which the control plane calls, and `wp instance data`, which fans the same call across the cluster for an operator. Every example below is real output, captured against a stub server rather than written from memory.

### REST: `GET /admin/instance/data`

Served on the admin port (default `9091`, `METRICS_PORT`). Always `200`; `405` for any method other than GET.

```bash
# 1. Everything this node holds
curl -s http://woodpecker-0:9091/admin/instance/data

# 2. Does one specific instance still have data here?
curl -s "http://woodpecker-0:9091/admin/instance/data?bucket_name=a-bucket&root_path=in01-abc"
```

`bucket_name` / `root_path` must be given **together or not at all** — a partial filter is ignored and you get every instance back (same rule as `/admin/log-health`; `bucketName` and `bucketname` spellings are also accepted). The reply echoes the filter that actually took effect as `filter_bucket_name` / `filter_root_path`, which is how a caller that misspelled a param can tell it received an unfiltered answer instead of one instance's data.

```json
{
  "node_id": "woodpecker-0",
  "storage_mode": "service",
  "storage_root": "/var/lib/woodpecker",
  "timestamp_ms": 1758300000123,
  "instance_count": 2,
  "total_size_bytes": 4823910,
  "scan_errors": 0,
  "instances": [
    {"bucket_name": "a-bucket", "root_path": "in01-71ba8da6ff40c33",
     "log_count": 12, "segment_count": 37, "live_segment_count": 30,
     "size_bytes": 1234567, "last_modified_ms": 1758299000456,
     "active_processors": 4, "delete_state": "", "marked_deleted_at_ms": 0},
    {"bucket_name": "a-bucket", "root_path": "in01-dead0beef000000",
     "log_count": 4, "segment_count": 9, "live_segment_count": 9,
     "size_bytes": 3589343, "last_modified_ms": 1755100000000,
     "active_processors": 0, "delete_state": "", "marked_deleted_at_ms": 0}
  ]
}
```

The second row is the orphan: nothing running against it, and the data has not been touched in weeks.

Three reads that decide whether you may act:

| Field | Read it as |
|---|---|
| `scan_errors > 0` | Some directories could not be read. An instance **missing** from `instances` is not proof it has no data here. Do not delete on this reply. |
| `active_processors`, `last_modified_ms` | A brand-new instance and a genuine orphan are identical on disk. Require `active_processors == 0` **and** an age threshold before treating a row as stranded. |
| `delete_state` (`""` / `"log"` / `"instance"`), `marked_deleted_at_ms` | Non-empty means the delete already landed and the grace window is running. Do not re-send it; do not read it as a failure. |

`segment_count` vs `live_segment_count`: the latter is the predicate `/admin/node/decommission/progress` uses for `has_local_data`, which ignores segments already durably compacted to object storage. A row where they differ is exactly why those two endpoints can disagree. In `minio` storage mode there is no node-local data at all, so an empty list means "not applicable" — `storage_mode` is in the reply so you can tell that from "clean".

### CLI: `wp instance data`

```bash
wp instance data node-1                                  # one node
wp instance data --all                                   # every node in the memberlist
wp instance data --all --strict                          # ...and fail if any node did not answer
wp instance data --all --bucket a-bucket --root in01-abc # one instance, cluster-wide
wp instance data --all -o json                           # machine-readable, node-addressable
```

```
NODE          BUCKET    ROOT_PATH             LOGS  SEGMENTS  LIVE  SIZE_BYTES  PROCS  DELETE_STATE
woodpecker-0  a-bucket  in01-71ba8da6ff40c33  12    37        30    1234567     4
woodpecker-0  a-bucket  in01-dead0beef000000  4     9         9     3589343     0
woodpecker-1  a-bucket  in01-71ba8da6ff40c33  12    37        30    1234567     4
woodpecker-1  a-bucket  in01-dead0beef000000  4     9         9     3589343     0
```

**`--strict` is a safety flag, not a formatting one.** Without it an unreachable node is just a row, and the command still exits `0`:

```
NODE          BUCKET       ROOT_PATH             LOGS  SEGMENTS  LIVE  SIZE_BYTES  PROCS  DELETE_STATE
woodpecker-0  a-bucket     in01-71ba8da6ff40c33  12    37        30    1234567     4
woodpecker-0  a-bucket     in01-dead0beef000000  4     9         9     3589343     0
woodpecker-1  UNREACHABLE

WARNING: 1/2 nodes unreachable — their instances are not listed; do not delete on this view
```

With it, the same situation is an error and exit code `6` (`StrictPartial`):

```
wp: StrictPartial: 1/2 nodes unreachable in strict mode
```

That node still holds intact data. Deleting on a partial view is the exact accident this endpoint exists to prevent, so any automation should use `--strict` and stop on a non-zero exit.

### Putting it together: reclaiming an orphan

The endpoint is node-local, like every other admin endpoint here, and there is no cluster-scoped delete — so the caller enumerates nodes for both halves.

```bash
# 1. Query. Stop here if any node failed to answer or reported scan errors.
wp instance data --all --strict

# 2. Decide: diff against your live-instance registry, then keep only rows with
#    PROCS == 0 and data old enough to rule out a just-created instance.

# 3. Delete — per node, only the nodes step 1 listed for that instance.
#    Deliberately not a wp subcommand: deletion should be explicit and targeted.
curl -X POST http://woodpecker-0:9091/admin/instance/delete \
  -H 'Content-Type: application/json' \
  -d '{"bucketName":"a-bucket","rootPath":"in01-dead0beef000000"}'

# 4. Poll until it is absent everywhere. Reclamation is asynchronous — there is no
#    `sync` option on the instance delete — so expect seconds, not immediacy.
wp instance data --all --strict --bucket a-bucket --root in01-dead0beef000000
```

For a control plane doing this unattended, step 1 is `GET /admin/instance/data` against every pod (from its own inventory or `GET /admin/memberlist`), and **the completeness check that `--strict` performs is the caller's responsibility** — it is the one part that must not be improvised.

### Where this is documented

| Doc | Covers |
|---|---|
| `common/http/README.md` | Full endpoint reference — request, response, every field, the reconciliation loop |
| `docs/admin-guides/wpcli/cookbook.md` recipe 12 | Operator runbook: how the stranded case arises, reading the columns, the delete, the poll |
| `docs/admin-guides/disk-watermark.md` | Check stranded data before expanding a PVC — none of the four remedies there frees it |
| `server/README.md` | Cross-reference from `has_local_data` on the decommission-progress reply |

---

### Design notes for review

**Node-local by design.** There is no `?scope=cluster`. Every admin endpoint in this tree answers only for its own node — there is no HTTP client in `common/http`, `server/`, or `cmd/main.go` that calls a peer — and cluster aggregation lives in wpcli (`wp cluster info`, `wp node list`, `wp config diff --all`). A server-side fan-out would have been the first node→node admin call here, invites an N² fan-out if a caller queries every pod, and buys the caller nothing it does not already need: **there is no cluster-scoped delete, so it must already enumerate nodes and handle an unreachable one to send the deletes at all.** The consequence is stated in the README rather than left implicit: the completeness check that gates any delete belongs to the caller.

**Filter semantics match `/admin/log-health`** — both params or the filter is ignored, and `bucket_name`/`bucketName`/`bucketname` are all accepted. I considered returning `400` on a partial filter instead, but two sibling endpoints taking the same two params with different rules is the worse trade; the echoed `filter_*` fields are what make the widening visible. If we want `400`, it should change both endpoints in a separate PR.

**Identity comes from the path tail** (`<logId>/<segId>/data.log`), not a fixed depth, so a `rootPath` containing separators stays one instance instead of being split.

**`HasLocalSegmentData` now shares the "is this data.log live?" predicate** with the inventory, so the decommission gate and the inventory an operator reads cannot drift apart. Its early exit is untouched — that is a requirement of the drain polling path.

**One deviation from the issue:** I dropped the planned singleflight around the walk, and [updated #349](https://github.com/zilliztech/woodpecker/issues/349) to say so. `HasLocalSegmentData` already walks the same tree unprotected on every `decommission/progress` poll — every 2s during a drain, far more often than a reconciliation loop will call this — so coalescing here alone would be inconsistent, and it has no honest test that doesn't require a test-only seam in production code. Worth revisiting together with that walk if either is ever measured to be a problem.

### Tests

| Area | Coverage |
|---|---|
| Scan | service/local/minio layouts, compacted-marked and zero-byte data.log, top-level `.deleted` skip, non-segment trees, multi-component rootPath, unreadable dir → `scan_errors` |
| Inventory | processor + marker merge, instance-mark outranks log-mark, oldest log mark wins, filter + echo, `[]` not `null`, deterministic ordering |
| Contract | exact JSON key sets for both structs, `filter_*` omitted when unfiltered |
| Handler | filter pass-through, partial filter ignored, alternate spellings, 405, JSON encoding |
| Wiring | `AdminCallbacks.GetInstanceData` → registered route → response |
| Convergence | query → delete → reclaim → re-query, and a node that never got the delete still reporting its stranded data |
| wpcli | single node, `--all` fan-out, filter → query params, scan errors surfaced, unreachable node visible, `--strict` fails |

Every one of these was mutation-verified: the corresponding production line was broken and the test confirmed to fail, so none of them is passing by accident.

`go build ./...`, `go vet`, `golangci-lint` (0 issues) and `go test -race -short` on `./server/ ./common/http/... ./cmd/wpcli/...` all pass. `make fmt` reformats ~35 unrelated files in this tree; that churn is deliberately excluded from this diff.


czs007:

Adversarial review of this PR surfaced five issues in the new local data inventory endpoint and its CLI wrapper: three correctness gaps that can mislead the orphan-reclamation flow described in `common/http/README.md`, plus two lower-severity reliability/efficiency concerns.

## Medium

**`server/local_data_inventory.go:96` — instance key mismatch between the scanner and the processor/marker annotation paths in `local` storage mode.**
In `local` mode the on-disk layout has no bucket level (`server/maintenance.go:125-126`), so `scanLocalInstances` takes the `bucketDepth = 0` branch, leaving `bucket` empty and producing the instance key `"/<rootPath>"` (`:142-144`, `:156`) with an empty `entry.BucketName` (`:159`). Both annotation paths use a non-empty bucket instead: `annotateActiveProcessors` (`:253-264`) keys off `GetLogKey(bucketName, rootPath, logId)` (`server/logstore.go:343-345`) where `bucketName` is the client-supplied value (default `a-bucket`, `config/woodpecker.yaml:298`), and `annotateDeleteMarkers` (`:280`) uses `GetInstanceKey(m.Bucket, m.RootPath)` — neither can ever match `"/<rootPath>"`. Failure: a node running with `woodpecker.storage.type: local` and a client actively writing under `bucketName: a-bucket` / `rootPath: files` → the live instance is reported with `active_processors=0` and `delete_state=""`, matching the README deletion criteria so the control plane issues `POST /admin/instance/delete` against live WAL data, while the follow-up `?bucket_name=a-bucket&root_path=files` query always returns `instances: []` (the filter at `:233-234` never matches an empty `BucketName`), making step 4 of the README loop immediately conclude the instance is gone. This is not dead code — `server/service.go:138` constructs the Server in local mode too, and the PR adds `TestScanLocalInstances_LocalModeHasNoBucketLevel` for this branch. Suggestion: unify the two sides — either populate `BucketName`/key from the configured bucket when scanning in local mode, or strip the bucket from the processor/marker keys in local mode. *(raised by xiaocai2333, bigsheeper)*

**`cmd/wpcli/cmd/instance.go:100` — `--strict` does not cover `scan_errors`, contradicting the command help and the README.**
The command's Long text (`:58-60`) lists UNREACHABLE and "any node reporting scan errors" together before saying "Use --strict to turn that into a non-zero exit", and `common/http/README.md:173-176` treats `scan_errors > 0` as a stop condition with `wp instance data --all --strict` offered at `:184-186` as the equivalent one-liner. In practice `fanRes.StrictFailure()` (`cmd/wpcli/client/fanout.go:57-59`) only checks `r.strict && r.Unreachable > 0`, and it runs before the body is parsed; `scan_errors` is only read at `:122-124`, accumulated into `scanErrorNodes`, and reduced to a WARNING at `:149-153` before `return nil`. JSON/YAML output returns even earlier at `:104-106` without the warning. Failure: all nodes reachable but one instance directory unreadable due to a permission/IO error (`server/local_data_inventory.go:110`/`:152`, response carries `scan_errors=3`) → `wp instance data --all --strict` exits 0, the operator's reclamation script treats the inventory as complete and proceeds to `POST /admin/instance/delete` on a partial view — exactly the accident this endpoint exists to prevent. Test coverage has the same gap: `TestInstanceData_ScanErrorsAreSurfaced` omits `--strict`, and `TestInstanceData_StrictFailsOnUnreachableNode` only exercises unreachable nodes. Suggestion: either fold `scan_errors` into the strict decision locally in this command (the global `--strict` is pure reachability semantics elsewhere, so keep the tightening scoped), or narrow the help text and README wording to promise reachability only. *(raised by bigsheeper)*

**`server/local_data_inventory.go:121` — the inventory only counts `data.log`, so tombstone trees left by compacted-file cleanup are invisible to the endpoint.**
`scanLocalInstances` treats only files named `data.log` as instance data (`:121-123`, `segmentDataFileName` at `:30`), but the compacted-file-cleanup task deliberately deletes `data.log` and keeps the mark and directory as a tombstone (`server/compacted_file_cleanup.go:106`, `:413`), with the mark intentionally retained long-term (`server/compacted_mark.go:32-34`). Failure: a long-idle, decommissioned instance on a service-mode node whose segments have all been compacted and had their `data.log` removed, leaving only `data.compacted` per segment directory → the instance is absent from `instances` with `scan_errors=0`, which per `common/http/README.md:158-160` and `:173-176` is proof that the node holds no data for it, so `POST /admin/instance/delete` is never issued and the tombstone tree (one directory plus one mark file per segment) accumulates on disk indefinitely with unbounded inode growth. Note this bites hardest at the intended target: README step 2 filters candidates by `active_processors == 0` plus a `last_modified_ms` age threshold (`:177-179`), and "old enough" is precisely the state in which every segment has already been compacted. The divergence from `HasLocalSegmentData` is reasonable there (the retirement gate only cares about undrained data) but not here, where the question is what remains on disk to reclaim. Suggestion: count segment directories carrying a compacted mark as well, possibly under a separate field such as `compacted_only_segment_count`, so such instances at least appear in the list. *(surfaced during verification)*

## Low

**`server/local_data_inventory.go:110` — concurrent background reclamation inflates `scan_errors` with predictable false positives.**
The walk callback only tolerates ENOENT when `p == root` (`:106-112`); any mid-walk disappearance on a non-root path increments `scanErrors`, as does an ENOENT from the deferred `d.Info()` call (`:150-154`). Both are routine in this repo rather than exceptional: `compactedFileCleanupTask` removes `<segDir>/data.log` one by one (`server/compacted_file_cleanup.go:413`) and `deletedLogReclaimTask` does `os.RemoveAll` over whole instance/log trees (`server/maintenance.go:213`), and both are periodic tasks that naturally overlap the polling required by README step 4. Failure: the control plane polls `GET /admin/instance/data` while compacted-file-cleanup or deleted-log-reclaim is running → the walk hits a just-deleted `data.log` or directory, `scan_errors` goes intermittently non-zero, and per `common/http/README.md:174` that means "stop", so reclamation stalls at random on a perfectly healthy node with no way for operators to distinguish this noise from a genuine permission/IO blind spot. The fail-safe direction is right, but merging "a file our own GC is deleting" with "a directory we cannot read" into one counter makes the signal unusable on busy nodes. Suggestion: pass through `os.IsNotExist(walkErr)` for non-root paths and ENOENT from `d.Info()` the same way the root case already does, counting only real permission/IO errors. *(raised by xiaocai2333, bigsheeper)*

**`server/local_data_inventory.go:227` — filtered queries still walk the entire storage root; the filter is applied only after the scan.**
`localDataInventory` calls `scanLocalInstances(l.cfg)` unconditionally (`:227`) and discards non-matching entries afterwards (`:232-236`), so a single-instance "is it gone yet?" query costs exactly as much as a full one: a walk over every directory level under the storage root plus an extra `os.Stat` per `data.log` via `isLiveSegmentData` → `hasCompactedMark` (`:165`, `server/compacted_mark.go:42-45`). README step 4 specifically asks for filtered polling until the instance disappears (`common/http/README.md:180-182`), and the target subtree is fully known — `localInstanceDataDir(cfg, bucket, rootPath)` already exists (`server/maintenance.go:143-156`) and could serve as the walk root. Failure: a service-mode node hosting tens of thousands of segment directories, polled every few seconds with a filter during reclamation → each poll re-walks the whole storage root and re-stats every `data.log`, contending with the WAL write path for disk IO and page cache and amplifying node latency for the duration of the reclamation window. Note the analogy to `HasLocalSegmentData` in the PR description does not hold: that function returns `filepath.SkipAll` on the first live `data.log` (`server/logstore.go:899`) and is typically O(1) on a node with data, whereas this scan has no early exit. *(surfaced during verification)*

tinswzy:
Renamed files after a pass over the naming conventions in each directory (thanks for the nudge):

- `cmd/wpcli/cmd/instance.go` was 181 lines holding both the family root and the whole `data` subcommand. Every other family root here is 12–20 lines and holds only the parent command plus `AddCommand` (`config.go` 12, `env.go` 12, `cluster.go` 16, `node.go` 19), with each subcommand in `<family>_<sub>.go`. Split into `instance.go` (12 lines) + `instance_data.go`, and `instance_test.go` → `instance_data_test.go` to match.
- `server/local_data_inventory.go` → `server/local_instance_data.go`, and `LocalDataInventory()` → `LocalInstanceData()` on both `LogStore` and `Server`. "Inventory" was a word only this PR used, while the types in the same file were `InstanceDataReport`/`InstanceDataEntry` — one file, two vocabularies. The symbols now read consistently as *local instance data*, alongside the existing `HasLocalSegmentData` / `localInstanceDataDir` / `removeLocalData`.

`common/http/management/instance_data_handler.go` is unchanged: a single-endpoint handler file named after its endpoint matches `log_health_handler.go` for `/admin/log-health`.

Note that `/admin/instance/delete` lives in `delete_handler.go` (grouped by operation, next to the log delete), so the `instance` noun is split across two handler files. That is pre-existing and left alone rather than pulled into this PR.

No behaviour change. Re-ran build / vet / golangci-lint (0 issues) and `go test -race -short` on `./server/ ./common/http/... ./cmd/wpcli/...`, all green, and re-checked a mutation on each renamed file to confirm the tests still fail when the corresponding production line is broken.

tinswzy:
Added a second commit doing the naming cleanup we discussed. Kept separate from the feature commit so the two can be read independently.

**`common/http/management/` before → after**

| Before | After | Serves |
|---|---|---|
| `instance_data_handler.go` | `detect_local_data_handler.go` | `GET /admin/instance/data` |
| `delete_handler.go` | `delete_local_data_handler.go` | `POST /admin/log/delete`, `POST /admin/instance/delete` |
| `logging_handler.go` | `log_level_handler.go` | `GET/POST /log/level` |
| (inline in `server.go`) | `memberlist_handler.go` | `GET /admin/memberlist` |
| — | `tenant_filter.go` | shared `(bucket_name, root_path)` parsing |

The detect/delete pair is the point: the read and write halves of node-local data now sit next to each other, instead of the `instance` noun being split across a file named after a verb.

`logging_handler.go` had three names for one thing — file said logging, endpoint is `/log/level`, constructor is `NewLogLevelHandler`.

The memberlist handler was the only one of the ten defined inline in `server.go`. It now takes plain callbacks like every other constructor in the package; it cannot take `AdminCallbacks`, since `http` imports `management` and not the reverse. Its content-negotiation test moved with it and gained a case for the nil callbacks the guards already handled but nothing exercised.

`tenantFilter` removes a verbatim ten-line copy of the filter parsing that `/admin/log-health` and `/admin/instance/data` each carried. The rule that a partial filter is dropped rather than silently widened to every tenant now lives in one place, which is what makes the consistency claimed in #349 structural instead of a convention.

Function names are unchanged and still track their endpoints (`NewInstanceDataHandler`, `NewLogDeleteHandler`, …), matching how `delete_handler.go` already worked — file named by what it does, constructors named by what they serve.

`admin_handler.go` is deliberately left as is.

No endpoint paths, payloads or behaviour change. build / vet / golangci-lint (0 issues) / `go test -race -short` across `./server/ ./common/http/... ./cmd/wpcli/...` all pass, and I mutation-checked the moved and extracted code (partial-filter drop, memberlist content negotiation, memberlist nil guards) to confirm the tests still fail when each is broken.

tinswzy:
Dropped the `wp instance data` command (third round of trimming).

I had justified it as "every other admin family has a `wp` command" — that was wrong. `/admin/log-health` already ships without one, so an endpoint with no CLI command is not a new inconsistency. The actual consumer here is the control plane, which speaks HTTP directly, so the command was 319 lines serving nobody in the loop this PR exists for. Easy to add later if operators reach for it.

Removed: `cmd/wpcli/cmd/instance_data.go`, its test, the `root.go` registration, and the `wp` example in the README.

History rewritten rather than appended, so the branch does not show the command being added and then removed:

- `enhance(admin): add GET /admin/instance/data …` — the feature, 13 files
- `refactor(admin): name each handler file after what it does to local data` — the renames

Re-ran build / vet / golangci-lint (0 issues) and `go test -race -short` on `./server/ ./common/http/... ./cmd/wpcli/...`; all pass. #349's scope section is updated to match.

tinswzy:
Docs pass.

`common/http/README.md` already carried the reference (overview-table row, full request/response/field section, and the reconciliation loop). What was missing was the two **live** docs an operator actually lands on when this problem bites:

- **`server/README.md`** documents `has_local_data` on the decommission-progress reply — which is exactly where a stuck drain leaves you. The flag says data is left but not whose it is, so it now points at `GET /admin/instance/data` for the attribution, which is what separates a cleanup still in flight from data stranded by an instance whose deletion never reached the node. `local_instance_data.go` added to Related Files.
- **`docs/admin-guides/disk-watermark.md`** offers four remedies for a node above the soft watermark, and all four assume the space is held by live instances. None of them frees a PVC filled by an instance that was never reclaimed. Added the check to run first, with the `active_processors` + `last_modified_ms` pair that tells a stranded instance from a merely idle one.

No new guide — the reference stays in one place and these link to it rather than restating it.

**No CLI command, by design** (dropped in the previous round): the consumer is the control plane over HTTP, and `/admin/log-health` already ships without a `wp` command. So there is nothing to add to `docs/admin-guides/wpcli/` — the cookbook is `wp`-recipes only, and adding a curl recipe there would put the same content in a second place.

`docs/wip/node-lifecycle/` and `docs/wip/wpcli/` still name the pre-rename handler files, but those are implementation plans (`Status: ✅ Implemented`, dated, checkbox steps) — records of what was planned at the time, not live references, so they are left alone. Say the word if you would rather they track the current tree.

czs007:

## Overview

Adversarial review surfaced three correctness issues in the new local instance-data inventory path, all introduced by this PR; each one makes `GET /admin/instance/data` report a state that contradicts the node's actual on-disk/in-memory state, which matters because the new README reconciliation loop treats that endpoint's output as a deletion gate.

## Medium

**`server/local_instance_data.go:156` — instance key mismatch in `local` storage mode makes `active_processors` / `delete_state` structurally empty**

In `local` storage mode the on-disk layout carries no bucket segment (`server/maintenance.go:125-126`, `localLogDataDir` returns `path.Join(root, rootPath, logIdStr)`), so `scanLocalInstances` takes the `bucketDepth = 0` branch and sets `bucket = ""` (`:95-96`, `:142-145`), building the key `GetInstanceKey("", rootPath)` = `"/<rootPath>"`. Both cross-annotation sources carry the real bucket: `annotateActiveProcessors` (`:261`) looks up keys derived from `GetLogKey(bucketName, rootPath, logId)` minus the trailing segment (`server/logstore.go:343-345, 352-353`), where `bucketName` originates from `cfg.Minio.BucketName` (`woodpecker/segment/segment_handle.go:109,167`); `annotateDeleteMarkers` (`:280`) looks up `GetInstanceKey(m.Bucket, m.RootPath)`. Both miss, so in `local` mode `active_processors` is always 0, `delete_state` always `""`, `marked_deleted_at_ms` always 0, and the reported `bucket_name` is an empty string rather than the instance's real bucket (the `?bucket_name=<b>&root_path=<r>` filter at `:232-234` therefore never matches either).

Failure: a node started from the standalone `cmd/main.go` server binary with `woodpecker.storage.type=local` and a non-empty `minio.bucketName`, while holding an actively-writing segment processor for an instance → `GET /admin/instance/data` reports that instance with empty `bucket_name`, `active_processors: 0` and empty `delete_state`, so a control plane following README step 2 (gate deletion on `active_processors == 0`) classifies a live instance as orphaned and issues `POST /admin/instance/delete`, removing Write-Ahead Log (WAL) data that is still being written.

Reachability note: `local` is an accepted storage type (`common/config/configuration.go:922`) and `NewServerWithConfig` does not reject it (`service.go:138` only builds an object-storage client for minio/service), so this configuration is not blocked; standard local/minio deployments do go through the embed client, which never calls `commonhttp.Start`, so the path is hard to reach in shipped deployment shapes.

Suggestion: in the `local` branch, populate `entry.BucketName` from `cfg.Minio.BucketName` and build the instance key from that value, so the reconstructed key matches the processor and marker keys.
*(raised by bigsheeper, xiaocai2333)*

---

**`server/local_instance_data.go:131` — an instance with an empty `rootPath` (bucket root) is silently dropped while `scan_errors` stays 0**

`ValidateRootPathValue` accepts an empty root path (`common/config/configuration.go:571-573`, `return nil // empty = bucket root`), and `ValidateMinioConfig` in service mode validates only that function (`:599-606`), so an empty `rootPath` is legal in service mode. The resulting layout is `root/<bucket>/<logId>/<segId>/data.log` — 4 segments relative to root — but `scanLocalInstances` requires `len(parts) >= bucketDepth+4 == 5` (`:131`), so every `data.log` under that tree is discarded by the `return nil` at `:131-133` **without incrementing `scanErrors`**. `HasLocalSegmentData` (`server/logstore.go:876-903`) imposes no path-shape requirement and still returns true for the same tree. The two endpoints share only the `isLiveSegmentData` liveness predicate (`:52-54`); path-shape parsing is not shared.

Failure: a service-mode node with an empty `rootPath` and live, uncompacted segment data → `GET /admin/instance/data` returns `instances: []`, `instance_count: 0`, `scan_errors: 0` while the same node's `decommission/progress` reports `has_local_data: true`; a control plane following README step 1 (stop only when `scan_errors > 0`) concludes the node is clean for that instance, skips `POST /admin/instance/delete`, and leaves the WAL data stranded permanently.

Suggestion: treat a path that is too short to parse as a scan blind spot (increment `scanErrors`) rather than silently skipping it, and accept the `bucketDepth+3` shape so bucket-root instances are inventoried. Related: the `rootPath == ""` check at `:147` is unreachable — reaching `:146` implies `logIdx > bucketDepth`, so the joined slice always has at least one non-empty segment — and can be removed.
*(raised by bigsheeper, xiaocai2333)*

---

**`server/local_instance_data.go:110` — `ENOENT` caused by this process's own background deletions is counted as a `scan_error`, halting the reclamation loop**

In the `WalkDir` callback, any `walkErr` below `root` increments `scanErrors` (`:106-112`; the `os.IsNotExist` exemption applies only when `p == root`), and a failed lazy `d.Info()` lstat does the same (`:150-154`). Two background tasks in the same process delete from that tree continuously: `server/compacted_file_cleanup.go:413` calls `os.Remove(dataLogPath)` each round, and the reclaim path at `server/maintenance.go:213` calls `os.RemoveAll` on log/instance directories.

Failure: `deletedLogReclaimTask` runs `os.RemoveAll` on an instance directory (or compacted-file cleanup removes a `data.log`) exactly while a control plane poll from README step 4 walks that path → the walk observes `ENOENT`, returns `scan_errors: 1`, and README step 1 ("stop if any answer has `scan_errors > 0`") aborts the reclamation loop at the precise moment the data was successfully cleaned, plus raises a scan-error alert for a blind spot that does not exist.

Suggestion: apply the distinction the code already makes elsewhere (`:107` for `root`, and `compacted_file_cleanup.go:414`'s `!os.IsNotExist`) to paths below `root` — a vanished path is evidence of cleanliness, not of an unreadable tree, and should not increment `scanErrors`.
*(raised by bigsheeper)*

tinswzy:
`wp instance data` is back, plus the piece that was actually missing.

On the file name: I looked for a `_command` suffix convention and there isn't one — `cmd/wpcli/cmd/` has 43 command files and 3 helpers, and **zero** use a suffix. The rule there is file name = command path (`node_list.go` ↔ `wp node list`, `logging_get_level.go` ↔ `wp logging get-level`), which is what lets you find a command's source from what you typed. A suffix would also be the third repetition of "command" in `cmd/wpcli/cmd/instance_data_command.go`, and would make this the only one of 43 files carrying it. So: `instance_data.go` ↔ `wp instance data`. Happy to rename if you still want the suffix — it is one `git mv` — but I did not want to break that lookup rule silently.

On "不方便查看和使用": the file name was not the real obstacle. `wp --help` finds the command, but nothing pointed an operator *to* it, and `docs/admin-guides/wpcli/cookbook.md` is where they look. Added **recipe 12 — Find local data left behind by a deleted instance**: how the stranded case arises, the three query forms, how to read `PROCS` / `DELETE_STATE` / `LIVE` (including why `SEGMENTS` > `LIVE` explains the disagreement with `has_local_data`), the curl for the per-node delete, and the poll until it is gone.

The recipe makes one thing explicit that the command alone does not: **`--strict` is a safety flag, not a formatting one.** Without it an unreachable node is just a row in the table; with it the command exits non-zero. An unreachable node still holds intact data, so acting on a partial view is the accident this whole endpoint exists to prevent.

The per-node delete stays curl-only and deliberately has no `wp` subcommand — deletion should be explicit and targeted at the nodes the query named.

Verified the documented flags against the real `--help` output rather than from memory. build / vet / golangci-lint (0 issues) / `go test -race -short` on `./cmd/wpcli/...` all pass. #349 scope and the PR description are back in sync.

PR #355, opened 2026-09-20
enhance(metrics): give a missing object its own status, not "error"

### LINKED ISSUE #354: enhance(metrics): stat_object counts a normal not-found as an error, so any object-storage error ratio reads ~80%
## Summary

`StatObject` records `status="error"` for every non-nil error, including a plain not-found:

```go
// common/minio/minio_handler.go:300-307
info, err := m.client.StatObject(ctx, bucketName, objectName, opts)
if err != nil {
    metrics.WpObjectStorageOperationsTotal.WithLabelValues(..., "stat_object", "error").Inc()
```

But `stat_object` is an existence probe, and "not there yet" is its normal answer. `readRemoteFooter` uses it for the idempotency check before compaction (`writer_impl.go:2138`), and the compacted-file cleanup reconcile uses it to decide whether a footer is durable (`compacted_file_cleanup.go:482`). In both, a miss is the expected outcome for a segment that has not been uploaded yet.

## What it costs

On an idle-to-moderate cluster the counter is dominated by these probes. Measured on a healthy three-node cluster under steady write load:

| operation | status | rate |
| --- | --- | --- |
| `stat_object` | error | **11.698/s** |
| `stat_object` | success | 1.737/s |
| `put_object` | success | 1.151/s |

Nothing was failing — `put_object` was 100% success — yet the obvious error ratio

```promql
sum(rate(woodpecker_server_object_storage_operations_total{status="error"}[5m]))
/ sum(rate(woodpecker_server_object_storage_operations_total[5m]))
```

reads **0.80**. Any alert on that ratio fires permanently on a healthy cluster, which in practice means it gets silenced and object storage ends up with no alerting at all.

The dashboards now work around it by excluding `stat_object` from the ratio, but every other consumer of this counter inherits the same flaw, and the workaround also discards genuine `stat_object` failures — a real object-storage outage during a probe is no longer distinguishable from a miss.

## Proposal

Give a negative lookup its own status rather than folding it into `error`:

```go
if err != nil {
    status := "error"
    if errors.Is(err, ...) || minio.ToErrorResponse(err).Code == "NoSuchKey" {
        status = "not_found"
    }
    metrics.WpObjectStorageOperationsTotal.WithLabelValues(..., "stat_object", status).Inc()
```

`GetObject` deserves the same treatment for the same reason.

Then the ratio means what it looks like, and `status="not_found"` is independently useful — a sudden change in probe miss rate says something about compaction progress.

### Compatibility

This narrows what `status="error"` counts, so anything already alerting on it gets quieter, not noisier. Worth calling out in the release notes all the same.

### Alternative considered

Leaving the counter alone and requiring every consumer to write `operation!="stat_object"`. That is what the dashboards do today, but it pushes a non-obvious exclusion onto every future query and still loses real probe failures.

tinswzy:
Closes #354.

## The problem

`StatObject` recorded `status="error"` for every non-nil error, including a plain not-found. But it is an existence probe — `readRemoteFooter`'s idempotency check and the compacted-file cleanup reconcile both use it — and "not there yet" is its normal answer for a segment that has not been uploaded.

Measured on a healthy three-node cluster under steady write load:

| operation | status | rate |
| --- | --- | --- |
| `stat_object` | error | **10.4/s** |
| `stat_object` | success | 1.7/s |
| `put_object` | success | 1.1/s |

Nothing was failing — `put_object` was 100% success — yet the obvious ratio

```promql
sum(rate(woodpecker_server_object_storage_operations_total{status="error"}[5m]))
/ sum(rate(woodpecker_server_object_storage_operations_total[5m]))
```

read **0.78**. An alert on that fires permanently, which in practice means it gets silenced and object storage ends up with no alerting at all.

## The change

`stat_object` labels a missing object `not_found`, on both backends.

That is the only operation where it is reachable. An earlier revision also classified `get_object` and `get_object_data_info`; review showed those branches were dead. Both backends return a lazily-backed reader — Azure's `NewBlobReaderWithSize` hands one back without issuing a request, and minio-go's `GetObject` only runs `CheckValidBucketName`/`CheckValidObjectName` first (`api-get-object.go:32-39`) — so the error observed there can only be a name-validation failure. A GET miss surfaces on the first `Read`, under `read_object_full`. (The first version of this description claimed the opposite; it was wrong.)

Mutating operations are untouched: for a put, a copy or a conditional put, an error is an error.

## Each backend classifies with its own predicate

`IsObjectNotExistsError` is dispatched per backend on the `ObjectStorage` interface, and the implementations are not interchangeable — the MinIO one parses an S3 `ErrorResponse`, the Azure one an `azcore.ResponseError`. So each backend classifies at its own recording site with its own predicate rather than sharing a copy of the rule.

MinIO records its metrics in the handler *beneath* the interface, where no `ObjectStorage` instance is in scope, so it calls the MinIO predicate directly. That is noted in the code, since it would otherwise read as an oversight.

### What I got wrong first

An earlier revision of this PR widened `IsObjectNotExists` to also accept `NotFound`, `NoSuchBucket` and a bare 404, reasoning that a HEAD carries no body for a code to be parsed from. Both halves were wrong:

- minio-go already normalises a 404 on a keyed request to `NoSuchKey` (`api-error-response.go:143`), so nothing needed widening.
- More seriously, that function is not a metrics helper. It backs `IsObjectNotExistsError`, which **fifteen production call sites branch on** — including `compacted_file_cleanup.go:484`, where it decides whether a footer is legitimately absent or the read failed. Folding `NoSuchBucket` into it would let a misconfigured bucket read as "not uploaded yet" on exactly the path that governs reclaiming local data.

Reverted to its original one-line form. The test now pins missing-bucket to `error` so it cannot drift back.

## Azure was missing the metrics entirely

It recorded only `condition_put_object` and `put_fenced_object` while MinIO counted nine operations. `stat_object` and `put_object` are added — the probe classified through the Azure predicate, the write left as a plain `error`, since a write is not an existence check. `put_object` also emits the `RequestBytes` observation that MinIO pairs with `BytesTransferred` everywhere.

`get_object` is deliberately **not** instrumented here. Recording at reader construction would mean an unconditional `success` and a latency sample timing a struct allocation; a counter that can only ever say "fine" is worse than none. Instrumenting `BlobReader.Read`/`ReadAt`, and the MinIO equivalent, belongs in its own change.

Azure's `readStatus` also separates a container 404 from an object 404. `IsObjectNotExistsError` keys only on the status code, and the container is created only when `Minio.CreateBucket` is set, so an absent container is a reachable steady state — without the distinction, a wholly unavailable backend would read as "the footer has not been uploaded yet".

## Verification

Built the change, rolled the three servers onto it, and measured over a window entirely after the rollout:

| operation | status | count / 2 min |
| --- | --- | --- |
| `stat_object` | **not_found** | 1221.8 |
| `stat_object` | success | 204.6 |
| `put_object` | success | 134.9 |
| any | **error** | **0** |

The old form of the ratio — counting `not_found` back in — still reads **0.7826** over those same calls. Same traffic, same volume, correct classification: that is the evidence the change is a relabelling and not a change in behaviour.

## Tests

Both directions are covered, and the second matters more: a genuine failure — `AccessDenied`, `InternalError`, `SlowDown`, a bare connection error, a missing bucket or container — must **not** be reclassified as a missing object. Getting that wrong silences a real outage, which is worse than the ratio this fixes.

`TestAzureStatObjectRecordsStatus` drives `StatObject` through the existing mock and reads the counter back, rather than asserting on the classifier. That distinction was review's sharpest point: a test on the pure function passes just as happily with the recording wired up wrong, which is exactly the defect that shipped in the previous revision. Verified by removing the `ContainerNotFound` branch — the test fails, and passes again once restored.

## Compatibility

This narrows what `status="error"` counts, so anything already alerting on it gets quieter, never noisier. Worth a release-note line all the same.


czs007:

Adversarial review of the object-storage metrics instrumentation found one high-severity correctness defect in the new Azure `get_object` recording site, plus two medium and two low-severity issues in the new instrumentation, and two real but pre-existing hazards listed separately for follow-up.

## High

**`common/objectstorage/azure_object_storage.go:279` — Azure `get_object` records an unconditional `success`; the `readStatus` branch is dead and the latency sample measures struct construction.**
`GetObject` (`:277-287`) classifies the error returned by `NewBlobReaderWithSize`, but that function (`:112-114`) does nothing except `return &BlobReader{...}, nil` — it never issues an HTTP request, so the error is structurally always `nil` and `a.readStatus(err)` at `:282` can never execute. The actual `DownloadStream` calls live in `BlobReader.Read` (`:126`) and `BlobReader.ReadAt` (`:155`), neither of which is instrumented. The `WpObjectStorageOperationLatency` observation at `:285` therefore times a struct allocation, collapsing the Azure `get_object` distribution at ~0 ms.
*Failure:* with `CloudProvider=azure`, reading a missing or forbidden blob (e.g. `reader_impl.go:690` fetching a block, `stagedstorage/writer_impl.go:2146` fetching a footer; deleted container, expired SAS/key 403, or a dropped connection) → `woodpecker_server_object_storage_operations_total{operation="get_object",status="success"}` still increments while `error`/`not_found` stay at 0, and the real `DownloadStream` failure inside `Read`/`ReadAt` is counted nowhere. Nine production call sites are affected: `server/storage/objectstorage/writer_impl.go:224,506,2151`; `objectstorage/reader_impl.go:141,343,690`; `server/storage/stagedstorage/writer_impl.go:2146`; `stagedstorage/reader_impl.go:323,1613`.
*Suggestion:* note that `minio-go` is lazy in the same way (`api-get-object.go:32-39` only validates the bucket/object names before returning), so `minio_handler.go:102-103` has also been recording `success` at reader construction. The right fix is to move both the `Inc` and the `Observe` to where the first byte actually arrives — `BlobReader.Read`/`ReadAt` on Azure, `obj.Stat()`/`ReadObjectFull` on MinIO — or to not ship an always-`success` `get_object` counter at all, rather than special-casing Azure. *(raised by bigsheeper, czs007)*

## Medium

**`common/objectstorage/azure_object_storage.go:448` — Azure `readStatus` misclassifies a container-level 404 (`ContainerNotFound`) as `not_found`.**
`readStatus` (`:448-453`) delegates to `IsObjectNotExistsError` (`:457-463`), which keys only on `azError.StatusCode == http.StatusNotFound` and ignores `ErrorCode`, while the same file explicitly distinguishes `bloberror.ContainerNotFound` at `:259`. The container is only checked/created when `c.Minio.CreateBucket` is set (`:249-273`) and that block is skipped entirely in service mode, so a genuinely absent container is a reachable steady state rather than a startup-only misconfiguration.
*Failure:* `CloudProvider=azure` with `Minio.CreateBucket=false` and a mistyped or deleted container, then any footer probe (`reader_impl.go:131`, `stagedstorage/writer_impl.go:2138`) hitting `GetProperties` → Azure returns 404 with `x-ms-error-code: ContainerNotFound`, recorded as `stat_object{status="not_found"}`, so a fully unavailable backend contributes zero errors and looks identical to "the footer has not been uploaded yet".
*Suggestion:* `azcore` v1.11.1 populates `ErrorCode` from the `x-ms-error-code` header (`internal/exported/response_error.go:23-27`), which is present on HEAD/`GetProperties` responses, so `readStatus` can branch on `bloberror.ContainerNotFound` before falling through to `not_found`. Worth noting the blast radius is bounded: `PutObject` (`:294-296`) and `condition_put_object` (`:333-336`) still record `error` once the container is gone, so the overall error rate does not go to zero. This is hardening and need not block the read-path fix above. *(raised by czs007, bigsheeper)*

**`common/objectstorage/azure_object_storage_test.go:984` — both new tests call the classifier directly and bypass the recording sites, which is where the defect actually is.**
`TestAzureReadStatus` (`:984-1006`) builds `a := &AzureObjectStorage{}` (`:985`) and asserts on `a.readStatus(err)`; `TestReadStatus` (`common/minio/minio_client_helper_test.go:487` onward) likewise asserts only on the pure function. What this PR actually changes is which status string reaches `WithLabelValues` at each call site, and nothing covers that layer.
*Failure:* running `make unit-test` (`go test -race -short -cover -failfast`) on this branch → the suite passes green while Azure `GetObject` emits `get_object{status="success"}` for a missing or forbidden blob (the High item above), so CI cannot stop that defect from merging.
*Suggestion:* the scaffolding already exists in the same file — `newTestAzureStorage` (`:361-365`) plus `mockBlobClient.downloadStreamFunc`/`getPropertiesFunc` (`:47`, `:50`, already used at `:183`, `:293`, `:423`, `:503` and elsewhere). Injecting `&azcore.ResponseError{StatusCode: 404}`, driving `a.StatObject`/`storage.GetObject`, and asserting the counter with `testutil.ToFloat64` pins the label production actually emits. The MinIO side takes the same shape through `NewMinioHandlerWithClient` (`common/minio/minio_handler.go:84`), already used in `minio_handler_impl_test.go`. *(raised by czs007)*

## Low

**`common/minio/minio_handler.go:97` — the new `readStatus` calls on `get_object` / `get_object_data_info` are dead branches, and the PR description has the reasoning inverted.**
In `minio-go` v7.0.73, `GetObject` (`api-get-object.go:32-39`) only runs `CheckValidBucketName`/`CheckValidObjectName` before returning a lazily-backed reader, so the error observed at `minio_handler.go:97` and `:113` can only be a name-validation error, never `NoSuchKey`. The `not_found` branch does not mislabel anything, but it is also unreachable. The PR body's claim that "`read_object_full` … a miss cannot surface there" is backwards: a GET miss surfaces precisely on the first `Read` — the `objectReader.Read` failure path in `ReadObjectFull` (`common/minio/minio_client_helper.go:291-292`, recorded as `read_object_full{status="error"}`) — and, for `get_object_data_info`, in `obj.Stat()` (`minio_handler.go:118-122`, recorded as a hardcoded `"error_stat"`). Severity stays low because all nine production `GetObject` call sites are guarded by a preceding `StatObject` (`writer_impl.go:207→224`, `reader_impl.go:131→141`, `stagedstorage/writer_impl.go:2138→2146`, `stagedstorage/reader_impl.go:309→323`), so a steady-state GET does not hit a missing object except when racing a delete, where `error` is a defensible label anyway.
*Suggestion:* correct the PR body and the inline comments, and either push the classification down to `obj.Stat()`/`ReadObjectFull` or drop these two `readStatus` calls. On either backend, `stat_object` is the only operation where `not_found` is genuinely reachable in this PR. *(raised by bigsheeper, czs007)*

**`common/objectstorage/azure_object_storage.go:291` — Azure `PutObject` records `BytesTransferred` but omits the paired `RequestBytes.Observe`.**
The new instrumentation records `WpObjectStorageBytesTransferred.Add(objectSize)` at `:299-301` and returns, with no `WpObjectStorageRequestBytes.Observe(...)`. The MinIO side emits the two together everywhere (`minio_handler.go:145-146` for `put_object`, `:174-175` for `condition_put_object`, `:256-259` for `put_fenced_object`, `:126-127` on the read path). The result is that request-size distribution (p50/p99 object size) for `put_object` has no series at all on the Azure backend.
*Suggestion:* add the matching `RequestBytes.Observe(float64(objectSize))` next to the existing `Add`. *(raised by czs007, bigsheeper)*

- **`common/objectstorage/azure_object_storage.go:339` — Azure `condition_put_object` / `put_fenced_object` are instrumented only on the failure path, so the success path records nothing.** In `PutObjectIfNoneMatch` (`:305-340`) the precondition-failed idempotent branch records `condition_put_object{success}` (`:327-329`) and the error branch records `error` (`:334-335`), but a normal successful write falls through to a bare `return err` at `:339` with no recording; `PutFencedObject` (`:342-378`) likewise has only the error branch at `:372-373`. *Failure:* on `CloudProvider=azure` with conditional writes enabled, a normal steady-state workload (e.g. every block write through `writer_impl.go:1203`) → `woodpecker_server_object_storage_operations_total{operation="condition_put_object"}` only ever gains `status="error"` samples, so the per-operation error ratio reads 100% and the write throughput/latency panels stay empty on Azure. Unrelated to this PR: the diff touches only `GetObject` (`:277-287`), `PutObject` (`:289-303`), `StatObject` (`:380-396`) and the new `readStatus` (`:444-453`) in this file — `:327-339` and `:371-377` are byte-identical to base (`c5adc7a`). Recommend landing that separate fix before this PR if the Azure write-side metrics are needed alongside these read-side ones. *(surfaced during verification)*

- **`common/minio/minio_handler.go:105` — `GetObjectDataAndInfo` has no production caller.** A repo-wide grep at `8b60e59` finds it only in the interface declaration (`minio_handler.go:56`), the implementation (`:107`), one unit test (`minio_handler_impl_test.go:485-492`) and the mockery-generated mock (`mocks/mocks_minio/mock_minio_handler.go:153` onward); none of the production `.GetObject(` call sites (`minio_object_storage.go:58` and nine under `server/`) route through it, so the instrumentation added for it — including the `readStatus` at `:113` — has no practical payoff. Note the function cannot simply be deleted on its own: it is part of the `MinioHandler` interface, so removal has to take the interface and the mock with it. Unrelated to this PR: the method and its lack of callers both exist unchanged at base (`c5adc7a`); this PR only touched the recording lines in `GetObject` (`:96-99`), the `GetObjectDataAndInfo` failure branch (`:112-115`) and `StatObject` (`:301-306`). *(raised by bigsheeper, czs007)*

tinswzy:
All correct, and the third item is the one I'll take away from this.

**`get_object` classification was dead on both backends.** Verified: `NewBlobReaderWithSize` returns `&BlobReader{...}, nil` without issuing a request, and minio-go's `GetObject` only runs the two name checks before handing back a lazily-backed reader (`api-get-object.go:32-39`). Those branches are gone, back to a plain `error`, with the reason in a comment so it doesn't look like an omission.

And you're right that my description had it backwards — a GET miss surfaces precisely on the first `Read`, under `read_object_full`, not "cannot surface there". Corrected in the PR body.

**Azure `get_object` is dropped rather than fixed.** Recording at reader construction is an unconditional `success` plus a latency sample timing a struct allocation, and a counter that can only ever say "fine" is worse than no counter. Instrumenting `BlobReader.Read`/`ReadAt` and the MinIO equivalent is a real change with its own shape, so it should not ride along here.

**`ContainerNotFound`** — taken. `IsObjectNotExistsError` keys only on the status code, and since the container is created only when `Minio.CreateBucket` is set, an absent one is a reachable steady state rather than a startup problem. `readStatus` now branches on `bloberror.ContainerNotFound` first.

**`RequestBytes` on Azure `PutObject`** — added.

**On the tests.** This was the sharpest of the five. Both of mine asserted on the classifier, and a test on a pure function passes just as happily with the recording wired up wrong — which is exactly the defect that shipped. `TestAzureStatObjectRecordsStatus` now drives `StatObject` through `newTestAzureStorage` + `getPropertiesFunc` and reads the counter back with `testutil.ToFloat64`. I checked it actually bites: removing the `ContainerNotFound` branch fails it, restoring it passes.

**On the two pre-existing hazards** — Azure's `condition_put_object`/`put_fenced_object` recording only on the failure path (so their error ratio reads 100%), and `GetObjectDataAndInfo` having no production caller. Both confirmed against base. Agreed they're separate; I'll file them unless you'd rather they go with whoever picks up the Azure write path.

czs007:

**Verified:**
- `common/objectstorage/azure_object_storage.go:277-283` — `GetObject` no longer records `get_object` ops/latency at all, so the unconditional-success counter and the latency sample that timed a struct allocation are both gone.
- `common/objectstorage/azure_object_storage.go:446-458` — the `ContainerNotFound` pre-check runs *before* `IsObjectNotExistsError` (which keys only on `StatusCode == 404`), so a missing container is labelled `error`, not `not_found`, and `IsObjectNotExistsError`'s own semantics are untouched for its fifteen production call sites.
- `common/objectstorage/azure_object_storage.go:295-298` — `WpObjectStorageRequestBytes` is passed exactly the four labels its definition declares (`service_metrics.go:239-245`) and observed only inside `if err == nil`, so failed uploads cannot inflate the size histogram.
- `common/minio/minio_handler.go:94-100` and `:112-118` — the `get_object` / `get_object_data_info` error paths are back to a hard-coded `"error"`, which is correct because minio-go's `GetObject` returns before issuing any request, making a `not_found` classification unreachable there.
- `common/minio/minio_handler.go:305-313` — `stat_object` is now the only `readStatus` consumer, so the revert leaves no unused helper behind.
- `common/objectstorage/azure_object_storage_test.go:1004-1015` — `TestAzureStatObjectRecordsStatus` drives the real recording site through the existing mock and asserts on before/after counter deltas, so a mis-wired label fails the test rather than passing a classifier-only assertion.

## Low — non-blocking, no change required in this PR

- **`common/minio/minio_handler.go:308` — a missing bucket is recorded as `stat_object{status="not_found"}` on the most common MinIO configuration.** `StatObject` issues a HEAD, and HEAD responses carry no body, so minio-go's `httpRespToErrorResponse` always falls back to its header branch and synthesises `Code: "NoSuchKey"` whenever `objectName` is non-empty (`api-error-response.go:133-150`). `readStatus` therefore returns `not_found` for a bucket-level 404, which is asymmetric with the Azure path, where `ContainerNotFound` is explicitly caught and recorded as `error` (`azure_object_storage.go:446-453`). Failure: with `minio.region` set to a non-empty value (the norm for real S3 deployments — `config/woodpecker.yaml:334`, threaded through at `minio_client_helper.go:134`) or with the bucket location already cached, misconfiguring or deleting `minio.bucketName` makes every footer `StatObject` return a bodyless 404 → `stat_object{status="error"}` stays at zero and the error-ratio alert this PR is restoring never fires for that entire class of backend failure on the stat path, whereas before this PR the same error was recorded as `error`. Empirically confirmed on PR head against an httptest gateway returning a bodyless 404: `Region:"us-west-2"` → HEAD → `NoSuchKey` → `not_found`; `Region:""` → GET bucket-location first → `NoSuchBucket` → `error`. Note that `minio_client_helper_test.go:514-521` pins `missing bucket → error` using a hand-constructed `ErrorResponse{Code: "NoSuchBucket"}`, which `StatObject` cannot actually produce under these configurations — the assertion gives less protection than it appears to. Suggestion: either document the limitation inline, or have `readStatus` disambiguate a HEAD 404 by re-checking bucket existence before classifying it as `not_found`. *(raised by czs007)*

- **`server/storage/stagedstorage/reader_impl.go:150` — a missing bucket/container is treated as "object does not exist", turning a recoverable backend outage into a permanent `ErrEntryNotFound` on the read path.** Neither backend's predicate separates the two cases: Azure's `IsObjectNotExistsError` inspects only the 404 status code (`azure_object_storage.go:462-467`), so `ContainerNotFound` returns `true` — the very condition this PR's own `readStatus` classifies as an unavailable backend; on the MinIO side a bucket-level HEAD 404 is synthesised into `NoSuchKey` as described above. Failure: the Azure container is deleted or misnamed (or the MinIO bucket is wrong with `minio.region` set) while a compacted segment's local `data.log` has already been reclaimed by compacted-file cleanup → `StatObject`'s 404 is read as "footer absent" and `reader_impl.go:169` returns `werr.ErrEntryNotFound`, a permanent, non-retriable error, even though the data in object storage is intact. Unrelated to this PR: the diff touches only the instrumentation and classification lines in `common/minio/*` and `common/objectstorage/azure_object_storage*.go`; `IsObjectNotExists`, `AzureObjectStorage.IsObjectNotExistsError` and all of `reader_impl.go` are outside it, and `readStatus` feeds metrics only, never this decision. The cleanup side is safe (`compacted_file_cleanup.go:482-489` and `:352-359` retain `data.log` and the mark), so there is no data-loss exposure. One request that *does* concern this PR: the description cites "missing bucket is recorded as error" as an existing safeguard — that framing does not hold, and adjusting the wording would avoid overstating current protection. A predicate or call-site fix that routes bucket/container-level codes separately should be a separate PR; recommend merging that separate fix before relying on this behaviour. *(surfaced during verification)*
- **`common/objectstorage/azure_object_storage.go:277` — Azure has no per-call `get_object` counter symmetric with MinIO's (`common/minio/minio_handler.go:104-105`).** This is a base-branch gap, not a regression: the pre-PR `GetObject` was a bare `return NewBlobReaderWithSize(...)` with no instrumentation, and this PR only adds a comment explaining why nothing is counted there — no metric was removed. Azure read transfers are not entirely unmeasured either, since `minioHandler.ReadObjectFull` is backend-agnostic and records ops/latency/`RequestBytes`/`BytesTransferred` for the Azure `*BlobReader` too (`minio_client_helper.go:291-301`). The remaining gaps are narrow: the missing per-call counter, plus `BlobReader.ReadAt`/`Seek` (`azure_object_storage.go:149-175`) and the `io.ReadFull` at `stagedstorage/reader_impl.go:333`, which bypass `ReadObjectFull` for both backends alike. Reasonable scope boundary for this PR. *(raised by bigsheeper, czs007)*

